PR #8, opened 2017-06-21
Fix broken links

kunanit:
Fixed some broken links to core components and html components pages. Also corrects a grammar error.


chriddyp:
thank you @kunanit !

PR #9, opened 2017-06-22
Added missing curly brace in code snippet

JamieHouston:
Added missing syntax for snippet to link plotly css file. Without it snippet doesn't work.


chriddyp:
great catch. thank you!!

PR #10, opened 2017-06-22
Fixed officially discouraged style

lebigot:
Python efficiently creates a single string literal from multiple consecutive ones. This is officially recommended, compared to adding strings, which is wasteful (multiple strings created in memory, that are concatenated one by one into bigger and bigger strings).


chriddyp:
ah yeah, way better! thanks!

PR #11, opened 2017-06-23
Fix: generate_table would otherwise fail with a more general index

lebigot:
Reference: https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.iloc.html


chriddyp:
ah yeah, much better. thank you @lebigot!

PR #12, opened 2017-06-25
Update README.md

Volodymyrk:
Instructions on how to run an app locally. Should save the next person few hours


chriddyp:
Thanks @Volodymyrk ! I added virtual env instructions as well. :clinking_glasses:

PR #14, opened 2017-07-25
Update deployment.py

cforelle:
Unless I'm mistaken, dash.Dash() is the Flask object (that is, the WSGI application), not the Flask "server" per se. I got a little tripped up by the "server" / "app" terminology in the docs, because for Flask in the usual conventions "app" is the WSGI callable, while for Dash "app.server" is.


chriddyp:
Thanks @cforelle ! That's definitely better 👍

PR #19, opened 2017-08-04
memoization example Python 2 and 3 compatible

taspotts:


chriddyp:
🙇  thank you @taspotts !

PR #23, opened 2017-08-21
More examples

alishobeiri:


chriddyp:
@alishobeiri - is this ready to review?

alishobeiri:
Hahah I actually made a pull request to view the file diffs, but if you could give it a could overview and let me know what you think it wouldn't hurt :D

chriddyp:
image

very nice!!

alishobeiri:
There's still some issues with this, like `object` type and there's also some other ones that I haven't fixed/don't quite know what to replace with. 

All the different `Type` cases were basically hardcoded so I was worried maybe it wouldn't scale.

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+import json
+import dash as _dash
+import pandas as pd
+import styles
+from datetime import datetime as dt
+
+_current_path = _os.path.join(_os.path.dirname(_os.path.abspath(dcc.__file__)),
+                              'metadata.json')
+
+def object_hook_handler(obj):
```
chriddyp: could you add some comments here?

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+            obj['Type'] = obj['type']['name']
+        obj.pop('type')
+
+    if 'defaultValue' in obj:
+        if(obj['defaultValue']['value'] == 'true'):
+            obj['defaultValue']['value'] = 'True'
+        elif(obj['defaultValue']['value'] == 'false'):
+            obj['defaultValue']['value'] = 'False'
+        elif(type(obj['defaultValue']['value']) == dict):
+            print(obj['defaultValue']['value'])
```
chriddyp: ⚡️  all `print` statements

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+            if(objVal['name'] == 'shape'):
+                for i in objVal['value']:
+                    holder['Array Of'][0][i] = objVal['value'][i]['name']
+                obj['Type'] = holder.items()[0]
+            elif('value' in objVal):
+                for i in objVal['value']:
+                    holder['Array Of'].append(str(i))
+                obj['Type'] = holder.items()[0]
+            else:
+                obj['Type'] = "Array of: [" + objVal['name'] + "]"
```
chriddyp: For array and object, this should probably recurse so that it can handle things like `arrayOf(shape({foo: shape({bar: bool})}))`. 

For example, `options` currently looks something like:
image

but we should make it more readable so that it reads like:
```
A list of `dict`. Each `dict` as the following keys: 'label', 'value', 'disabled'.
- 'label' (string, required) The visible label of the dropdown
- 'value' (string, required) The hidden value of the dropdown. This value gets passed into dash callbacks.
- 'disabled' (boolean, optional) - If True, the option is disabled
```

I do this type of recursive document generation in the component docstrings. Here's what this looks like for `help(dcc.Graph)`:
```
 |  Keyword arguments:
 |  - id (string; required)
 |  - clickData (dict; optional): Data from latest click event
 |  - hoverData (dict; optional): Data from latest hover event
 |  - clear_on_unhover (boolean; optional): If True, `clear_on_unhover` will clear the `hoverData` property
 |  when the user "unhovers" from a point.
 |  If False, then the `hoverData` property will be equal to the
 |  data from the last point that was hovered over.
 |  - selectedData (dict; optional): Data from latest select event
 |  - relayoutData (dict; optional): Data from latest relayout event which occurs
 |  when the user zooms or pans on the plot
 |  - figure (dict; optional): Plotly `figure` object. See schema:
 |  https://plot.ly/javascript/reference
 |  - style (dict; optional): Generic style overrides on the plot div
 |  - animate (boolean; optional): Beta: If true, animate between updates using
 |  plotly.js's `animate` function
 |  - animation_options (dict; optional): Beta: Object containing animation settings.
 |  Only applies if `animate` is `true`
 |  - config (optional): Plotly.js config options.
 |  See https://plot.ly/javascript/configuration-options/
 |  for more info.. config has the following type: dict containing keys 'staticPlot', 'editable', 'edits', 'autosizable', 'queueLength', 'fillFrame', 'frameMargins', 'scrollZoom', 'doubleClick', 'showTips', 'showAxisDragHandles', 'showAxisRangeEntryBoxes', 'showLink', 'sendData', 'linkText', 'displayModeBar', 'modeBarButtonsToRemove', 'modeBarButtonsToAdd', 'modeBarButtons', 'displaylogo', 'plotGlPixelRatio', 'topojsonURL', 'mapboxAccessToken'.
 |  Those keys have the following types:
 |    - staticPlot (boolean; optional): no interactivity, for export or image generation
 |    - editable (boolean; optional): we can edit titles, move annotations, etc - sets all pieces of `edits`
 |  unless a separate `edits` config item overrides individual parts
 |    - edits (optional): a set of editable properties. edits has the following type: dict containing keys 'annotationPosition', 'annotationTail', 'annotationText', 'axisTitleText', 'colorbarPosition', 'colorbarTitleText', 'legendPosition', 'legendText', 'shapePosition', 'titleText'.
 |  Those keys have the following types:
 |    - annotationPosition (boolean; optional): annotationPosition: the main anchor of the annotation, which is the
 |  text (if no arrow) or the arrow (which drags the whole thing leaving
 |  the arrow length & direction unchanged)
 |    - annotationTail (boolean; optional): just for annotations with arrows, change the length and direction of the arrow
 |    - annotationText (boolean; optional)
 |    - axisTitleText (boolean; optional)
 |    - colorbarPosition (boolean; optional)
 |    - colorbarTitleText (boolean; optional)
 |    - legendPosition (boolean; optional)
 |    - legendText (boolean; optional): edit the trace name fields from the legend
 |    - shapePosition (boolean; optional)
 |    - titleText (boolean; optional): the global `layout.title`
 |    - autosizable (boolean; optional): DO autosize once regardless of layout.autosize
 |  (use default width or height values otherwise)
 |    - queueLength (number; optional): set the length of the undo/redo queue
 |    - fillFrame (boolean; optional): if we DO autosize, do we fill the container or the screen?
 |    - frameMargins (number; optional): if we DO autosize, set the frame margins in percents of plot size
 |    - scrollZoom (boolean; optional): mousewheel or two-finger scroll zooms the plot
 |    - doubleClick (a value equal to: false, 'reset', 'autosize', 'reset+autosize'; optional): double click interaction (false, 'reset', 'autosize' or 'reset+autosize')
 |    - showTips (boolean; optional): new users see some hints about interactivity
 |    - showAxisDragHandles (boolean; optional): enable axis pan/zoom drag handles
 |    - showAxisRangeEntryBoxes (boolean; optional): enable direct range entry at the pan/zoom drag points
 |  (drag handles must be enabled above)
 |    - showLink (boolean; optional): link to open this plot in plotly
 |    - sendData (boolean; optional): if we show a link, does it contain data or just link to a plotly file?
 |    - linkText (string; optional): text appearing in the sendData link
 |    - displayModeBar (a value equal to: true, false, 'hover'; optional): display the mode bar (true, false, or 'hover')
 |    - modeBarButtonsToRemove (list; optional): remove mode bar button by name.
 |  All modebar button names at https://github.com/plotly/plotly.js/blob/master/src/components/modebar/buttons.js
 |  Common names include:
 |   - sendDataToCloud
 |  - (2D): zoom2d, pan2d, select2d, lasso2d, zoomIn2d, zoomOut2d, autoScale2d, resetScale2d
 |  - (Cartesian): hoverClosestCartesian, hoverCompareCartesian
 |  - (3D): zoom3d, pan3d, orbitRotation, tableRotation, handleDrag3d, resetCameraDefault3d, resetCameraLastSave3d, hoverClosest3d
 |  - (Geo): zoomInGeo, zoomOutGeo, resetGeo, hoverClosestGeo
 |  - hoverClosestGl2d, hoverClosestPie, toggleHover, resetViews
 |    - modeBarButtonsToAdd (list; optional): add mode bar button using config objects
 |    - modeBarButtons (boolean | number | string | dict | list; optional): fully custom mode bar buttons as nested array,
 |  where the outer arrays represents button groups, and
 |  the inner arrays have buttons config objects or names of default buttons
 |    - displaylogo (boolean; optional): add the plotly logo on the end of the mode bar
 |    - plotGlPixelRatio (number; optional): increase the pixel ratio for Gl plot images
 |    - topojsonURL (string; optional): URL to topojson files used in geo charts
 |    - mapboxAccessToken (boolean | number | string | dict | list; optional): Mapbox access token (required to plot mapbox trace types)
 |  If using an Mapbox Atlas server, set this option to '',
 |  so that plotly.js won't attempt to authenticate to the public Mapbox server.
 ```

I recommend adapting this code to look more like the code recursive code here: https://github.com/plotly/dash/blob/master/dash/development/base_component.py#L287-L419

Comment (chriddyp) [inline tutorial/core_component_examples.py]: I would play around with a few different markup types too. Perhaps make the `Type` column wider or you could even display a separate table below the main table, something like:


image

File: tutorial/core_component_examples.py [hunk 4]
```diff
@@ -0,0 +1,1005 @@
...
+    table = html.Table(
+            [html.Tr([html.Th(col, style={'text-align': 'center'}) for col in
+                     dataframe.columns])] + rows)
+
+    return table
+
+
+Dropdown = html.Div(children=[
+
+    html.H3('Dropdown Extra Examples'),
```
chriddyp: `Dropdown Examples and Reference`

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+    html.Hr(),
+    html.H4('Searching Dropdown'),
```
chriddyp: This should be `Option Search and Custom Placeholder`

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+    dcc.SyntaxHighlighter('''import dash_core_components as dcc
+
+dcc.Dropdown(
+    options=[
+        {'label': 'New York City', 'value': 'NYC'},
+        {'label': 'Montreal', 'value': 'MTL'},
+        {'label': 'San Francisco', 'value': 'SF'}
+    ],
+    searchable=True,
+    clearable=False,
```
chriddyp: This option doesn't reflect the example name, so we should remove it

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+                {'label': 'San Francisco', 'value': 'SF'}
+            ],
+            searchable=True,
+            clearable=False,
+            placeholder='Try typing: "New York"'
+    ),
+
+    html.Hr(),
+    html.H4('Clearing Dropdown'),
+    html.P("Try selecting New York or San Francisco"),
```
chriddyp: I don't quite get this example. `clearable=True` is the default option, and it affects whether there is an `x` or not in the dropdown. Clicking this `x` sets the value to `None`

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+    options=[
+        {'label': 'New York City', 'value': 'NYC'},
+        {'label': 'Montreal', 'value': 'MTL'},
+        {'label': 'San Francisco', 'value': 'SF'}
+    ],
+    searchable=True,
+    clearable=False,
+    placeholder="Try typing: "New York"
+)
+
```
chriddyp: I think we should start out with a very simple example that demonstrates what the "default options" are and how you use it in practice. Something like:

"Default Dropdown"

This is the minimal set of options used to construct a Dropdown. By default, the options in the Dropdown are searchable and by default you can click on the `x` next the options to set the `value` to `None.

```
import dash
import dash_core_components as dcc

app = dash.Dash()
app.layout = html.Div([
    dcc.Dropdown(
        id='my-dropdown',
        options=[
            {'label': 'New York City', 'value': 'NYC'},
            {'label': 'Montreal', 'value': 'MTL'},
            {'label': 'San Francisco', 'value': 'SF'}
        ],
       value='NYC'
    ),
    html.Div(id='output-container')
)

@app.callback(
    dash.dependencies.Output('output-container', 'children'),
    [dash.dependencies.Input('my-dropdown', 'value')])
def update_output(value):
    return 'You have selected "{}"'.format(value)

if __name__ == '__main__':
    app.run_server(debug=True)
```

And then, the rest of the examples can be the simple declaration of the `Dropdown` and the rest of the examples can specify just the non-default properties like:
- `clearable=False`
- `placeholder="Select a city"`
- `searachable=False`
- `disabled=True`
- `multi=True` - This example should set `value=['MTL', 'NYC']` as its initial value so that users can see how this works
- `disabled` inside the options

chriddyp: And a similar approach should be taken across the rest of the components: 
- The first example is simple and contains a self-contained dash app with a short description on the features of the component
- The rest of the examples demonstrate the non-default options of the component

tutorial/core_component_examples.py
chriddyp: and similarly across the rest of the examples

tutorial/core_components.py
```diff
@@ -55,9 +56,11 @@
         {'label': 'Montréal', 'value': 'MTL'},
         {'label': 'San Francisco', 'value': 'SF'}
     ], multi=True, value="MTL", id='section2-dropdown-2'),
-
+    html.Br(),
+    dcc.Link(html.A('More Dropdown Examples'),
```
chriddyp: `More Dropdown Examples and Reference` (and similarly across the rest of the components)

tutorial/core_component_examples.py
chriddyp: Also, this can be an `H2`

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+    generate_table(get_dataframe('Graph'))
+])
+
+
+# DatePickerRange
+DatePickerRange = html.Div(children=[
+    html.H3("DatePickerRange Extra Examples"),
+    html.Hr(),
+    html.H4('Calendar Orientation'),
+    dcc.Markdown('Calendar can either be displayed vertically or horizontally'),
```
chriddyp: Let's do complete sentences for these. So: `The calendar can be displayed vertically or horizontally.` (period at the end)

chriddyp: Maybe it should read: 

> By default, the calendar is displayed horizontally. The calendar can be displayed vertically be setting `calendar_orientation='vertical'`

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+    dcc.SyntaxHighlighter('''import dash_core_components as dcc
+from datetime import datetime as dt
+
+dcc.DatePickerRange(
+    id='date-picker-range',
+    calendar_orientation='vertical',
+    start_date_placeholder_text='Test out',
```
chriddyp: `start_date_placeholder_text` isn't related to this particular example, so let's remove this option

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+    dcc.DatePickerRange(
+        id='date-picker-range',
+        calendar_orientation='vertical',
+        start_date_placeholder_text='Test out the',
+        end_date_placeholder_text='calendar!'
+    ),
+    html.Hr(),
+    html.H4('Show outside days'),
+    dcc.Markdown('To show days outside the curret calender month,\
+                 `show_outside_days` needs to be set to True'),
```
chriddyp: Instead of "needs to be", maybe
> By default, the calendar does not display dates outside of the current calendar month. To display days outside of the current calendar month, set `show_outside_days` to `True`.

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+    dcc.DatePickerRange(
+        id='date-picker-range',
+        show_outside_days=False,
+        start_date_placeholder_text='Without',
+        end_date_placeholder_text='Outside Days',
+    ),
+    html.Hr(),
+    html.H4('Display Formats'),
+    dcc.Markdown('You can show months in a variety of display formats.\
+                  Display formats denote how, your selected dates will look'),
```
chriddyp: ⚡️  the comma. Also, it seems like this relates to "dates" not just "months". I also find the syntax for specifying these dates pretty confusing. Instead, maybe something like:
> Dates may be displayed in a variety of formats through the `display_format` attribute. These dates are specified through a combination year, month, and day tokens.

| Input | Example | Description |
| ----------------------------- |
| YYYY | 2014 | 4 or 2 digit year |
| YY | 14 | 2 digit year |
| Y | -25 | Year with any number of digits and sign |
| Q | 1..4 | Quarter of year. Sets month to first month in quarter. |
| M MM | 1..12 | Month number |
| MMM MMMM | Jan..December | Month name in locale set by moment.locale() |
| D DD | 1..31 | Day of month |
| Do | 1st..31st | Day of month with ordinal |
| DDD DDDD | 1..365 | Day of year |
| X | 1410715640.579 | Unix timestamp |
| x | 1410715640579 | Unix ms timestamp |

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+    display_format='MMMM, D, Y',
+    end_date_placeholder_text='MMMM, D, Y'
+),
+
+dcc.DatePickerRange(
+    id='date-picker-range',
+    start_date=dt(2017, 8, 5),
+    display_format='MM || DD || Y',
+    end_date_placeholder_text='MM || DD || Y'
+)
```
chriddyp: Let's just include examples that would be commonly used. So:
- MM DD, YYYY
- MMMM D, Y
- YYYY-MM-DD
- DD/MM/YYY

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+        id='date-picker-range',
+        start_date=dt(2017, 8, 5),
+        display_format='MM || DD || Y',
+        end_date_placeholder_text='MM || DD || Y'
+    ),
+    html.Hr(),
+    html.H4('Month Formats'),
+    dcc.Markdown('This prop determines how the month formats will be\
+                 displayed. Click on the date pickers below to see the\
+                 results.'),
```
chriddyp: I find "Month formats" to be confusing. Maybe instead something like:
> The `month_format` property sets the format of the label of the date inside the date picker pop-up. The syntax is the same as the `display_format` above.

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+
+    dcc.DatePickerRange(
+        id='date-picker-range',
+        start_date_placeholder_text='Click',
+        end_date_placeholder_text='me!',
+        month_format='MMMM || DD || Y'
+    ),
+    html.Hr(),
+    html.H4('Min, Max and Initial Visible Month'),
+    dcc.Markdown('The `min_date_allowed` prop determines the minimum\
```
chriddyp: while `prop` is common inside `react`-land, it's not as common in Python land. We should use `property` or `attribute` instead.

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+                 selectable date on the calendar, the `max_date_allowed`\
+                 determines the maximum selectable date on the calendar'),
```
chriddyp: missing an explanation of what the `initial_visible_month` does. Maybe
> The `min_date_allowed` property sets the minimum selectable date on the calendar, `max_date_allowed` sets the maximum selectable date, and `initial_visible_month` sets the initial visible month of the calendar.

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+
+''', language='python', customStyle=styles.code_container),
+    dcc.DatePickerRange(
+        id='date-picker-range',
+        min_date_allowed=dt(2017, 8, 5),
+        max_date_allowed=dt(2017, 8, 27),
+        initial_visible_month=dt(2017, 8, 1)
+    ),
+    html.Hr(),
+    html.H4('Clearable Property'),
```
chriddyp: Just `Clearable` to match the style of the rest of the headers

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+    dcc.SyntaxHighlighter('''import dash_core_components as dcc
```
chriddyp: missing a description here like the rest of the examples

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+from datetime import datetime as dt
+
+dcc.DatePickerRange(
+    id='date-picker-range',
+    min_date_allowed=dt(2017, 8, 5),
+    max_date_allowed=dt(2017, 8, 27),
+    start_date=dt(2017, 8, 5),
+    end_date=dt(2017, 8, 20),
+    initial_visible_month=dt(2017, 8, 1)
```
chriddyp: i don't quite understand this example - should there be a `clearable` property in here?

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+    ),
+
+
+    html.Hr(),
+    html.H4("Dropdown Proptypes"),
+    generate_table(get_dataframe('Dropdown'))
+])
+
+
+# RangeSlider
```
chriddyp: `Slider`

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+Slider = html.Div(children=[
+    html.H3('RangeSlider Extra Examples'),
```
chriddyp: ~RangeSlider~ `Slider Examples`

tutorial/core_component_examples.py
```diff
@@ -0,0 +1,1005 @@
...
+    html.Hr(),
+    html.H4('Slider Dots'),
```
chriddyp: I'd focus this section on:
- Examples of different `marks` - The temperature examples here are really good: https://ant.design/components/slider/
  - Examples of styling different marks
  - Examples of different number lines (linear, non-linear)
- Relationship between `step` and `marks` (`step=None` but `marks` being defined)
- Relationship between `included` and `marks`
- Simple example of `min`, `max`, and `step`
- Remove `vertical` for now if it doesn't work

chriddyp: And then the same thing for `RangeSlider` except an example of multiple values and `crossable` (or w/e/ it is)

tutorial/core_components.py
```diff
@@ -357,6 +397,9 @@
     quotes, and more.'''.replace('  ', ''),
     containerProps={'className': 'example-container'}),
 
+    html.Br(),
+    dcc.Link(html.A('More Markdown Properties'),
```
chriddyp: I'd update this to include some markdown examples. See https://guides.github.com/features/mastering-markdown/ for a good set of examples

tutorial/core_components.py
```diff
@@ -400,14 +443,16 @@
         id="my-graph"
     ),
 
-    html.Div(style={'marginBottom': 50}),
+    html.Br(),
+    dcc.Link(html.A('More Graph Examples'),
+             href="/dash/dash-core-components/graph"),
```
chriddyp: Let's remove this section since it will never be as complete as plot.ly/python. We can just change this link to be 
> View the [plotly.py](https://plot.ly/python) documentation.

tutorial/core_components.py
```diff
@@ -400,14 +443,16 @@
...
 
     html.Div(id='hidden', style={'display': 'none'})
 ])
```
chriddyp: let's also add a section for `html.Button`'s `n_clicks` property (with an example callback and typical usage with `State`)

tutorial/run.py
```diff
@@ -176,6 +177,66 @@ def create_contents(contents):
...
+    },
+
+    'link-examples': {
+        'url': '/dash/dash-core-components/link',
+        'content': examples.Link
+    },
+
+    'textarea-examples': {
+        'url': '/dash/dash-core-components/textarea',
+        'content': examples.Textarea
```
chriddyp: 👍  this organization and these urls look good

alishobeiri:
@chriddyp Let me know what you think or if I should make any changes. 

I'll start working on fixing the tables now

chriddyp:
I'll start working on this now and bring it to production. Thanks for kicking this off @alishobeiri !

PR #26, opened 2017-10-02
update versions on authentication pg

cldougl:
cc @chriddyp


chriddyp:
great! can you also update `requirements.txt` with the latest? https://github.com/plotly/dash-docs/blob/master/requirements.txt

PR #31, opened 2017-11-09
Use suppress spelling instead of supress for the guide

jtpio:
Since the typo has been fixed in Dash and both versions are handled for backward compatibility (https://github.com/plotly/dash/pull/131), should we use the right spelling for the guide?


chriddyp:
Thanks @jtpio !!

PR #32, opened 2017-12-04
Follow PEP 8 standards.  Remove unnecessary parens - (). Surround top-level function and class…

jbampton:
… definitions with two blank lines.  https://www.python.org/dev/peps/pep-0008/#blank-lines  Use single quotes consistently. Remove duplicate whiteSpace key from Python dictionary. Indent Python code.  Standard import "from datetime import datetime as dt" should be placed before "import dash" (wrong-import-order). Standard import "import time" should be placed before "import dash" (wrong-import-order). Standard import "import math" should be placed before "import dash" (wrong-import-order). Standard imports "import json" and "from textwrap import dedent as d" should be placed before "import dash" (wrong-import-order). Standard import "import datetime" should be placed before "import dash" (wrong-import-order)

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -45,12 +45,12 @@ with open('tutorial/examples/getting_started.py', 'r') as f:
 
     # return the layout instead of assigning it to the global app
     if 'app.layout = ' not in example:
-        raise Exception("app.layout not assigned")
+        raise Exception('app.layout not assigned')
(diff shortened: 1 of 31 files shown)


chriddyp:
Looks good. Thanks @jbampton ! These changes will be automatically deployed once the PR merges 🚀

PR #42, opened 2017-12-15
Standard imports "import json" and "import os" should be placed befor…

jbampton:
…e "import dash_html_components as html". Remove unused variable "js_type_name".  PEP 8 - Comparison with None should use "is not None".  Remove unused import. Fix indentation. 2 blank lines between top level functions.


chriddyp:
thanks @jbampton !

PR #51, opened 2018-01-11
Fixes the 'Chapter' function to output the correct type of link: html…

jbampton:
….A.  With this update the links on the dash docs homepage page now link correctly to the top of the page that opens when clicked.  Previously if you were on the homepage https://plot.ly/dash/ and clicked on 'URL Routing and Multiple Apps' it would open the next page and scroll the browser down missing the intended target heading.  This also allows the link to show up in the status bar at the bottom of the web browser. Also PEP8 fix by adding an extra line.  Added the extra anchor '#chapter' to href.  In Firefox when you scroll to the bottom of the page you can hit the url again and it will scroll back to the top of the page where the id="chapter" or page heading is defined.

### DIFF
--- a/css/base.css
+++ b/css/base.css
@@ -19,7 +19,7 @@
 –––––––––––––––––––––––––––––––––––––––––––––––––– */
 /* plotly.js's modebar's z-index is outta the roof by default - lower it down */
 .js-plotly-plot .plotly .modebar {
-    z-index: 0 important!;
+    z-index: 0 !important;
 }
 
 /* Grid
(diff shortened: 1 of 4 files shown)


chriddyp:
I would actually like to keep `dcc.Link` and fix the underlying issue with scroll in that component https://github.com/plotly/dash-core-components/issues/86. Without `dcc.Link`, each page will load _very slowly_. Keeping the sort of buggy behaviour with `dcc.Link` in the user guide is more motivation for me and the community to get https://github.com/plotly/dash-core-components/issues/86 sorted out.

PR #58, opened 2018-01-25
adding search

cldougl:
initial addition


cldougl:
@chriddyp this should be ready if you want to take a look

dash_search_index.py
```diff
@@ -0,0 +1,32 @@
+## Update dash-doc search index: https://www.algolia.com/apps/7EK9KHJW8M/explorer/browse/dash_docs
+
+import json
+import os
+
+from algoliasearch import algoliasearch
```
chriddyp: can you add this to `requirements.txt`? I believe that is why deployment and tests are failing

dash_search_index.py
```diff
@@ -0,0 +1,32 @@
...
+from tutorial.chapter_index import chapters
+
+## Algolia Credentials
+client = algoliasearch.Client("7EK9KHJW8M", os.environ['ALGOLIA_API_KEY'])
```
chriddyp: is this added to heroku? we can also add this to percy.io and include nice integration screenshot tests

dash_search_index.py
```diff
@@ -0,0 +1,32 @@
...
+for chapter in chapters:
+    chap = {}
+    chap['name'] = chapters[chapter]['name']
+    chap['permalink'] = 'https://plot.ly'+chapters[chapter]['url']
+    chap['description'] = chapters[chapter]['description']
+    if chapter not in ['getting-started', 'getting-started-part-2', 'graphing']:
+        chap_content = str(chapters[chapter]['content'])
+        chap_content = chap_content.replace("'", "")
+        chap_content = chap_content.replace('"', '')
+        chap_content = chap_content.replace('\\n', '')
```
chriddyp: let's make this one line: `chap_content.replace("'", '').replace('"', '').replace('\\n', '')`

chriddyp:
this looks good to me. let's also add a quick integration test where we visit the page, and search `"dropdown"`, and then take a percy screenshot. 

here are some steps to do this:
1. create a new function in here: https://github.com/plotly/dash-docs/blob/90bb98660d2c3dc4a8ca4ae1c6b5171b21ae1bad/tests/test_integration.py#L64 called `test_search(self):`
2. visit the search page and wait for it to load
```python
        self.startServer(app, '/dash/search')

        try:
            self.wait_for_element_by_id('wait-for-layout')
        except Exception as e:
            print(self.wait_for_element_by_id(
                '_dash-app-content').get_attribute('innerHTML'))
            raise e
```
3. once loaded, locate the search input by its ID (you will need to add an `id` to the `dcc.Input` if you havent already):
```
search_element = self.driver.find_element_by_id('search')
```
4. add some characters to the input
```
search_element.clear()
search_element.send_keys('dropdown')
```
5. take a screenshot
```
self.snapshot('search-dropdown')
```

you can run these tests locally with
```
pip install -r requirements.txt
python -m unittest tests.test_integration
```
you'll also need to add the environment variable to circle: https://circleci.com/gh/plotly/dash-docs/edit#env-vars

dash_search_index.py
chriddyp: nvm we don't need to add it to percy.io, but we do need to add it to circle

dash_search_index.py
cldougl: added- I don't think that's why deployment/tests are failing

dash_search_index.py
cldougl: the api key is added to circle now though this script shouldn't be run anywhere at this point (it updates the index on algolia, not run on each search)

chriddyp:
looks like we're also missing `dash_dangerously_set_inner_html` is missing from requirements:
```
>>> from tutorial.run import app
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "tutorial/run.py", line 19, in <module>
    import chapter_index
  File "tutorial/chapter_index.py", line 17, in <module>
    import search
  File "tutorial/search.py", line 4, in <module>
    import dash_dangerously_set_inner_html
ImportError: No module named dash_dangerously_set_inner_html
```

tutorial/server.py
```diff
@@ -11,7 +11,7 @@
 server = Flask(__name__, static_url_path='/dash/static', static_folder='./static')
 server.secret_key = os.environ.get('secret_key', 'secret')
 app = Dash(__name__, server=server, url_base_pathname='/dash/', csrf_protect=False)
-app.css.config.serve_locally = False
-app.scripts.config.serve_locally = False
+app.css.config.serve_locally = True
+app.scripts.config.serve_locally = True
```
chriddyp: let's actually keep it as False now, since we're including remote JS assets

chriddyp:
nicely done @cldougl 👏 👏 👏 👏

PR #60, opened 2018-01-26
1st draft at onprem troubleshooting section

cldougl:
re https://github.com/plotly/streambed/issues/9934#issuecomment-345374688
cc @chriddyp


chriddyp:
fantastic, thanks!

PR #77, opened 2018-03-08
japanese translation

chriddyp:
proof of concept architecture for supporting japanese language translation.

see discussion here: https://community.plot.ly/t/can-i-translate-english-tutorial-into-japanese-one/8859

### DIFF
--- a/tutorial/getting_started_part_1.py
+++ b/tutorial/getting_started_part_1.py
@@ -22,324 +22,352 @@ examples = [
 ]
 
 
-layout = html.Div([
-
-    dcc.Markdown('''
-    # Dash Tutorial - Part 1: App Layout
-
-    This tutorial will walk you through the fundamentals of creating Dash apps
-    through {} self-contained apps.
-
-    '''.format(len(examples)).replace('    ', '')),
-
-    dcc.Markdown('''***
-
-1. [Installation](#installation)
-2. [Dash App Layout](#dash-app-layout)
-    - Generating HTML with Dash
-    - Data Visualization in Dash
-    - Markdown
-    - Core Components
-    - Calling `help`
-3. [Interactivity](/getting-started-part-2)
-    - Fundamentals
-    - Multiple Inputs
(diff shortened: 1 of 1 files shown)


Marc-Andre-Rivet:
@chriddyp Reviving this one. Often times the i18n/l10n support involves taking out the actual text from the page and creating resource files that can be translated / worked on independently from the presentation / application structure.

A very JS-centric links but the underlying ideas should apply here: http://i18njs.com/, https://www.i18next.com/

Handling branching (e.g. window vs non-window, R bs py, etc.) can be tricky but is typically handled relatively well with this type of approach / fragmentation.

What we want the granularity to be is also a good question. Large chunk of Markdown documentation might not be the easiest level to work on. Breaking it up could be counter-intuitive too. My experience is more towards supporting this in apps -- mostly single words and at a few sentences / a paragraph.

Pulling in @byronz for additional insight.
Pulling in @rpkyle as he's in the initial stages of creating the R documentation structure and baking in i18n/l10n at this point is much easier than retrofitting it.

rpkyle:
> @chriddyp Reviving this one. Often times the i18n/l10n support involves taking out the actual text from the page and creating resource files that can be translated / worked on independently from the presentation / application structure.

Marc-André and I chatted a bit about this yesterday. Given that we're just starting the DashR docs, and that creating the resource files should not be a tremendous amount of work, I'm willing to help with this/test however I can.

byronz:
> > @chriddyp Reviving this one. Often times the i18n/l10n support involves taking out the actual text from the page and creating resource files that can be translated / worked on independently from the presentation / application structure.
> 
> Marc-André and I chatted a bit about this yesterday. Given that we're just starting the DashR docs, and that creating the resource files should not be a tremendous amount of work, I'm willing to help with this/test however I can.

we can discuss this in person today, I had some little experiments yesterday, and let's talk more in depth

alexcjohnson:
Stale, closing.

PR #82, opened 2018-03-26
improve import order across all examples

ned2:
This improves the import order across examples in the Dash guide. 

Firstly, imports are now PEP8 compliant. As a reference document that is likely to be read by many people new to Python, I think it's desirable to model best practice. PEP8 compliant imports are also easier to read quickly.

Secondly, this also moves `from X import Y` style imports to below `import X` style imports. While this is not  required by PEP8, it's a convention that many adhere to, again, for readability. Additionally, it has the useful effect of putting the imports of both `dash_html_components` and `dash_core_components` before doing `from dash.dependencies import Input, State, Event`. Combined with the changes from plotly/dash-core-components#177 and plotly/dash-html-components#39 this will result in users getting a more helpful error message when they name the their script `dash.py`, which will fail due to shadowing the Dash package itself.

### DIFF
--- a/tutorial/auth.py
+++ b/tutorial/auth.py
@@ -75,8 +75,8 @@ layout = html.Div([
 
     dcc.SyntaxHighlighter('''import dash
 import dash_auth
-import dash_html_components as html
 import dash_core_components as dcc
+import dash_html_components as html
 import plotly
 
 # Keep this out of source code repository - save in a file or a database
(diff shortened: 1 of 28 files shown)


ned2:
Just a note to emphasise (in case it wasn't clear in the initial description): this PR will make the better handling of `dash.py` files in the user's working directory that was recently merged into dash-core-components and dash-html-components be triggered much more frequently. Currently, the import order in the docs will still trigger the more arcane and unhelpful exception due to the `from dash.dependencies import Input, State` line occurring first.

chriddyp:
This looks good to me! It would be great to enforce PEP8 in this repo with some lint tests (e.g. https://github.com/plotly/dash/blob/master/tox.ini#L26-L27) but that doesn't need to happen in this PR 😺

PR #84, opened 2018-03-27
Filestystem sessions example

chriddyp:


chriddyp:
This PR documents a pattern that I've been sharing a lot in the Dash community forum but haven't yet documented. It's another method for sharing state between callbacks, using the filesystem like caching but keeping things session-specific like the 'hidden div' method.

Originally discussed in https://community.plot.ly/t/capture-window-tab-closing-event/7375/2 and in private email conversations with a customer. Subsequently referenced in:
- https://community.plot.ly/t/apache-arrow-and-dash-community-thread/7381
- https://community.plot.ly/t/append-to-a-json-list-stored-in-a-hidden-div/8909/9
- https://community.plot.ly/t/upload-component-file-uploaded-multiple-times/7590
- https://community.plot.ly/t/saving-changes-in-table/8613/7
- https://community.plot.ly/t/fast-way-to-share-data-between-callbacks/8024
- https://community.plot.ly/t/dynamic-options-for-drop-down/7550/4
- https://community.plot.ly/t/fast-way-to-share-data-between-callbacks/8024/2

@plotly/dash - please review. In particular, @charleyferrari @nicolaskruchten @bpostlethwaite .

also @ned2 , thought you might be interested in reviewing as well.

chriddyp:
You can view the review app here: https://dash-docs-pr-84.herokuapp.com/sharing-data-between-callbacks (see the "View deployment" button above)

ned2:
Yeah it'd be great to have a user-session caching example in the guide. I like the example. A couple of thoughts about the initial discussion:

The title and text refers to the use of the filesystem. While this could be done with the filesystem (as the comment "You can also save to the filesystem." indicates) the actual example doesn't use it, and to certain extent the choice of cache type is not really the point of the example. As it is, I feel like the description is a little misleading, as people may be lead to believe that the use of the filesystem is necessary for this approach to user-based sessions.

What about changing the title to "User-based Sessions" or something similar and making the text more clearly indicate that this example is about user sessions which can be realised through a cache, whether it be the filesystem or Redis. Then also perhaps changing the example to make the `CACHE_TYPE` a filesystem one, since this is going to to be the path of least friction for most users.

chriddyp:
or maybe "User-Based Session Data on the Server"?

ned2:
Or "Server-side User Sessions"?

chriddyp:
> Or "Server-side User Sessions"?
ah yeah, that's way better 👍

tutorial/sharing_state.py
```diff
@@ -456,6 +469,67 @@ def update_graph_4(value):
...
+        The previous example cached computations on the filesystem and
+        those computations were accessible for all users.
+
+        In some cases, you want to keep the data isolated to user sessions:
+        one user's derived data shouldn't update the next user's derived data.
+        One way to do this is to save the data in a hidden `Div`,
+        as demonstrated in the first example. 
+
+        Another way to do this is to save the data on the 
+        filesystem cache with and reference the data using
```
cldougl: this sentence seems a little confusing/incomplete to me: 
> save the data on the filesystem cache with and reference the data using a session id.

tutorial/sharing_state.py
```diff
@@ -456,6 +469,67 @@ def update_graph_4(value):
...
+        a session id. In this method, since data is saved on the server,
+        instead of transported over the network, it is generally faster than the 
+        "hidden div" method.
+
 
+        This example was originally discussed in a
+        [Dash Community Forum thread](https://community.plot.ly/t/capture-window-tab-closing-event/7375/2?u=chriddyp).
+
+        This example:
+        - Caches data using the `flask_caching` filesystem cache. You can also save to an in-memory database like Redis..
```
cldougl: I would remove the .. and just have 
> like Redis.

tutorial/sharing_state.py
```diff
@@ -456,6 +469,67 @@ def update_graph_4(value):
...
+            alt='Example of a Dash App that uses User Session Caching'
+        )
+    ),
+
+    dcc.Markdown(s('''
+        There are three things to notice in this example:
+        - The timestamps of the dataframe don't update when we retrieve
+        the data. This data is cached as part of the user's session.
+        - Retrieving the data initially takes 5 seconds but successive queries
+        are instance, as the data has been cached.
```
cldougl: I think instance should be instant

nicolaskruchten:
I like this. It highlights a good variant on the existing cache/signal system and the existing hidden-div system. This is similar to the way PHP session IDs used to work 15 years ago or so. Any chance of getting standard cookie-based session IDs in the framework someday? :) They would benefit from encryption, expiry rules etc...

I can't think of an attack right now but I wonder if this kind of thing isn't vulnerable to https://en.wikipedia.org/wiki/Session_fixation or similar exploits. We may want to add a big warning that this isn't necessarily a secure way of doing sessions.

PR #93, opened 2018-04-18
Change academy link

nicolaskruchten:
The React tutorial on academy.plot.ly is quite old and shows a very clunky way to use plotly.js with React so we're sunsetting that microsite.


bpostlethwaite:
:dancer: from me!

nicolaskruchten:
@chriddyp should I just go into Percy and approve these or is that something you do or... ?

chriddyp:
@ feel free to go into percy and approve and then merge

PR #104, opened 2018-06-14
Add section on configuring OnPremise system dependencies

bpostlethwaite:
@cldougl 

I think you will be able to see what it looks like in the Percy snapshots... if we test the OnPremise section. If not I can post a screenie here.

### DIFF
--- a/tutorial/on_premise_deployment.py
+++ b/tutorial/on_premise_deployment.py
@@ -371,16 +371,18 @@ def generate_instructions(chapter, platform):
 
             You may include a pre-deploy script that executes in
             your Dash App's environment. For the case of adding an
-            ODBC driver we may want to copy ODBC initialization files
-            from our application folder into system paths where the
-            ODBC driver expects them to be.
+            ODBC driver we need to add ODBC initialization files into
+            the correct systems paths. To do so we include the ODBC
+            initialization files in the application folder and then
+            copy them into system paths in the pre-deploy script.
 
             ##### Add A Pre-Deploy Script
             Let's generate a file to do this. Note that the file can
             have any name as we must specify the name in an application
             configuration file `app.json`.
             For the purposes of this example we assume we have
-            named it `setup_pyodbc`.
+            named it `setup_pyodbc` and installed it in the root of our
+            application folder.
 
             ''')),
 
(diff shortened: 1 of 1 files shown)


bpostlethwaite:
Yep Percy screeny for the new section is here:
https://percy.io/plotly/dash-docs/builds/79160

1 diff change is curious, Radio items for "canada" and "america" swapped places even though they have nothing to do with the files I modified. I'll take a look.

bpostlethwaite:
Oh there is a View Deployment option for each PR. Nice!

chriddyp:
> Radio items for "canada" and "america" swapped places even though they have nothing to do with the files I modified


don’t worry about this - the example uses dicts which are unordered and occasionally swap places

tutorial/on_premise_deployment.py
```diff
@@ -338,6 +341,82 @@ def generate_instructions(chapter, platform):
...
+            unixodbc-dev
+            '''), customStyle=styles.code_container, language="text"),
+
+            dcc.Markdown(s('''
+
+            #### Configure System Dependencies
+
+            You may include a pre-deploy script that executes in
+            your Dash App's environment. For the case of adding an
+            ODBC driver we may want to copy ODBC initialization files
```
cldougl: I would probably take at that "may" cause this isn't really an optional step for this case

tutorial/on_premise_deployment.py
```diff
@@ -338,6 +341,82 @@ def generate_instructions(chapter, platform):
...
+            from our application folder into system paths where the
+            ODBC driver expects them to be.
+
+            ##### Add A Pre-Deploy Script
+            Let's generate a file to do this. Note that the file can
+            have any name as we must specify the name in an application
+            configuration file `app.json`.
+            For the purposes of this example we assume we have
+            named it `setup_pyodbc`.
```
cldougl: should mention this is in the root directory of the application

tutorial/on_premise_deployment.py
```diff
@@ -338,6 +341,82 @@ def generate_instructions(chapter, platform):
...
+            '''), customStyle=styles.code_container, language='json'),
+
+            dcc.Markdown(s('''
+            ***
+
+            Now when the application is deployed it will install the apt
+            packages specified in `apt-packages` and run the setup file
+            specified in `app.json`. In this case it allows us to install
+            and then configure the ODBC driver.
+            '''))
```
cldougl: I would add a link to the PR examples [pyodbc example](https://github.com/plotly/dash-on-premise-sample-app/pull/3#issue-144272510) as well (thus far it was useful for most of the users to work off of that example as some of them are unfamiliar with deployment processes)

bpostlethwaite:
@cldougl ok ready for another review thanks!

bpostlethwaite:
Actually looks like you have given me a :dancer:  - ok I'll approve the Percy shots and merge.

cldougl:
yep- nothing blocking :) the changes look good to me 🎉

PR #111, opened 2018-07-03
Use prerelease 0.23.1rc1 to test out Link component

valentijnnieman:
@chriddyp Let's test out that Link component :-)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ click==6.7
 configparser==3.5.0
 dash==0.21.1
 dash-auth==1.0.0
-dash-core-components==0.23.0
+dash-core-components==0.23.1rc1
 dash-dangerously-set-inner-html==0.0.1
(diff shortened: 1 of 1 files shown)


valentijnnieman:
These tests are failing too - could it be related to https://github.com/plotly/dash-core-components/pull/224 ?

chriddyp:
i just redeployed the test app, it's available here:  https://dash-docs-pr-111.herokuapp.com/

chriddyp:
I'm looking into the tests now.

chriddyp:
Hm, tests are passing locally.
```
python -m unittest tests.test_integration.Tests
.......
Ran 1 test in 170.539s

OK
```

It must just be some timeout.

chriddyp:
OK, rebuilding works, so it looks like this wasn't an issue with the code, just a timeout issue.

Of course, we should try to make this less susceptible to timeouts :)

PR #112, opened 2018-07-06
have multiple remotes for multiple dash apps

gzork:
If a user deploys/work on multiple dash apps, they can't rename all remotes related to the apps with the same name, so we can have a remote for each dash app that the dev works on.

I suggest the name plotly-[dash-app-name]

### DIFF
--- a/tutorial/on_premise_deployment.py
+++ b/tutorial/on_premise_deployment.py
@@ -255,7 +255,7 @@ def generate_instructions(chapter, platform):
             ''')),
 
             dcc.SyntaxHighlighter(s('''$ cd dash-on-premise-sample-app
-                $ git remote add plotly dokku@your-dash-app-manager:your-dash-app-name'''),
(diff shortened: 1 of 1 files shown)


gzork:
@cldougl @chriddyp do you guys think this should make it to master ?

chriddyp:
Yeah, I agree no for now, since this is an area of confusion and frustration already.

In the future, I think that we could put this under an "advanced section" where we could talk about a more concrete use case: having a "Testing" app and a "Production" app and deploying to these different apps with different remotes

gzork:
I just changed one instruction, I didn't add any, but maybe it's not worth it as they won't need to do it most of the time.. I'll go ahead and cancel this PR

cldougl:
but thanks for making the pr in general @gzork cause it's really helpful to get more opinions on these docs to try and make them more user-friendly!!!

Issue #115, opened 2018-07-09
A "React for a Python Developer" guide


chriddyp:
I think that we should do whatever we can to reduce the barriers of entry in creating custom Dash components. One of the largest barriers to entry is React and JavaScript itself: our community is (by definition) mostly Python developers.

One way that we can help out our community in creating or customizing Dash components is by providing a really friendly, step-by-step guide to creating a Dash component. I'll call this the "React.js for the Python Developer Guide".

Below is a rough outline of what I'm imagining. Would love feedback on this!

***

For Tutorial 1, here's a rough syllabus that I've used in the workshops:
1. Introduction:
    - React and JavaScript is easier than it seems, you can do it!
    - Most Dash component's are simple wrappers around existing components. And there are hundreds of thousands of them.
2. Developer Setup
    - Installing Node, NPM, and `create-react-app`
        - In practice, it would be nice to use our `dash-component-archetype` here once we make it better
3. Making your first change
    - Opening up `App.js` and changing "React" to "Dash"
    - Observing the hot-reload
    - Making an error and seeing where to find error messages: `throw error`
    - Doing `console.log` and seeing the data
4. Quick React vs Python:
    - React has classes's like Python classes
    - Declare variables in JavaScript: `const` and `let`
    - JavaScript "objects" are Python's "dict"
        - Except the keys are always strings
    - `console.log` instead of `print`
    - `import` and `export`
5. Quick create your own component in the same file as `App.js`:
    1. Create `Header` and include it in `App`:
    ```
    class Header extends Component {
        render() {
            return 'Header'
        }
    }

    class App extends Component {
        ...

        render() {
            ...
            return <Header/>
        }
    }
    ```

    2. Thread a property through:
    ```
        return <div>{this.props.name}</div>
    ```

    ```
        const myHeaderName = 'Dash Header'
        return <Header name={myHeaderName}/>
    ```

    3. Add another property like style:
    ```
        return <div style={this.props.style}>{this.props.name}</div>
    ```

    ```
        const myHeaderName = 'Dash Header';
        const style={'color': 'blue', 'fontSize': 20}
        return <Header name={myHeaderName}/>
    ```
6. Making this component Dash-ready:
    - Quick pause: this "component" is hydrated from JavaScript, now let's make it a standalone Dash component
    - Move it to a new file
    - Add `propTypes`
    - Run the toolchain
    - Fire up `usage.py` and have:
    ```
    app.layout = html.Div([
[...]

PR #116, opened 2018-07-16
React for python devs

valentijnnieman:
Here's a first draft (incomplete!) of what I've been working on. Thought it would be good to create a work-in-progress PR so we can all give some feedback on it! It's going to be quite long as it is, so some feedback on what to cut/change would be great! Ref for #115


tutorial/react_for_python.py
```diff
@@ -5,4 +5,346 @@
...
+  // other imports
+  import PropTypes from 'prop-types';
+
+  class TextInput extends Component {
+    // class definition here
+  }
+
+  TextInput.propTypes = {
+    /** 
+    * This comment will also get parsed by Dash, so that it is available in help(component.TextInput)
```
Bachibouzouk: This line goes beyond the text area in the rendered version

tutorial/react_for_python.py
```diff
@@ -5,4 +5,346 @@
...
+      })
+    }
+    render() {
+      return <input value={this.props.value} onChange={this.handleInputChange}/>
+    }
+  }
+
+  TextInput.propTypes = {
+    /** 
+    * This comment will also get parsed by Dash, so that it is available in help(component.TextInput)
```
Bachibouzouk: ibid

xhlulu:
This is great; I was planning to make a component for a while and this is definitely going to be the changing factor for me to finally make it :) I was just wondering; for very large markdown text, would it be better to write it in a markdown document first, then import it inside the python file? I am not sure if that is standard practice.

tutorial/react_for_python.py
```diff
@@ -0,0 +1,358 @@
...
+  you too can use it to write applications for the web. That being said, all unfamiliar technology will have a learning curve, but with practice
+  and patience you will master it!
+
+  Dash uses React under the hood to render the user interface you see when you load a web page created with Dash. Because React allows you to write
+  your user interface in encapsulated components that manage their own state, it is easy to split up parts of code for Dash too - at the end of this
+  tutorial, you will see that Dash components and React components map 1 to 1! For now, it is important to know that Dash components are 
+  mostly just simple wrappers around existing React components. This means the entire React ecosystem is potentially usable in a Dash application!
+
+  ## Installing everything you need
+  Let's start by setting up our JavaScript development environment. We will use Node.js, NPM, and `create-react-app` to write our first React
```
nicolaskruchten: There is an opportunity here to frame up the tools in terms of Python tools... `node` is like `python`, `npm` is like `pip`... NPM will install packages to `node_modules` just like `pip` will install to your virtualenv, etc. Is there a Python workalike to `create-react-app`?

tutorial/react_for_python.py
```diff
@@ -0,0 +1,358 @@
...
+
+  ## Quick intro to React
+
+  Now, let's go ahead and see what the code for our new React application looks like. Open up your favourite code editor, and open the `src/App.js` file.
+  This is your first React component! In React, most of the code you write will reside in some sort of a component. This is the top-level component, apptly named `App`, which job it is to basically be the entry point of the app. Here you will render all other components that you want to be displayed. Try changing the 
+  "Welcome to React" text to "Welcome to Dash", and save it. Now, go back to the browser, and notice how it hot-reloads the page!
+
+  #### JSX
+  You probably already noticed that the text you changed was wrapped in a `<h1>` tag. It looks like HTML, but slightly different! What gives? This is
+  what is called JSX. A language developed by the React team to have easy, inline, HTML markup in your JavaScript components! Don't let the word 'language' scare
```
nicolaskruchten: here we kind of contradict the "react is not a language" thing...

tutorial/react_for_python.py
nicolaskruchten: ah, never mind, i somehow missed the next bit :)

nicolaskruchten: Take 2... Is it worth talking about how you can `node file.js`? Mostly in this document we talk about running stuff with `npm start` or `npm init` and so on...

valentijnnieman: Yeah I just added that to make it more clear what `node` is for Python developers. Because I felt like just saying it is a 'JavaScript runtime' might not be enough. Hmm!

tutorial/react_for_python.py
valentijnnieman: "Don't let the word 'language' scare you!" :) Well in any case, JSX is a language isn't it? Perhaps we could say something like 'A markdown language similar to HTML' to make it clear from the get-go it is not something new?

nicolaskruchten: Well, I would say that JSX is a pretty big/weird departure from Javascript and Python-like code :)

valentijnnieman: I've always found JSX to be pretty easy to comprehend - 'just html in JS', but I do realise that there are quite a lot of people out there that find it weird/a big departure. So what are you suggesting? Remove the 'React is not a language' thing? I still don't really see why calling JSX a language is contradicting that React is not a language - JSX is a language used in React, but React itself isn't a language, right?

tutorial/react_for_python.py
```diff
@@ -0,0 +1,358 @@
...
+  to running a local server so you can see what your app looks like. It even sets up a system called `webpack`, which bundles all your code and assets
+  together and - you guessed it - packs it up for the web.
+
+  - To install Node.js, go to [the Node.js website](https://nodejs.org/en/) to download the latest (stable) version. 
+  - Node.js will automatically install the Node Package Manager `npm` on your machine
+  - Verify that node is installed by running: `node -v`
+  - Verify that npm is installed by running: `npm -v`
+
+  Now that we have Node.js up and running, we'll go ahead and initialize a new project using `create-react-app` by running:
+  - `npm init react-app my-app`
```
Bachibouzouk: Is it supposed to create a folder in the directory I was when I ran the command?

valentijnnieman: It sure is! It's supposed to create a dir named `my-app`, initialised as a `react-app`. It didn't do that for you?

tutorial/react_for_python.py
```diff
@@ -0,0 +1,358 @@
...
+
+  This sets up everything you need for a React application! It will give you some instructions on how to start it up after it's done
+  installing, most likely it will tell you to run `yarn start`. Yarn is another JavaScript package manager, which you can also use
+  instead of NPM. It's up to you - either `npm start` or `yarn start` will work!
+
+  ## Quick intro to React
+
+  Now, let's go ahead and see what the code for our new React application looks like. Open up your favourite code editor, and open the `src/App.js` file.
```
Bachibouzouk: Where am I supposed to find `src/App.py`? I typed the command in a folder and it only created a `package.json` file. It would be nice to be more specific here :)

tutorial/react_for_python.py
Bachibouzouk: if `npm - v` returns a version number inferior to 6, then it won't work but `npx create-react-app my-app` does

Bachibouzouk: `npm i -g npm` updates to the latest version :), thanks Valentin for your help!

tutorial/react_for_python.py
```diff
@@ -0,0 +1,358 @@
...
+    }
+    render() {
+      return <input />
+    }
+  }
+  
+  export default TextInput;
+  ```
+
+  Now, let's `import` that component and use it in our `App` component! Add the `import TextInput from './TextInput';` line to `App.js`, and somewhere
```
Bachibouzouk: suggestion to make it more specific and directed ...use it in `App.js` ! Add the `...` line to the top of the `App.js` file , then add `<TextInput />` somwhere in the return of the render(0 method of `App`.

tutorial/react_for_python.py
```diff
@@ -0,0 +1,358 @@
...
+  ``` 
+
+  Props always flow down, but you can set a method as a prop too, so that a child can call a method of a parent. For more information, please refer to the [React docs](https://reactjs.org/docs/components-and-props.html)
+
+  These are just the basics of React, if you want to know more, the [React docs](https://reactjs.org/docs) are a great place to start!
+
+  ## Using your React components in Dash
+
+  We can use most, if not all, components build in React in Dash! Dash uses React under the hood, specifically in the `dash-renderer`. The `dash-renderer` is basically just a React app, that renders the
+  layout defined in your Dash app as `app.layout`! It is also resposible for assigning the callbacks you write in Dash to the proper components, and keeping everything up-to-date. The `App` component
```
Bachibouzouk: respoNsible ;)

tutorial/react_for_python.py
```diff
@@ -0,0 +1,358 @@
...
+  we have been working with so far can simply be thought of as `dash-renderer`!
+
+  There are, of course, some steps you must take for your component to work nicely with Dash. The `props` you want to be available in Dash, should be specified as `propTypes` on the React component, so that the
+  Dash Component Archetype (note: this here will need some clarification) can read them and convert them to React. In our example, we can easily do this by changing our code like so:
```
Bachibouzouk: maybe add 'in `TextInput.js`' ?

tutorial/react_for_python.py
```diff
@@ -0,0 +1,358 @@
...
+        />
+      </div>
+    );
+  }
+}
+
+export default App;
+```
+
+We set `setProps` here to be a new function (an arrow function, read about them [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions)) that calls
```
Bachibouzouk: I am not sure I understood what was the benefit of using `setProp` here, maybe you could give a more specific example which highlights an active use of `setProps` which results in a different rendering in the browser? That would help me to make sure I understood ;)

tutorial/react_for_python.py
nicolaskruchten: Well, JSX is pretty complex actually... it's magically compiled down to `React.createComponent` calls, you can only return single objects from `{}` blocks, you can nest `{<X>{<Y>{ check ? <Z/> : null }</Y>}</X>}` (or similar), it's `className` instead of `class`, you have to use `prop={{key:value}}` (which always bites me!), you can do `<X {...obj} />` ... and on and on. Most importantly: there's no equivalent I know of in Python where there's an embedded/transpiled sublanguage, so it's definitely worth spending a few sentences on in a "React for the Python dev" guide :)

valentijnnieman: I don't think it is necessary to know that JSX elements are compiled down to React.createComponent calls, for beginners. Do we really want newcomers to get a big info dump like that, instead of getting up and running as quickly as possible? All you need to know (for now!) is that it's almost just like HTML, with a few exceptions like `className`, and that you can embed JavaScript with `{}`. At least that's how I feel about it - maybe I'm more of a practical learner! I agree with you that JSX is much more complex, but I feel like spending too much time on such details can really slow things down.

nicolaskruchten: I'd say there's no "one size fits all" here unfortunately... Some readers will feel put off by a breezy "it's magic" style of writing and others will be put off by an intense "here's how it works" type treatment :)

It might just be easier to skip JSX altogether and treat it as an advanced topic and just deal with `React.createComponent` which at least follows normal lexical rules, and then have a section at the bottom that explains why when they read others' React code there's JSX all over the place.

nicolaskruchten: (as a Python developer, I would find it less overwhelming if I could see the actual function calls, it maps more cleanly onto how Python works IMO)

valentijnnieman: But then readers would have to write React.createComponent calls in their `create-react-app`, which kind of defeats the point of `create-react-app`... Then they'd might be better off just including React.js in a basic HTML file on Codepen or something and writing it all by hand - I still think it's better to start top down and work your way through (if a user is confused by JSX, click the link and read up on it!) but I get that some people might not! Is the best option to more clearly link to a JSX explanation document?

nicolaskruchten: I think the underlying issue here is that culturally, Python developers shun 'magic' type code and prefer a more explicit "just do it the pythonic way" approach, whereas Javascript developers tend to be more enthusiastic about layering on transpilers and such... Given that we're writing a guide to the latter for the former, I would personally pull towards a gentler intro.

valentijnnieman: Sure, that's fair, I'm usually the same way about 'magic' type code. It's just that in this case, if the alternative is using `React.createComponent()` in the tutorial or spending a lot of time on explaining what JSX is from top-to-bottom, I'd rather gloss over it (and again, link to the docs for a far better and longer explanation than I could ever write). There's a LOT more magic in `create-react-app` and modern JavaScript & React, which you can only gloss over in a document of this type.

tutorial/react_for_python.py
valentijnnieman: This is related to the `npm init` command not working on npm < 6.0

tutorial/react_for_python.py
```diff
@@ -0,0 +1,360 @@
...
+
+  `props` are a component's properties. They are passed down from a component's parent, and are available as the `props` attribute. Calling `super()` on `props` in the constructor, makes our props available in the component as `this.props`. The `this` keyword in JavaScript is Python's `self`. We'll show you how to pass down
+  `props` a bit later on.
+
+  We're also defining `this.state` here, which is an `object`. `object`'s in JavaScript are a lot like `dict`'s in Python. They are specified in a notation
+  called `JSON`, which stands for JavaScript Object Notation. We're setting a `key` on our `state` object, called `value`, to be a string, 'default'.
+
+  Next, let's define our `render()` method, because all components should render something! Every `render()` method should `return` something, in our case, we'll return a HTML/JSX tag, `<input/>`. 
+  We'll also go ahead and `export` our component, as the `default`. This means whenever we're trying to `import` something from this file, and we don't specify a name, we'll get the `default` export.
+
```
Mnuzz: It might be worth elaborating on what exactly is being returned and how that is being used. It may be obvious to some that `return <input />` means that the `<input />` is intended to be some html code inserted into a location consistent with the location of the component in the DOM. But I feel that this explanation glosses over the important details, yet still seems to be written as though the author intended to explain how it works. Even an external link to a good technical description of what goes on here would suffice.

As someone who is learning React myself, I've found no end to the level of frustration due to having to work extra hard to fill in the missing explanatory gaps. This guide is very helpful.

tutorial/react_for_python.py
```diff
@@ -0,0 +1,360 @@
...
+  Props always flow down, but you can set a method as a prop too, so that a child can call a method of a parent. For more information, please refer to the [React docs](https://reactjs.org/docs/components-and-props.html)
+
+  These are just the basics of React, if you want to know more, the [React docs](https://reactjs.org/docs) are a great place to start!
+
+  ## Using your React components in Dash
+
+  We can use most, if not all, components build in React in Dash! Dash uses React under the hood, specifically in the `dash-renderer`. The `dash-renderer` is basically just a React app, that renders the
+  layout defined in your Dash app as `app.layout`! It is also responsible for assigning the callbacks you write in Dash to the proper components, and keeping everything up-to-date. The `App` component
+  we have been working with so far can simply be thought of as `dash-renderer`!
+
```
Mnuzz: This would be a great place to include a link to a more advanced technical overview of how dash-renderer works. With many weeks spent working with this, I have yet to find such an overview.

tutorial/react_for_python.py
```diff
@@ -0,0 +1,360 @@
...
+  This means we could use our component in Dash:
+  ```
+  app.layout = html.Div[
+    ourCustomComponent.TextInput(label='great!')
+  ]
+  ```
+
+  We also have a special prop available called `setProps`. This is Dash's way of keeping state *within Dash* (so not internally in our React component). You can think of it
+  as `setState()`, but for Dash. In your React component, you can thus call `this.props.setProps` (if it is set) *instead* of `this.setState()`, and the state will be handled
+  in Dash instead of in React! We can re-write our `TextInput` component as follows:
```
Mnuzz: I have been constantly confused by these setProps shenanigans, and unfortunately this explanation doesn't seem to get me that much closer to having a solid understanding of what I need to be doing with it. I've worked with components and even a did a deep dive into the entire Dash code base and I could not for the life of me figure out what determines how setProps is defined on a Component. This is a topic that deserves a lot of attention to help the reader understand. Even if it is out of scope for this particular document, I would hope it is addressed in a guide at some point.

Mnuzz:
I added a few comments in-line with the code

valentijnnieman:
Thanks @Mnuzz appreciate the feedback!

tutorial/chapter_index.py
```diff
@@ -193,34 +193,31 @@
...
     },
-### End of home.py ###
 
-### Start Components ###
+    'react-for-python': {
+        'url': '/react-for-python',
+        'content': react_for_python.layout,
+        'name': 'React for Python developers',
+        'description': 'A tutorial on how to program in React and JavaScript for Python developers.'
+    },
```
chriddyp: Let's move the "Build Your Own Components" tutorial under this new "Creating Your Own Components" section
image

tutorial/chapter_index.py
```diff
@@ -193,34 +193,31 @@
...
                        'Dash workshops, sponsored feature requests and ' \
                        'customizations.'
     },
-### End of home.py ###
 
-### Start Components ###
+    'react-for-python': {
+        'url': '/react-for-python',
+        'content': react_for_python.layout,
+        'name': 'React for Python developers',
```
chriddyp: capital `Developers`

tutorial/chapter_index.py
```diff
@@ -193,34 +193,31 @@
...
-        'description': 'More information for Dash demos, Enterprise trials, ' \
+        'description': 'More information for Dash demos, On-Premise trials, ' \
                        'Dash workshops, sponsored feature requests and ' \
                        'customizations.'
     },
-### End of home.py ###
 
-### Start Components ###
+    'react-for-python': {
+        'url': '/react-for-python',
```
chriddyp: let's use the full name `/react-for-python-developers`

chriddyp:
In the future, could we rebase off of master instead of merging master into the branch? It's really hard to tell which files have changed as part of this PR vs which ones were from master by looking at the files changed section: https://github.com/plotly/dash-docs/pull/116/files. Or, if it's not too much work, could we rebase _this_ PR and force push the changes up?

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+
+import styles
+
+layout = html.Div([dcc.Markdown('''
+  # React for Python developers: a primer
+
+  ## Introduction
+
+  If you're a Dash developer, at some point or another you probably have thought about writing your own set of components for Dash. 
+  You might have even taken a peek at some of our source code, or taken the `dash-component-archetype` for a spin. However, if you've
```
chriddyp: `dash-component-boilerplate`

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+
+layout = html.Div([dcc.Markdown('''
+  # React for Python developers: a primer
+
+  ## Introduction
+
+  If you're a Dash developer, at some point or another you probably have thought about writing your own set of components for Dash. 
+  You might have even taken a peek at some of our source code, or taken the `dash-component-archetype` for a spin. However, if you've 
+  never programmed in JavaScript and/or used React before, you might feel slightly confused. This guide hopes to bridge the gap between 
+  your knowledge of Python and Dash and the languages and frameworks we use to create awesome Dash components!
```
chriddyp: Let's say, 
```
By the end of this guide, you should feel comfortable creating your own Dash component in React and JavaScript, even if you have never programmed in those languages before. [Let us know how it went by commenting in this thread](https://github.com/plotly/dash-docs/issues/194).

File: tutorial/react_for_python.py [hunk 19]
```diff
@@ -0,0 +1,387 @@
...
+
+  #### JavaScript
+  JavaScript is the language of the web - all modern browsers can run it, and most modern web pages use it to make their pages interactive.
+  It is the de-facto standard of front end development, and has come a long way since it's inception. Today, modern JavaScript features
```
chriddyp: `its`

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+  a rich set of features, designed to create a development experience perfectly suited for the web.
```
chriddyp: I don't think these last two sentences add much, maybe let's say:
```
JavaScript is a fully featured programming language on its own. In web pages, it's primarily used to respond to user interactions (like running custom functions when a user clicks on an element) and to render and modify the HTML and CSS on the web page.
```
```

File: tutorial/react_for_python.py [hunk 21]
```diff
@@ -0,0 +1,387 @@
...
+  #### JavaScript
+  JavaScript is the language of the web - all modern browsers can run it, and most modern web pages use it to make their pages interactive.
+  It is the de-facto standard of front end development, and has come a long way since it's inception. Today, modern JavaScript features
+  a rich set of features, designed to create a development experience perfectly suited for the web. 
+
+  #### React
+  React is JavaScript library for building user interfaces, written and maintained by Facebook. It has been very popular over the last few
+  years, mainly because it brings the power of reactive, declarative programming to the world of front end development. 
+  What this means is that it brings power to those programming complex applications on the web, giving them the ability to write highly 
+  interactive web pages in a single language.
```
chriddyp: The last sentence here doesn't flow very well. Let's maybe say something like:

> React has made it easier to reason about user interface code and it's programming model encourages user interface code that modular and reusable. It also has a huge, vibrant open source community that has published all sorts of reusable UI components, from sliders to data tables, dropdowns to buttons.

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+
+  It is important to realise that React is just JavaScript. React is not a language on it's own, nor is it any sort of domain-specific framework
```
chriddyp: `its`

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+  `python my-code.py` to run Python code in a terminal, you'd run `node my-code.js` to run JavaScript code in a terminal. 
+  Node comes in very handy when developing, even when you intend to run the code in the browser.
+
+  #### NPM
+  With Node we can take advantage of the extensive tooling, provided by NPM, which stands for Node Package Manager. Besides being a package manager 
+  (not unlike `pip` for Python), `npm` also allows you to do other stuff like run scripts and perform tasks, such as creating a project for you by running `npm init`, starting up a project by running `npm start`, or firing custom scripts by running `npm run custom-script`. These scripts are defined in a `package.json` file, which every project that uses `npm` has. The `package.json` file holds your `requirements` and `devRequirements`, which can be installed using `npm install`, the same way `pip` has a `requirements.txt` option you can use in `pip install -r requirements.txt`. `package.json` also holds a `scripts` section, where custom scripts can be defined. It is usually a good idea to check out a new project's `package.json` file, to see what kind of scripts you can run!
```
chriddyp: > The `package.json` file also contains the project's dependencies. In this way, it's similar to the `requirements.txt` file that we frequently use in Python projects. `package.json` also holds a ...

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+  mostly just simple wrappers around existing React components. This means the entire React ecosystem is potentially usable in a Dash application!
+
+  ## Installing everything you need
+  Let's start by setting up our JavaScript development environment. We will use Node.js, NPM, and our `dash-component-boilerplate` to write our first React
+  application. Node.js is a JavaScript runtime, which allows you to run JavaScript code outside of the browser. Just like you would run 
+  `python my-code.py` to run Python code in a terminal, you'd run `node my-code.js` to run JavaScript code in a terminal. 
+  Node comes in very handy when developing, even when you intend to run the code in the browser.
+
+  #### NPM
+  With Node we can take advantage of the extensive tooling, provided by NPM, which stands for Node Package Manager. Besides being a package manager
```
chriddyp: >  With Node we can take advantage of the extensive tooling, provided by NPM, which stands for Node Package Manager.

Let's say:
> NPM is the "Node Package Manager" and it is used to install packages and run scripts.

tutorial/react_for_python.py
chriddyp: >  It is usually a good idea to check out a new project's package.json file, to see what kind of scripts you can run!

Let's remove the `, ` and the `!`

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+
+  If you go to the [dash-component-boilerplate repo](https://github.com/plotly/dash-component-boilerplate), you'll find the instructions to need to set up some React boilerplate code
+  that helps you quickly get up a React development environment, complete with the scripts we need to finally build our React component for Dash. It also sets up Babel, a transpiler for React so that we can write JSX, and Webpack, a tool
+  that bundles all of our code into one `bundle.js` file, optimized for speed.
```
chriddyp: If you go to the dash-component-boilerplate repo, you'll find the instructions to need to set up some React boilerplate code that helps you quickly get up a React development environment, complete with the scripts ~we need to finally~ `that are necessary to` build our React component for Dash. ~It also sets up Babel, a transpiler for React so that we can write JSX, and Webpack, a tool that bundles all of our code into one bundle.js file, optimized for speed.~ `These scripts will use a variety of technologies (e.g. Babel, Webpack, and more) to compile our code into a web-ready package.`

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+
+  - To install Node.js, go to [the Node.js website](https://nodejs.org/en/) to download the latest version.
```
chriddyp: Let's add "We recommend installing the LTS version."

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+  - To install Node.js, go to [the Node.js website](https://nodejs.org/en/) to download the latest version. 
+  - Node.js will automatically install the Node Package Manager `npm` on your machine
+  - Verify that node is installed by running: `node -v`
+  - Verify that npm is installed by running: `npm -v`
+
+  Now that we have Node.js up and running, go ahead and follow the instructions on the [dash-component-boilerplate repo](https://github.com/plotly/dash-component-boilerplate) to set up your component.
+
+  ## Quick intro to React
+
+  Now, let's go ahead and see what the code for our new React application looks like. Open up your favourite code editor, and open the `src/lib/components/ExampleComponent.react.js` file.
```
chriddyp: > Now, let's go ahead and see what the code for our new React application looks like. ~Open up your favourite code editor, and open the src/lib/components/ExampleComponent.react.js file.~In your favorite code editor open the src/lib/components/ExampleComponent.react.js file.

chriddyp: > In React, our user interfaces are composed of "components" and by convention there will usually be one main component per file.

chriddyp: > This component is being called in src/demo/App.js, which is the top-level component, apptly named App. 

➡️ 

> This project imports the `ExampleComponent` in `src/demo/App.js`, the demo application.

chriddyp: > It's job is to basically be the entry point of the app. Here you will render all other components that you want to be displayed. Try making some changes to both the top-level App component, and the ExampleComponent, like adding a <h1>Hello, Dash!</h1> in between the <div></div> tags, and notice how the page hot-reloads on changes!

Let's add a new line after the ", the demo application", and then update this to be:

> The demo application (`src/demo/App.js`) is what you see after you run `npm run demo`. To see how `src/demoApp.js` and `ExampleComponent.react.js` work, try adding `<h1>Hello, Dash!</h1>` inside the `render()` function of either of the files. When you make the save the files, your browser should automatically refresh and you should see your change.

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+  This is our first React component! In React, most of the code you write will reside in some sort of a component. This component is being called in `src/demo/App.js`, which is the top-level component, apptly named `App`. It's job is to basically be the entry point of the app.
+  Here you will render all other components that you want to be displayed. Try making some changes to both the top-level `App` component, and the `ExampleComponent`, like adding a `<h1>Hello, Dash!</h1>` in between the `<div></div>` tags, and notice how the page hot-reloads on changes!
+
+  #### JSX
+  The `<h1>` and `<div>` tags you see look exactly like HTML tags, however, they are slightly different! These tags are
```
chriddyp: > The `<h1>` and `<div>` tags you see look exactly like HTML tags, however, they are slightly different!

➡️ 

> > The `<h1>` and `<div>` tags you see look exactly like HTML tags however they are slightly different.

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+  Here you will render all other components that you want to be displayed. Try making some changes to both the top-level `App` component, and the `ExampleComponent`, like adding a `<h1>Hello, Dash!</h1>` in between the `<div></div>` tags, and notice how the page hot-reloads on changes!
+
+  #### JSX
+  The `<h1>` and `<div>` tags you see look exactly like HTML tags, however, they are slightly different! These tags are   
+  what is called JSX - a syntax extension to JavaScript developed by the React team to have easy, inline, HTML-like markup in your JavaScript components. It is mostly exactly the same as regular HTML. The main difference lies in the naming of attributes - `class` (for appending a CSS class to an HTML tag) is
+  named `className` here, as to avoid confusion with the actual `class` keyword in JavaScript (yes, JavaScript has classes, just like in Python!). Under the hood, React will call `React.createElement('h1')` for our `<h1>` JSX tags. React doesn't require you to use JSX,
```
chriddyp: > It is mostly exactly the same as regular HTML.

➡️ 

> There are a few key differences between JSX tags and HTML tags:
- The `class` keyword is renamed `className` (similarly as in Dash)
- In HTML, we write inline styles with strings like `<h1 style="color: hotpink; font-size: 12px">Hello Dash</h1>`. In JSX (as in Dash), we use objects with camelCased properties: `<h1 style={{"color": "hotpink", "fontSize": "12px"}}>Hello Dash<h1>`
- In JSX, any variables that we want to render need to be wrapped in `{}`. For example:
```
render() {
    var myText = 'Hello Dash!';
    return (
          <h1>{myText}</h1>
    );
}
```

chriddyp: > Under the hood, React will call React.createElement('h1') for our <h1> JSX tags. React doesn't require you to use JSX, but most people find it helpful as a visual aid when working with UI inside JavaScript. To learn more about how JSX works and why people use it, read the documentation here!

I'd say let's remove this. I never interact with `React.createElement` in practice and we should just present JSX like it's the standard, we don't have to win anyone over.

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+
+  #### JSX
+  The `<h1>` and `<div>` tags you see look exactly like HTML tags, however, they are slightly different! These tags are   
+  what is called JSX - a syntax extension to JavaScript developed by the React team to have easy, inline, HTML-like markup in your JavaScript components. It is mostly exactly the same as regular HTML. The main difference lies in the naming of attributes - `class` (for appending a CSS class to an HTML tag) is
+  named `className` here, as to avoid confusion with the actual `class` keyword in JavaScript (yes, JavaScript has classes, just like in Python!). Under the hood, React will call `React.createElement('h1')` for our `<h1>` JSX tags. React doesn't require you to use JSX, 
+  but most people find it helpful as a visual aid when working with UI inside JavaScript. 
+  To learn more about how JSX works and why people use it, read the [documentation here!](https://reactjs.org/docs/introducing-jsx)
+
```
chriddyp: Let's add a section like:

#### A Quick Primer on the JavaScript Language

1. **Variable Declaration**

In JavaScript, we have to declare our variables with `let` or `const`. `const` is used when the variable shouldn't changed, `let` is used elsewhere:
```
const color = 'blue';
let someText = 'Hello World';
let myText;
myText = 'Hello Dash';
```

2. **Comments**

Single line comments are prefixed with `//`. Multi-line comments are wrapped in `/* */`
```
/*
 * This is a multi line comment
 * By convention, we use a `*` on each line, but it's
 * not strictly necessary.
 */
const color = 'blue'; // This is a single line comment
```

3. **Strings**

Strings are defined the same way in JavaScript: single or double quotes:
```
const someString = 'Hello Dash';
const anotherString = 'Hello Dash';
```

Instead of Python's `format`, JavaScript allows you to embed variables directly into strings by wrapping the variable in `${}` and wrapping the string in backticks:
```
const name = 'Dash';
const someString = `Hello ${name}`;
```

4. **Dictionaries**

In Python, we use dictionaries for key-value pairs. In JavaScript, we use "objects" and they are instantiated and accessed very similarly:
```
const myObject = {"color": "blue", "size": 20};
myObject['color']; // is blue
myObject.color;  // another way to access the color variable
```

In Python, the keys of a dictionary can be any type. But in JavaScript, the keys can only be strings. JavaScript allows you to omit the quotes around the strings and we frequently do:
```
const myObject = {color: "blue"}; // notice how there are no strings around color
```

So, if you want to set a dynamic key in an object, you have to wrap it in square brackets:
```
const styleProperty = "color";
const myObject = {[styleProperty]: "blue"};
myObject.color;
```

5. **Lists**

In JavaScript, lists are called "arrays" and they're instantiated and accessed the same way:
```
const myList = ["Hello", "Dash", "!"];
myList[0];  // Hello
myList[1];  // Dash
myList[myList.length - 1]; // -1 references aren't allowed in JavaScript
```

6. **Semicolons**

In JavaScript, the convention is to end each line in a semicolon. It's not strictly necessary to do so anymore but it's still the convention.

7. **Print Statements, Errors, and the Console**

In JavaScript, we use `console.log` to print statements into the "console":

```
console.log("Hello Dash");
```

Since JavaScript runs in the web browser, we won't see these statements in our terminal like we would in Python.
Instead, we'll see these statements in the browser's "dev tools console". To access the console:
1. Right click on the web page
2. Click "Inspect Element"
3. Click on the "Console" tab.

To see for yourself, add a `console.log` statement inside the `render` method of `ExampleComponent`, refresh the page, and then inspect your browser's console.

Like Python, error messages and exceptions will also appear inside this console.

8. **If, For, While**

`if`
```
if (color === 'red') {
  console.log("the color is red");
} else if (color === 'blue') {
  console.log("the color is blue");
} else {
  console.log("the color is something else");
}
```

`for`
```
for (let i = 0; i < 10; i++) {
  console.log(i);
}
```

`while`
let i = 0;
while (i < 10) {
  i += 1;
}

9. Functions

In JavaScript, you'll see functions defined in two ways:

The new style way:
```
const add = (a, b) => {
  // The inside of the function
  const c = a + b;
  return c;
}

console.log(add(4, 6)); // 10
```

The traditional way:
```
function (a, b) {
  // The inside of the function
  const c = a + b;
  return c;
}

console.log(add(4, 6)); // 10
```

10. Classes

JavaScript classes are very similar to Python classes. For example, this Python class:
```
class MyComponent(Component):
    def __init__(self, a):
        self.a = a;
        super().__init__(a)

    def render(this):
        return self.a;
```

would be written in JavaScript as:
```
class MyComponent extends Component {
    init() {
        super();
        this.a = a;
    }

    render() {
        return this.a;
    }
}
```

11. Importing and Exporting

In Python, we can import any variable from any file. In JavaScript, we have to explicitly specify which variables we want to make "importable" by "exporting" the variables.

If we only want to export a single variable, we'll write `export default`:

`some_file.js`
```
const text = 'hello world';
export default text;
```

`another_file.js`
```
import text from './some_file.js';
```

If we want to export multiple variables, we'll just write `export`:
`some_file.js`
```
const text = 'hello world';
const color = 'blue';
const size = '12px';

export text;
export color;
```

`another_file.js`
```
import {text, color} from './some_file.js';  // note that we can't import size because we didn't export it
```

11. The Standard Library and Ramda

Unlike Python, JavaScript's "standard library" is pretty small. At Plotly, we use the 3rd party library [`ramda`](https://ramdajs.com/docs/) for many of our common data manipulations.

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+
+  #### Virtual DOM
+  If we look at the `App` component again, we see that it is `export`ed at the bottom, and if you open up the `src/demo/index.js` file you can see that it's imported there, so it can be used in a call to `ReactDOM.render()`. 
+  This `ReactDOM.render()` method is only called here, and only once. All it does is render our `App` component in a `div` element with the `id` of `root`. React handles all of the rendering of 
+  components internally, using a very smart system called the Virtual DOM. It does this so it can update DOM nodes internally in the most efficient and fast way, meaning we
+  only have to concentrate on declaring how we want things to look like, and how we want things to behave!
```
chriddyp: Let's simplify this a bit:
> All it does is render our `App` component in a `div` element with the `id` of `root`. React handles all of the rendering of components internally, using a very smart system called the Virtual DOM. It does this so it can update DOM nodes internally in the most efficient and fast way, meaning we only have to concentrate on declaring how we want things to look like, and how we want things to behave! 

➡️ 

>`ReactDOM.render()` is what actually renders our React code into HTML on the web page. This ReactDOM.render() method is only called here, and only called once.

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+  components internally, using a very smart system called the Virtual DOM. It does this so it can update DOM nodes internally in the most efficient and fast way, meaning we
+  only have to concentrate on declaring how we want things to look like, and how we want things to behave! 
+
+  #### Classes
+  Now, as a Python developer, classes are probably nothing new to you. In the JavaScript community, however, classes are a relatively new addition. They've become available to us
+  in `ES6`, which stands for EcmaScript 6 - an updated version of JavaScript. React greatly benefits from this enhanced versions of JavaScript, so they decided
+  to set us up with a transpiler called Babel. Babel is nothing more than a tool that converts (or transpiles) your ES6 (or ES7 and ES8) code into JavaScript
+  all browsers can understand (even older ones like Internet Explorer!). This allows us to write modern JavaScript - which includes classes - whilest not having to worry (too much) about browser support. Babel comes included with `create-react-app`.
```
chriddyp: I think we should ⚡️  most of this. It's a lot of new technologies and I'm afraid of overwhelming folks. If we want to talk about ES6, we could do so when we mention classes above in our "JavaScript Primer" with a little blockquote like:

```
> Heads up! Classes, among other features, are new language features in JavaScript. Technically, they're part of a new version of JavaScript called ES6. When we build our JavaScript code, a tool called Babel will convert these new language features into simpler JavaScript that older browsers like IE11 can understand.
```

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+
+  We see here in our `App` component that it is defined as a `class` which `extends` from the `Component` class of React. This provides some methods to us, for example the `render()` method we're using here. `render()` is the method that is called by the *component that is rendering it*. In our case, `render()` is called
+  by the `ReactDOM.render()` call in `index.js`. Notice how the `<App />` is called in the `ReactDOM.render()` method: our `App` component is used as a JSX tag!
+
+  #### Other methods on React.Component
+  Other methods provided by React are mostly related to component state management. We've got lifecycle hooks like `shouldComponentUpdate` and `componentDidMount` which allow you to better specify when and how a component should update. For these methods, please refer to the [React documentation](https://reactjs.org/docs/state-and-lifecycle.html)
```
chriddyp: Missing a period at the end of the sentence.

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+
+  ## Our very own React component
+  Now, let's create our very own component. Create a file in `src`, and let's name it `TextInput.js`. In the `TextInput.js` file, let's first import React
```
chriddyp: > Create a file in `src`, and let's name it `TextInput.js`

➡️ 

> Create a file named `TextInput.react.js` inside the `src/lib/components/` folder.

chriddyp: >  In the TextInput.js file, let's first import React and React.Component (the { Component } syntax is a shorthand) the same way as in App.js by writing import React, { Component } from 'react'.
> Next, let's define our component by writing it as a class, which extends React.Component.

Let's get rid of the text and just say:
> In `TextInput.react.js`, write:
```
import React, { Component } from 'react';

class TextInput extends Component {
  // we'll be working inside this section of the code
}
```

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+    constructor(props) {
+      super(props);
+      this.state = {
+        value: 'default'
+      }
+    }
+  }
+  ```
+
+  `props` are a component's properties. They are passed down from a component's parent, and are available as the `props` attribute. Calling `super()` on `props` in the constructor, makes our props available in the component as `this.props`. The `this` keyword in JavaScript is Python's `self`. We'll show you how to pass down
```
chriddyp: > React is in charge of keeping track of our components and all of it's state, so it keeps up-to-date. It does this very efficiently, with the use of something called the virtual DOM and React Fiber (here is a good place to start if you want to learn more about this).

Let's ⚡️  this section just to keep things more concise

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+  `props` a bit later on.
+
+  We're also defining `this.state` here, which is an `object`. `object`'s in JavaScript are a lot like `dict`'s in Python. They are specified in a notation
```
chriddyp: > We're also defining this.state here, which is an object. object's in JavaScript are a lot like dict's in Python. They are specified in a notation called JSON, which stands for JavaScript Object Notation. We're setting a key on our state object, called value, to be a string, 'default'.

If we have a "JavaScript Primer" section above, then let's ⚡️ this section too

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+  called `JSON`, which stands for JavaScript Object Notation. We're setting a `key` on our `state` object, called `value`, to be a string, 'default'.
+
+  Next, let's define our `render()` method for our new component. In React, we are declaring UI components, and the `render()` method is the method that is called
+  when React wants to render those component. React is in charge of keeping track of our components and all of it's state, so it keeps up-to-date. It does this very efficiently, with the use of something called the virtual DOM and React Fiber ([here](https://reactjs.org/docs/faq-internals.html) is a good place to start if you want to learn more about this).
+
+  A component's `render` method can return a basic string, for example `return "Hello, World!"`. When this component is used somewhere, it's `render()` method is called and "Hello, World!" will be displayed on our page. Likewise, you can return a React element (specified using JSX) and React will render that element.
+  There are more things you can do in `render()`, but typically you shouldn't modify state in this method. Read more [here](https://reactjs.org/docs/react-component.html#render)
```
chriddyp: > There are more things you can do in render(), but typically you shouldn't modify state in this method. Read more here

Let's ⚡️ this section too, just to keep things concise.

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+
+  We'll also go ahead and `export` our component, as the `default`. This means whenever we're trying to `import` something from this file, and we don't specify a name, we'll get the `default` export.
```
chriddyp: > We'll also go ahead and export our component, as the default.

Let's break this section down into step-by-step instructions. I think it'll make it more digestible/attainable and easier for users to reference if they have questions. So:
1. Create a boilerplate component
2. Put something in `render`
3. Export the component and import it into the demo app
4. View the component
(we should include a screenshot)
5. Render the data in `this.state`
6. Wire in `setState`

chriddyp: we could bold the titles too, so that it's easier to follow

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+  Tada!
+
+  We've got a text input. 
+  
+  However, this input doesn't really do much - it's not connected to anything, nor does it save what you type in. Let's change our `render()` method of `TextInput` to
+  set the HTML `value` attribute on our `<input />` tag, so it looks like this: `<input value='dash' />`. Save it, and we should now see that the value of our <input> tag
+  is set to 'dash'! We can also change our value to be that which is defined in our `state` object, so `<input value={this.state.value} />`. The `{}` syntax in JSX means 
+  that we want to write inline JavaScript in our JSX/HTML, so our `this.state.value` statement can be computed. Great! Now our input says 'default'! Unfortunately, our input
+  is still not very useful, because we can't change our input's value, try as we might.  
+
```
chriddyp: In the tutorial, I tried to explain why we can't type into it and I think the attendees found it helpful. I'm not sure how concise we can be, but maybe something like:

> It may seem odd to you that we can't type anything into the `<input/>` box. However, it's consistent with the React model: in our render method, we are telling React to render an input _with a particular value set_. React will faithfully render this input with that value no matter what, even if we try typing in it. In order to have the input update when we type, we have explicitly program this interaction: we have to listen to keypresses in the input and then tell React to _rerender_ the component with the new data. We'll do this by updating the our `state` on the input's `onChange` event:
> [Code snippet follows]

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+      })
+    }
+    render() {
+      return <input value={this.state.value} onChange={this.handleInputChange}/>
+    }
+  }
+  
+  export default TextInput;
+  ```
+  Here we wrote a method which we set on our HTML's input `onchange` attribute (written as `onChange` in JSX), which will fire
```
chriddyp: I would make this more consise with:
> our ~HTML's~ input ~`onchange`~ `onChange` attribute ~(written as `onChange` in JSX)~

chriddyp: >  which will fire every time an onchange event happens on our input

Maybe 

>  which will fire every time we type into the input

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+  the basic idea is that we have a `const` keyword to make a variable constant, i.e. you cannot modify it after creation. You can use the `let` keyword, which is not constant, but is bound to
+  the scope of the enclosing method, or the `var` keyword, which is also not constant, but, it's scope is broader. In React, it's best to just use `let` for normal variables, and `const` if 
+  you want the variable to be unchangable after definition.
+
+  Next, we use a method called `setState()` on our class, provided by `React.Component`. This method will handle updates to our `state` object. Because of how JavaScript works, setting `state`
+  directly will result in the *component not updating, thus not firing render() again, so we are not able to see our changes!*. This has to do with the updating of the internal state array - because
+  it is not a new object (as happens in `setState()`, it is not recognized as new) [here's a good article on the subject.](https://daveceddia.com/why-not-modify-react-state-directly/)
```
chriddyp: I'd simplify this a bit with something like:

> Next, we'll call a method called `setState()` that's provided by the `React.Component`. This method is really special, it'll do two things:
> 1. It'll merge the object that you provide with whatever was currently in `this.state`.
> 2. Then, it'll _rerender_ the component. That is, it'll tell React to call the components `render` method again with the new data set in `this.state`.

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+  export default TextInput;
+  ```
+  Here we wrote a method which we set on our HTML's input `onchange` attribute (written as `onChange` in JSX), which will fire
+  every time an `onchange` event happens on our input. This method has a parameter we named `e` for event, on which certain attributes are
+  set: `target.value` is what we need. This is how the HTML DOM works - for more information check out [these docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text).
+
+  You might also wonder what the `const` keyword means. In JavaScript, we have a variety of keywords used to define *in which scope a variable lives*. I won't go into the details too deeply, but
+  the basic idea is that we have a `const` keyword to make a variable constant, i.e. you cannot modify it after creation. You can use the `let` keyword, which is not constant, but is bound to
+  the scope of the enclosing method, or the `var` keyword, which is also not constant, but, it's scope is broader. In React, it's best to just use `let` for normal variables, and `const` if 
+  you want the variable to be unchangable after definition.
```
chriddyp: Let's keep this in the "JavaScript Primer" so that it's all consilidated. That'll make this section a bit shorter and fluid. Also, in the JavaScript Primer I kept out `var`. I think closures and scope is too in-depth for this guide.

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+
+  Next, we use a method called `setState()` on our class, provided by `React.Component`. This method will handle updates to our `state` object. Because of how JavaScript works, setting `state`
+  directly will result in the *component not updating, thus not firing render() again, so we are not able to see our changes!*. This has to do with the updating of the internal state array - because
+  it is not a new object (as happens in `setState()`, it is not recognized as new) [here's a good article on the subject.](https://daveceddia.com/why-not-modify-react-state-directly/)
+
+  See how this now allows you to type in our input component? We can also display our state, by writing our `render()` method something like:
```
chriddyp: Let's make this *7. Display more data in `render`*

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+      <p>{this.state.value}</p> // here we're displaying our state
+    </div>
+  }
+  ``` 
+
+  Notice that we're not allowed to return multiple elements from `render()`, but an element with children is totally fine.
+
+  We can also pass along properties to our components, via the before mentioned `props`. This works the same as assigning attributes on a component, as we'll demonstrate by adding a `label` prop to our `TextInput` component!
+  Let's edit our call to `<TextInput />` in `App.js` to say `<TextInput label='dash-input' />`. This means we now have a `prop` called `label` available on our `TextInput` component! 
+  In `TextInput`, we can reference this via `this.props`. Let's extend our `render()` method further so it renders our `label` prop:
```
chriddyp: Let's make this another section **8. Add Props to the Component** and let's say something like:

> Components can have their own "input arguments". In React, these are called "props" and a component's `props` are accessible as an object in `this.props`.
> 
> Let's add a `label` prop to `<TextInput/>`:
```
render() {
  return <div>
    <label>{this.props.label}</label>
    <input value={this.state.value} onChange={this.handleInputChange} />
    <p>{this.state.value}</p>
  </div>
}
```

> and let's pass in a label to the component from the parent in `src/demo/App.js`:
```
render () {
    return <TextInput label="My Dash Input"/>;
}
```

and this might be a good place to include a simple screenshot.

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+  render() {
+    return <div>
+      <label>{this.props.label}</label>
+      <input value={this.state.value} onChange={this.handleInputChange} />
+      <p>{this.state.value}</p>
+    </div>
+  }
+  ``` 
+
+  Props always flow down, but you can set a method as a prop too, so that a child can call a method of a parent. For more information, please refer to the [React docs](https://reactjs.org/docs/components-and-props.html)
```
chriddyp: In the workshop, I did another example that explained this and I think it was really helpful in understanding how dash-renderer and `setProps` work. Here's a draft:
Let's add a section like:

#### A Quick Primer on the JavaScript Language

1. **Variable Declaration**

In JavaScript, we have to declare our variables with `let` or `const`. `const` is used when the variable shouldn't changed, `let` is used elsewhere:
```
const color = 'blue';
let someText = 'Hello World';
let myText;
myText = 'Hello Dash';
```

2. **Comments**

Single line comments are prefixed with `//`. Multi-line comments are wrapped in `/* */`
```
/*
 * This is a multi line comment
 * By convention, we use a `*` on each line, but it's
 * not strictly necessary.
 */
const color = 'blue'; // This is a single line comment
```

3. **Strings**

Strings are defined the same way in JavaScript: single or double quotes:
```
const someString = 'Hello Dash';
const anotherString = 'Hello Dash';
```

Instead of Python's `format`, JavaScript allows you to embed variables directly into strings by wrapping the variable in `${}` and wrapping the string in backticks:
```
const name = 'Dash';
const someString = `Hello ${name}`;
```

4. **Dictionaries**

In Python, we use dictionaries for key-value pairs. In JavaScript, we use "objects" and they are instantiated and accessed very similarly:
```
const myObject = {"color": "blue", "size": 20};
myObject['color']; // is blue
myObject.color;  // another way to access the color variable
```

In Python, the keys of a dictionary can be any type. But in JavaScript, the keys can only be strings. JavaScript allows you to omit the quotes around the strings and we frequently do:
```
const myObject = {color: "blue"}; // notice how there are no strings around color
```

So, if you want to set a dynamic key in an object, you have to wrap it in square brackets:
```
const styleProperty = "color";
const myObject = {[styleProperty]: "blue"};
myObject.color;
```

5. **Lists**

In JavaScript, lists are called "arrays" and they're instantiated and accessed the same way:
```
const myList = ["Hello", "Dash", "!"];
myList[0];  // Hello
myList[1];  // Dash
myList[myList.length - 1]; // -1 references aren't allowed in JavaScript
```

6. **Semicolons**

In JavaScript, the convention is to end each line in a semicolon. It's not strictly necessary to do so anymore but it's still the convention.

7. **Print Statements, Errors, and the Console**

In JavaScript, we use `console.log` to print statements into the "console":

```
console.log("Hello Dash");
```

Since JavaScript runs in the web browser, we won't see these statements in our terminal like we would in Python.
Instead, we'll see these statements in the browser's "dev tools console". To access the console:
1. Right click on the web page
2. Click "Inspect Element"
3. Click on the "Console" tab.

To see for yourself, add a `console.log` statement inside the `render` method of `ExampleComponent`, refresh the page, and then inspect your browser's console.

Like Python, error messages and exceptions will also appear inside this console.

8. **If, For, While**

`if`
```
if (color === 'red') {
  console.log("the color is red");
} else if (color === 'blue') {
  console.log("the color is blue");
} else {
  console.log("the color is something else");
}
```

`for`
```
for (let i = 0; i < 10; i++) {
  console.log(i);
}
```

`while`
let i = 0;
while (i < 10) {
  i += 1;
}

9. Functions

In JavaScript, you'll see functions defined in two ways:

The new style way:
```
const add = (a, b) => {
  // The inside of the function
  const c = a + b;
  return c;
}

console.log(add(4, 6)); // 10
```

The traditional way:
```
function (a, b) {
  // The inside of the function
  const c = a + b;
  return c;
}

console.log(add(4, 6)); // 10
```

10. Classes

JavaScript classes are very similar to Python classes. For example, this Python class:
```
class MyComponent(Component):
    def __init__(self, a):
        self.a = a;
        super().__init__(a)

    def render(this):
        return self.a;
```

would be written in JavaScript as:
```
class MyComponent extends Component {
    init() {
        super();
        this.a = a;
    }

    render() {
        return this.a;
    }
}
```

11. Importing and Exporting

In Python, we can import any variable from any file. In JavaScript, we have to explicitly specify which variables we want to make "importable" by "exporting" the variables.

If we only want to export a single variable, we'll write `export default`:

`some_file.js`
```
const text = 'hello world';
export default text;
```

`another_file.js`
```
import text from './some_file.js';
```

If we want to export multiple variables, we'll just write `export`:
`some_file.js`
```
const text = 'hello world';
const color = 'blue';
const size = '12px';

export text;
export color;
```

`another_file.js`
```
import {text, color} from './some_file.js';  // note that we can't import size because we didn't export it
```

11. The Standard Library and Ramda

Unlike Python, JavaScript's "standard library" is pretty small. At Plotly, we use the 3rd party library [`ramda`](https://ramdajs.com/docs/) for many of our common data manipulations.


***

**10. Control the state in the parent**
In this example, the `value` of the `TextInput` is controlled by the component.

Let's move that `value` up to the parent component and control the state there.

In `src/demo/App.js`, add state and pass the value into the component:
```
component App extends Component {
  constructor() {
    super(props)

    this.state = {
        value: 'dash'
    };
  }

    render() {
      return <ExampleComponent label={'Dash'} value={this.state.value}/>
    }
}
```

In `src/lib/components/TextInput.react.js`, use the `value` prop instead of the state:
```
component TextInput extends Component {
  constructor() {
    super(props)

  }

    render() {
      return (
          <div>
            <label>{this.props.label}</label>
            <input value={this.props.value}/>
            <p>{this.props.value}</p>
          </div>
        )
    }
}
```

Now, as before, the `<input/>` won't actually update when you type into it. We need to update the component's `value` property as we type. To do this, we'll define a function in our parent component
that will update the parent's component state and we'll pass that function down into our component.
We'll call this function `setProps`:

```
component App extends Component {
  constructor() {
    super(props)

    this.state = {
        value: 'dash'
    };
  }

    setProps(newProps) {
      this.setState(newProps);
    }

    render() {
      return (
        <ExampleComponent
          label={'Dash'}
          value={this.state.inputValue}
          setProps={this.setProps}
        />
    }
}
```

and in `TextInput`, we'll call this function when the `value` of our `input` changes. That is, when we type into the input:

```
component TextInput extends Component {
  constructor() {
    super(props)

  }

    handleInputChange = (e) => {
      const newValue = e.target.value;
      this.props.setProps({value: newValue});
    }

    render() {
      return (
          <div>
            <label>{this.props.label}</label>
            <input value={this.props.value} onChange={this.handleInputChange}/>
            <p>{this.props.value}</p>
          </div>
        )
    }
}
```

To review, this is what happens when we type into our `<input>`:
1. The `handleInputChange` is called with whatever value we typed into the `<input/>`
2. `this.props.setProps` is called which in turn calls the `setState` property of the `App` component.
3. `this.setState` in `App` is called. This updates the `this.state` of `App` _and_ implicitly calls the `render` method of `App`
4. When `App.render` is called, it calls `TextInput.render` with the new properties, rerendering the `<input/>` with the new `value`.

In Dash apps, the `dash-renderer` project is very similar to `App.js`. It contains all of the "state" of the application and it passes those properties into the individual components. When a component's properties change through user interaction (e.g. typing into an `<input/>` or hovering on a graph), the component needs to call `setProps` with the new values of the property. Dash's frontend (`dash-renderer`) will then rerender the component with the new property _and_ make the necessary API calls to Dash's Python server callbacks.

**11. Handling the case when `setProps` isn't defined**

In Dash, `setProps` _is only defined if the particular component is referenced in an `@app.callback`_. If the component isn't referenced in a callback, then Dash's frontend will not pass in the `setProps` property, it will be undefined.

> As an aside, why does Dash do that? In some cases, it could be computationally expensive to determine the new properties. In these cases, Dash allows component authors to skip doing these computations if the Dash app author doesn't actually need the properties. That is, if the component isn't in any `@app.callback`, then it doesn't need to through the "effort" to compute it's new properties and inform Dash.
>

In most cases, this is a non-issue. After all, why would you render an `Input` on the page if you didn't want to use it as an `@app.callback`? However, sometimes we still want to be able to interact with the component, even if it isn't connected to Dash's backend. In this case, we'll manage our state locally _or_ through the parent. That is:
1. If `setProps` is defined, then the component will call this function when its properties change and Dash will faithfully rerender the component with the new properties that it passed up.
2. If `setProps` isn't defined, then the component isn't "connected" to Dash's backend through a callback and it will manage its state locally.

Here's an example with our `TextInput` component:
```
component TextInput extends Component {
  constructor() {
    super(props)
    this.state = props;
  }

    handleInputChange = (e) => {
      const newValue = e.target.value;
      this.props.setProps({value: newValue});
    }

    render() {
      let value;
      if (this.props.setProps) {
        value = this.props.value;
      } else {
        value = this.state.value;
      }
      return (
          <div>
            <label>{this.props.label}</label>
            <input value={value} onChange={this.handleInputChange}/>
            <p>{value}</p>
          </div>
        )
    }
}
```

chriddyp: and then we might want to update the section below accordingly.

tutorial/react_for_python.py
```diff
@@ -0,0 +1,387 @@
...
+    /** 
+    * The value of the text input (coming from Dash)
+    **/
+    value: PropTypes.string,
+
+    /** 
+    * Dash-provided callback for keeping state
+    **/
+    setProps: PropTypes.function
+  }
```
chriddyp: Let's add a quick section on `propTypes` too. Eventually, we might split it out into https://github.com/plotly/dash-component-boilerplate or something else, but for now this is our most extensive section. Maybe something like:

**12. Annotate your function with `propTypes`**

The final step in authoring your Dash component is to describe which properties are available.

This is done through React's `propTypes`. At the end of your file, write:
```
TextInput.propTypes = {
  ...
}
```

(fill that in with the actual ones)

These `propTypes` describe the input properties of the component, their types, and whether or not they are required. These are required. Dash's React-to-Python toolchain looks for these `propTypes` in order to automatically generate the Dash component Python classes. A few notes:
- The comments above each property are translated directly into the Python component's docstrings. For example, compare the output of `>>> help(dcc.Dropdown)` with that component's propTypes: https://github.com/plotly/dash-core-components/blob/c82bea0696dfa7af7f6a665bae81051523fff8b2/src/components/Dropdown.react.js
- The `id` property is required in all Dash components.
- The list of available types are available here: https://reactjs.org/docs/typechecking-with-proptypes.html
- In the future, we will use these propTypes to provide validation in Python. That is, if you specify that a property is a `PropTypes.string`, then Dash's Python code will throw an error if you supply something else. Track our progress in this issue: [dash#](https://github.com/plotly/dash/issues/264).

**13. Build your component in Python**

Now that you have your React component, you can build it and import it into your Dash program. View instructions in how to build the component in https://github.com/plotly/dash-component-boilerplate. 

In this tutorial, we rebuilt the `ExampleComponent` that was provided in https://github.com/plotly/dash-component-boilerplate/. So, the Python component code in `usage.py` should look familar - the properties and behaviour of `ExampleComponent` are exactly the same as our `TextInput`.

***

Feedback, questions? Let us know in on GitHub in our [dedicated feedback issue](https://github.com/plotly/dash-docs/issues/194).

valentijnnieman:
@

> In the future, could we rebase off of master instead of merging master into the branch? It's really hard to tell which files have changed as part of this PR vs which ones were from master by looking at the files changed section: https://github.com/plotly/dash-docs/pull/116/files. Or, if it's not too much work, could we rebase _this_ PR and force push the changes up?

I tried rebasing, got a whole bunch of merge conflicts, and for some reason the `react_for_python.py` file was missing most of the text (but not all of it for some dark reason)

tutorial/react_for_python.py
valentijnnieman: i.e. "Most modern web pages use it to make their pages interactive" and then "In web pages, it's primarily used to ... "

tutorial/react_for_python.py
valentijnnieman: > `These scripts will use a variety of technologies (e.g. Babel, Webpack, and more) to compile our code into a web-ready package.`
Are you sure about that last addition? Wouldn't it be good to explain (if only briefly) what Babel and Webpack do?

chriddyp:
Nice, this is looking good to me! I'm 💃  with this, I think the community will really benefit from this at this point. We can follow up with other improvements in other PRs (for example, I think @ashmpace is going to take a look too. I'm OK with incorporating her review in a subsequent PR, It'd be good to get this one out 😸  🚀 )

byronz:
> This is great; I was planning to make a component for a while and this is definitely going to be the changing factor for me to finally make it :) I was just wondering; for very large markdown text, would it be better to write it in a markdown document first, then import it inside the python file? I am not sure if that is standard practice.

yes I think it would be nice to separate the markdown content in `.md`, especially for our tutorial case, it's very large and you don't have the native support tool from all IDE like side preview.

and it makes the code much easier to maintain, as it will mainly focus on high-level layout and structure.

PR #122, opened 2018-07-19
consistent intros

chriddyp:
@cldougl here’s some improvements based off of what we discussed today:
- “Part 1”, “Part 2”, etc is now consistent
- There is a little intro at the top of each page in the tutorials that describe the previous and the next chapters
- A little outro provides a link to the next tutorial

### DIFF
--- a/tutorial/getting_started_part_2.py
+++ b/tutorial/getting_started_part_2.py
@@ -5,12 +5,13 @@ import dash_core_components as dcc
 import dash_html_components as html
 
 import plotly
+from textwrap import dedent as s
 
 import styles
 from tools import load_example
 
 examples = [
-    load_example(s) for s in [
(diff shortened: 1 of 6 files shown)


tutorial/getting_started_part_2.py
```diff
@@ -299,14 +257,17 @@
...
-    dcc.Markdown('''
-    The next chapter explains how to use these principles with the
-    `dash_core_components.Graph` component to make applications that
-    respond to interactions with graphs on the page.
-    '''.replace('    ', '')),
+
+    dcc.Markdown(s('''
+        The next part of the Dash tutorial covers an additional concept of
+        Dash callbacks: `State`
+        interactive.
```
cldougl: not sure interactive is the right word here? Is it `State` interactions?

PR #123, opened 2018-07-19
Add docs for the ConfirmDialog

T4rk1n:
Updated the version of dash-core-components to use `0.25.1`  and added examples and references for the new `dcc.ConfirmDialog` and `dcc.ConfirmDialogProvider`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ click==6.7
 configparser==3.5.0
 dash==0.24.1
 dash-auth==1.0.0
-dash-core-components==0.26.0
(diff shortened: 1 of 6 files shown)


chriddyp:
Looks great!

For some reason the confirm dialog isn't appearing for me on the components page: https://dash-docs-pr-123.herokuapp.com/dash-core-components

confirm-dialog-docs

Other than that, 💃

T4rk1n:
I think that `n` is `setProps` and since it's not used in a callback it is not included in the props. Didn't know about that when I made the component, I'll add a `setState` fallback.

PR #128, opened 2018-07-27
[WIP] Add tabs example

valentijnnieman:
I might not have enough time to really flesh this out, so I'm creating a work-in-progress PR here, so someone can take this over the finish line when I'm busy with moving to Montreal! :)


chriddyp:
looks good @valentijnnieman ! We can fill this out with more examples once you get back, for now this looks like a great start 👍

PR #129, opened 2018-07-27
ssh-add -k should be lower case k

cldougl:
http://man.openbsd.org/ssh-add#k
Upper case K errors out for linux

cc @michaelbabyn @chriddyp


michaelbabyn:
This command works on linux now so I'd say it's good as long as the lower case k in `ssh-add -k ~/.ssh/id_rsa` still works on Mac. :dancer:

cldougl:
yep all good- thnks!!

PR #130, opened 2018-07-27
Dash Deployment Platform -> Dash Deployment Server

nicolaskruchten:


nicolaskruchten:
@cldougl I did a bit more renaming here, hopefully that doesn't interfere with any work you're doing...

cldougl:
@nicolaskruchten nope doesn't interfere

nicolaskruchten:
@chriddyp sorry, I tacked on a few more changes, can you confirm your 💃 stands? :)

nicolaskruchten:
@cldougl we'll need to get new screencaps of the DDS so they don't say "App Manager" and there are some `app-manager` strings here that I didn't touch...

PR #135, opened 2018-08-07
new plugin docs

chriddyp:


chriddyp:
I've decided to keep the documentation around the components contained in the GitHub repo rather than keeping it up to date in two different places.

PR #139, opened 2018-08-13
Tab documentation edits

chriddyp:


chriddyp:
FYI @valentijnnieman - You might wanna take a read through these commits when you're back!

PR #141, opened 2018-08-15
Add external css/js `Dash.__init__` documentation.

T4rk1n:
Add documentation for using the `external_scripts` and `external_stylesheets` keywords from plotly/dash#305.


tutorial/external_css_and_js.py
```diff
@@ -37,6 +39,7 @@
 
     **Table of Contents**
     - Adding Your Own CSS and JavaScript to Dash Apps
+    - Adding external CSS and JavaScript
```
chriddyp: Let's just capitalize `external` to be consistent

PR #145, opened 2018-08-21
add images section

bcdunbar:
Resolves https://github.com/plotly/dash-docs/issues/140

- adds "Embedding Images in Your Dash Apps"

cc @cldougl do we want something basic like this or a little more info?


chriddyp:
this works for me! 💃

bcdunbar:
Thanks! 😄

PR #146, opened 2018-08-21
FAQs and Gotchas section

ned2:
This PR adds a new FAQs and Gotchas section, which includes 1) a list of frequently asked questions about using Dash and 2) a series of "gotchas". 

The gotchas currently largely pertain to the callbacks, and are are either limitations, or counter-intuitive properties which I've had to learn the hard way, and would have saved me time if I had this available as a reference.

I'd love to get feedback on the current FAQs and gotchas, as well as any suggestions for additions to this these lists!

### DIFF
--- a/tutorial/faqs.py
+++ b/tutorial/faqs.py
@@ -1,0 +1,237 @@
+# -*- coding: utf-8 -*-
+import os
+from textwrap import dedent as s
+
+import dash_core_components as dcc
+import dash_html_components as html
+
+from tutorial.components import Example, Syntax
+from tutorial import styles
+from tutorial import tools
+
+
+def get_example_name(path):
+    """Returns the name of an example given its path."""
+    # name is the filename without the suffix
+    return os.path.splitext(os.path.basename(path))[0]
+
+
+examples = {get_example_name(path): tools.load_example(path) for path in
+            ['tutorial/examples/faqs/last_clicked_button.py']}
+
+
+layout = html.Div([
+    dcc.Markdown(s('''
+    # FAQs and Gotchas
+
+    > This is the *7th* and final chapter of the essential [Dash Tutorial](/).
+    > The [previous chapter](/sharing-data-between-callbacks) described how to
+    > share data between callbacks. The [rest of the Dash documentation](/)
+    > covers other topics like multi-page apps and component libraries.
+
+
+    ## Frequently Asked Questions
+
+    **Q:** *How can I customize the appearance of my Dash app?*
+
+    **A:** Dash apps are rendered in the browser as modern standards compliant
+      web apps. This means that you can use CSS to style your Dash app as you
+      would standard HTML.
+
+      All `dash-html-components` support inline CSS styling through a `style`
+      attribute. An external CSS stylesheet can also be used to style
+      `dash-html-components` and `dash-core-components` by targeting the ID or
+      class names of your components. Both `dash-html-components` and
+      `dash-core-components` accept the attribute `className`, which corresponds
+      to the HTML element attribute `class`.
+     
+      The [Dash HTML Components](/dash-html-components) section in the Dash User
+      Guide explains how to supply `dash-html-components` with both inline
+      styles and CSS class names that you can target with CSS style sheets. The
+      [Adding CSS & JS and Overriding the Page-Load
+      Template](/external-resources) section in the Dash Guide explains how you
+      can link your own style sheets to Dash apps.
+
+    ------------------------
+    
+    **Q:** *How can I add JavaScript to my Dash app?*
+
+    **A:** You can add your own scripts to your Dash app, just like you would
+      add a JavaScript file to an HTML document. See the [Adding CSS & JS and
+      Overriding the Page-Load Template](/external-resources) section in the
+      Dash Guide.
+
+    ------------------------
+    
+    **Q:** *Can I make a Dash app with multiple pages?*
+    
+    **A:** Yes! Dash has support for multi-page apps. See the [Multi-Page Apps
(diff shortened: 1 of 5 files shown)


tutorial/faqs.py
```diff
@@ -0,0 +1,194 @@
...
+
+    ### Callbacks require *all* `Inputs`, `States`, and `Output` to be present in the layout
+    
+    If you have disabled callback validation in order to support dynamic
+    layouts, then you won't be automatically alerted to the situation where a
+    component within a callback is not found within a layout. In this situation,
+    where a component registered with a callback is missing from the layout, the
+    callback will fail to fire. For example, if you define a callback with only
+    a subset of the specified `Inputs` present in the layout, the callback will
+    simply not fire at all.
```
chriddyp: Is there a way that we can make this clearer that it's all of the `Input`, `States`, and `Output` _for a particular callback_ at a _given time_?

Maybe instead of saying "to be present in the layout" we say "to be rendered on the page"? as `layout` might imply `app.layout`

tutorial/faqs.py
```diff
@@ -0,0 +1,194 @@
...
+    For a given component/property pair (eg `'my-graph'`, `'figure'`), it can
+    only be registered as the `Output` of one callback. If you want to associate
+    two logically separate sets of `Inputs` with the one output
+    component/property pair, you’ll have to bundle them up into a larger
+    callback and detect which of the relevant `Inputs` triggered the callback
+    inside the function. For `html.Button` elements, detecting which one
+    triggered the callback ca be done using the `n_clicks_timestamp`
+    property. See this [Dash Community
+    forum](https://community.plot.ly/t/input-two-or-more-button-how-to-tell-which-button-is-pressed/5788/29)
+    post for an example.
```
chriddyp: Let's also add a follow up question like "How do I determine which `Input` has changed?

chriddyp:
Looks great to me! A couple of small edits, otherwise 💃 

Would love to hear from other members from the community or from @plotly/dash  too

T4rk1n:
Looks good to me, 💃

chriddyp:
Another big FAQ or gotcha is working with children. It takes a while for new users to realize that they can update huge swaths of their app by updating the `children` property of some parent component. For example, consider this community question: https://community.plot.ly/t/switching-multiple-apps-with-dropdown/12949

valentijnnieman:
Was this not automatically published to Heroku? I don't see a link anywhere!

valentijnnieman:
Looks very good to me! Nice to see this section, I think it's very helpful!

ned2:
Thanks for the feedback everyone! I've made some small fixes here and there based on those comments. 

I also added a *How do I determine which Input has changed?* question as you suggested @chriddyp. Could I get you to quickly review the answer plus code example, which I took from your forum post on how to use n_clicks_timestamp?

After that, I think it'll be good to merge.

ned2:
@cldougl, perhaps you could review my latest changes before I merge this? I just want to make sure someone is across the added example of using the n_clicks_timestamp.

tutorial/faqs.py
```diff
@@ -0,0 +1,229 @@
...
+    **Q:** *Can I use jQuery with Dash?*
+
+    **A:** For the most part, you can't. Dash uses React to render your app on
+      the client browser. React is fundamentally different to jQuery in that it
+      makes use of a virtual DOM (Document Object Model) to manage page
+      rendering. Since jQuery doesn't speak React's virtual DOM, you can't use
+      any of jQuery's DOM manipulation facilities to change the page layout,
+      which is frequently why one might want to use jQuery. You can however use
+      parts of jQuery's functionality that do not touch the DOM, such as
+      registering event listeners to cause a page redirect on a keystroke.
```
chriddyp: I'd also add:
"In general, if you are looking to add custom clientside behavior in your application, we recommend encapsulating that behavior in a [custom Dash component](https://dash.plot.ly/plugins).

tutorial/faqs.py
```diff
@@ -0,0 +1,229 @@
...
+    ------------------------
+
+    **Q:** *How do I determine which `Input` has changed?*
+
+    **A:** In addition to the `n_clicks` property (which tracks the number of
+    times a component has been clicked), all `dash-html-components` have an
+    `n_clicks_timestamp` property, which records the time that the component was
+    last clicked. This provides a convenient way for detecting which
+    `html.Button` was clicked in order to trigger the current callback. Here's
+    an example of how this can be done:''')),
```
chriddyp: I'd also add: "`n_clicks` is the only property that has this `timestamp` property. We will add general support for this in the future, you can track our progress in [dash#291](https://github.com/plotly/dash/issues/291)."

tutorial/sharing_state.py
```diff
@@ -17,12 +17,12 @@
...
-    > The [previous chapter](/interactive-graphing) covered how to use callbacks
-    > with the `dash_core_components.Graph` component.
-    > The [rest of the Dash documentation](/) covers other topics like multi-page
-    > apps and component libraries.
-    > Just getting started? Make sure to [install the necessary dependencies](/installation).
+    > This is the *6th* chapter of the essential [Dash Tutorial](/).  The
+    > [previous chapter](/interactive-graphing) covered how to use callbacks
+    > with the `dash_core_components.Graph` component.  The [rest of the Dash
+    > documentation](/) covers other topics like multi-page apps and component
+    > libraries.  Just getting started? Make sure to [install the necessary
```
chriddyp: Now that we have a chapter after this, let's update this to say "The [next and final chapter](/faq) covers frequently asked questions and gotchas."

PR #147, opened 2018-08-21
Circle update

T4rk1n:
### Update the config of circleci to version 2

- Removed tox
- Added python 3.6 and 3.7 containers.
- Pylint was failing with -28, did not include.
- Imports were incompatibles with python 3, fixed them all.


chriddyp:
💃 Thanks!

PR #160, opened 2018-08-28
Add error for pushing from non-master branch.

shammamah:


tutorial/dash_deployment_server_examples.py
```diff
@@ -1758,6 +1758,22 @@ def display_instructions2(method):
...
+        dcc.SyntaxHighlighter(s(
+            '''...
+            remote:            Adding BUILD_ENV to build environment...
+            remote:            Unable to select a buildpack'''),
+                              customStyle=styles.code_container, language='python'),
+        dcc.Markdown(s(
+            ''''
+            &nbsp;
+
+            This error might occur if you are trying to push from a branch that is not your `master` branch. Get the name of your current branch by running `git branch`. Then, to push from this branch to the remote server, run `git push plotly your-branch-name:master`.
```
chriddyp: Could we break this into multiple lines?

chriddyp:
Looks like there is a typo, it's causing tests to fail

Marc-Andre-Rivet:
@shammamah Is this PR still alive?

shammamah:
@Marc-Andre-Rivet Working on it right now!

shammamah:
@Marc-Andre-Rivet I'm not sure what the visual changes are, and I don't have access to view the details. How can I fix these issues?

Marc-Andre-Rivet:
@shammamah The difference is what would be expected (the added entry) + some dates changing from the previous snapshot (I've already approved those)

image

Someone should give you access shortly to the Percy snapshots for dash-docs.

Otherwise this looks fine to me. Will just ping @gzork and @chriddyp so they can have another look at the fixed version before merging.

Issue #161, opened 2018-08-29
Dash installation instructions needlessly require dash-renderer and plotly install


felixvelariusbos:
I recently re-installed Dash on a new computer, and noticed that when I `pip install dash`, it automatically installs dash-renderer and plotly; I don't need to do that.

It's not the worst thing in the world, but the installation page doesn't need to instruct me to install dash-renderer or plotly. Unless there's some edge case I can't think of where it's needed to know all the relevant Dash libraries, could the installation instructions be changed to reflect this? 3 packages is much less intimidating than 5. :)

rmarren1:
Are you talking aboue [here](https://dash.plot.ly/installation)? This will give you the most up to date packages. Sometimes installing `dash` will give you the version of `dash-renderer` required for that `dash` version, but we could have updated `dash-renderer` with more features in the meanwhile.

For example, installing `dash` now gives you 
https://github.com/plotly/dash/blob/7219b3d9f827e95bbbace801aa955b8c21ff8ffe/requirements.txt#L6
when you probably want all the goodies in `0.13.1`

T4rk1n:
Look like `plotly` and `dash_renderer` were added in plotly/dash#252 so now they install the latest version when installing dash.

felixvelariusbos:
@rmarren1 Yes sorry for being unclear; that's the page I was talking about.

rmarren1:
Oh I thought we pulled dependencies from `requirements.txt`.

It makes sense for `dash-renderer` to be removed, since users will never directly interact with it when making a Dash app.

chriddyp:
There is still some formatting to do, but this is how it could look:
![image](https://user-images.githubusercontent.com/1280389/44819267-3cf8b880-abba-11e8-9db0-5a02f9a87c33.png)

One downside of this is that the installation page is now a lot longer. If the text is appropriately de-emphasized, perhaps it's OK. 

Thoughts?

felixvelariusbos:
I hadn't thought of it being a handy way to see the current versions, so I see you're point there.

Since the instructions are telling users to use pip anyways, they can also just use `pip list` or `pip freeze` to check versions. However showing the __version__ variable is also doubling to show some functionality of the library (I can think of some use cases where I would check dash versions before doing some logic). So either/or there. :)

[...]

PR #163, opened 2018-08-29
include a note about versions

chriddyp:
see discussion in https://github.com/plotly/dash-docs/issues/161


jbrandes:
This still isn't working for me. Dash core components still doesn't load even with those instructions.

  import dash_html_compontents as html
ModuleNotFoundError: No module named 'dash_html_compontents'

alexcjohnson:
@jbrandes this PR is pretty old, installation is much simpler now than it was last September. Per https://dash.plot.ly/installation, all you really need to do is `pip install --upgrade dash` (for the latest version) `pip install dash==1.0.0` (if you need a specific version >=0.37.0) and you will get the proper versions of `dash_core_components`, `dash_html_components` (note the spelling) and `dash_table`

PR #165, opened 2018-08-31
version bumps

chriddyp:
:arrow_up: :arrow_up: :arrow_up:


chriddyp:
woops, dash-html-components 0.12.0 isn't published yet

chriddyp:
heads up @T4rk1n @rmarren1 @valentijnnieman 

It would be good if we could start updating this `requirements.txt` file after we publish releases. I find that it's a good final QA step to ensure that the publish worked OK

PR #169, opened 2018-09-02
Bump dash-html-components version to 0.12.0

rmarren1:


rmarren1:
I opened this too fast after publishing, can someone re-trigger just to be sure for the 3.6 CI? (Circle doesn't let me)

PR #170, opened 2018-09-02
add '--no-cache' flag to pip install.

rmarren1:
Without this you need to wait a few minutes after updating packages to run CI.


chriddyp:
💃

rmarren1:
Actually it looks like this doesn't really help since the cache isn't cached by CircleCI, I think there is just random lag time between PyPi upload and when it becomes available.

PR #176, opened 2018-09-07
create section describing new auth methods and logout.

michaelbabyn:
for https://github.com/plotly/streambed/issues/11325. @cldougl


tutorial/auth.py
```diff
@@ -200,6 +201,125 @@ def update_graph(dropdown_value):
...
+    '''.replace('    ', '')),
+
+    dcc.SyntaxHighlighter('''pip install dash==0.26.3
+pip install dash-auth==1.1.2''', customStyle=styles.code_container),
+
+    dcc.Markdown('''
+    Example Code:
+    '''.replace('    ', '')),
+
+    dcc.SyntaxHighlighter('''import dash
```
T4rk1n: I think we should have an example component here that will run. It would be a good way to test new versions of `dash-auth`. I think we can use some example account like the one we use for tests.

tutorial/auth.py
```diff
@@ -200,6 +201,125 @@ def update_graph(dropdown_value):
...
+        'data': [{'x': [1, 2, 3], 'y': [4, 1, 2]}]
+    }, style={'display': 'none'}),
+    auth.create_logout_button(
+        label='Sign out',
+        redirect_to='https://plot.ly')],
+     className='container')
+
+
+@app.callback(Output('title', 'children'), [Input('title', 'id')])
+def _give_name(title):
```
T4rk1n: Remove the `_` before methods in examples, it is not intended to be imported.

tutorial/auth.py
T4rk1n: `ldap_dn` is only for `on-prem` when it is setup with `ldap`.

I wanted an example that will actually run on the dash-docs website, like the reader can press login and use his plotly account to login and see the example app. But I tested and it doesn't work, `PlotlyAuth` will wrap all routes of `dash-docs` plotly/dash-auth#59.

You can remove the credentials for now and keep it in SyntaxHighlighter, I'll try to make that work later.

tutorial/auth.py
```diff
@@ -200,6 +201,131 @@ def update_graph(dropdown_value):
 app.scripts.config.serve_locally = True
 app.css.append_css({'external_url': 'https://codepen.io/chriddyp/pen/bWLwgP.css'})
 
+if __name__ == '__main__':
```
T4rk1n: I think we could add information about where PlotlyAuth will find the credentials to use before this example. It first check the `os.environ` then it check the file in `~/.plotly/credentials`. It need to be setup before starting `PlotlyOauth`.

michaelbabyn:
@T4rk1n , does this need anything else to be merged?

T4rk1n:
Looks good to me :dancer:, @chriddyp can you review too ?

tutorial/auth.py
```diff
@@ -200,6 +201,136 @@ def update_graph(dropdown_value):
...
+    `auth.set_user_data` to create a cookie containing an object associated with
+    their account and then we determine whether they can view the graph by checking their
+    permissions in that cookie using `auth.get_user_data`.
+    
+    Finally, we can logout the user by clearing the cookies. To do so, you can create a 
+    logout button and insert it in the layout or use `auth.logout()` in a callback.
+    
+    Note that `PlotlyAuth` uses your plot.ly user credentials. If it does not find 
+    `PLOTLY_USERNAME` and `PLOTLY_API_KEY` in your environment variables, it will search 
+    for them in `~/.plotly/.credentials`. See [here](https://plot.ly/python/getting-started/#initialization-for-online-plotting) for more details on setting your plotly
```
chriddyp: I try to avoid links that say `here`. [This is a good summary on why we avoid 'click here'](https://www.smashingmagazine.com/2012/06/links-should-never-say-click-here/).

So in this case, it could say `[Learn more about setting credentials](...)`

chriddyp: On second thought, I think we shouldn't actually promote the `~/.plotly/.credentials`  folder for dash apps because it makes initialization hard. Instead, let's remove that "Note that" and say 
```
Plotly Auth uses the environment variables `PLOTLY_USERNAME` and `PLOTLY_API_KEY`. You can find your username and API key at https://plot.ly/settings/api or, if you are using [Dash Deployment Server](https://plotly/dash/pricing), at https://your-plotly-server.com/settings/api.

You can set these variables directly in your code with:
```
import os

os.environ['PLOTLY_USERNAME'] = 'your-username'
os.environ['PLOTLY_API_KEY'] = 'your-api-key'
```

or, if you are using [Dash Deployment Server, you can keep your environment variables secret (view the docs)](https://dash.plot.ly/dash-deployment-server/environment-variables).
```

tutorial/auth.py
```diff
@@ -200,6 +201,136 @@ def update_graph(dropdown_value):
...
+            'width': '100%', 'border': 'thin lightgrey solid',
+            'border-radius': '4px'
+        }),
+
+    dcc.Markdown('''
+    Installation:
+    '''.replace('    ', '')),
+
+    dcc.SyntaxHighlighter('''pip install dash=={}  # The core dash backend
+        pip install dash-html-components=={}  # Dash Auth components
```
chriddyp: this should be `pip install dash-auth`

tutorial/auth.py
```diff
@@ -200,6 +201,136 @@ def update_graph(dropdown_value):
...
+        if perm_view_graph:
+            return {}
+        else:
+            return {'display': 'none'}
+    else:
+        return {'display': 'none'}
+
+
+app.scripts.config.serve_locally = True
+app.css.append_css({'external_url': 'https://codepen.io/chriddyp/pen/bWLwgP.css'})
```
chriddyp: Let's remove these two lines. The docs still use them all over the place even though we have a new way of setting dash assets

Issue #127, opened 2018-07-27
Document the tabs component


chriddyp:
Now that tabs are released, it's time to update our user guide with some examples 🎉 

See https://github.com/plotly/dash-docs/pull/123 for a similar set of files that will need to be updated.

For the examples, I like to:
- Include a bare bones example for the index page: https://dash.plot.ly/dash-core-components/. Basically including only the props that you need.
- In the "Examples and Reference" link for the page (it'll be https://dash.plot.ly/dash-core-components/Tabs), include several useful examples with some text explaining them and, for the first couple of examples, callbacks tying it all together. https://dash.plot.ly/dash-core-components/dropdown is a good example
- In all of the examples, I try to make them useful and aesthetically pleasing/inspiring

For tabs, I think these examples would be useful:
- Explaining the inline content vs content in a callback and clearly explaining the tradeoffs (loading all of the content vs loading it on the fly)
- An example overriding the colors with some other nice colors
- An example changing the style from full-width tabs to GitHub style tabs (as I attempted to do in the original PR)

And then we'll want to include the reference table for both the `Tabs` component and the `Tab` component.

valentijnnieman:
Because #183 is now merged, I think we can close this issue!

PR #183, opened 2018-09-13
Update tabs section

### LINKED ISSUE #173: Update Tabs docs
We've merged some PR's that contain fixes for the Tabs component, notably bugs that we mention in the docs. We should update the docs section about Tabs!

valentijnnieman:
This updates `dash-core-components` to the latest version, which includes fixes for bugs in the Tabs component, that are mentioned in the docs. It also removes those mentions, since the bugs should be fixed in this release. Closes #173


valentijnnieman:
@chriddyp I've added 3 examples on how to style Tabs, as per #127. Could you rereview? Thanks!

chriddyp:
Looks good! A couple suggestions:
- [ ] Can we include what the stylesheet looks like for the first example? You could include another code block above or below the python example. Above the css code, indicate the name of the file like *`asssets/tabs.css`* and also mention that you should 
```
Save the file in an `assets` folder (it can be named anything you want). Dash will automatically include this CSS when the app is loaded ([learn more about including CSS in your app](/external-resources)
```
image

- Could we do another example that includes really long tab names? This actually comes up a lot and I'd like to have a recipe handy for it. The difficulty with the CSS is figuring out how to have the tabs all the same height when one tab overflows to two lines and the other tab stays on one line. Here's a good example: https://dash-intro-presentation.herokuapp.com/dash/presentation/

- For the shorter tab example, is there any way to get the bottom border line to extend to the entire container? Similar to GitHub
image
image

- Could we replace the second example (the circular buttons) with one like this? 
image

- Finally, since this page is getting long, could we add a sort table of contents at the top? Similar to what we have here: https://dash.plot.ly/external-resources

chriddyp:
Another nice CSS example might be how to do multi-line tabs when you have many tabs. For example, here's the settings page from gmail - if there isn't enough space, the tabs move on to 2 or 3 lines:
image
image

valentijnnieman:
@chriddyp Alright, updated the existing examples a bit. Want to re-review?

chriddyp:
Looking at the review app, it looks like the graphs aren't appearing for the first example
image

chriddyp:
> Looking at the review app, it looks like the graphs aren't appearing for the first example

Looks like it's intermittent. If I refresh, then it works 😕

chriddyp:
image

The flow seems a little off here. Maybe just say "An alternative to providing CSS classes is to provide style dictionaries directly:"

chriddyp:
Otherwise, the examples look really great! Thanks for cleaning these up, I think the community will really appreciate the expert CSS support here 💯

chriddyp:
💃

PR #187, opened 2018-09-16
Added comment number 5 on Including Local CSS and JavaScript

RafaelMiquelino:
Include instruction to ensure local assets will always be loaded.


chriddyp:
Thanks @RafaelMiquelino !

PR #188, opened 2018-09-16
Corrected Running an app locally instructions

RafaelMiquelino:
For some reason, the `run.py` file is not located inside the `tutorial` folder, so the command to run the app doesn't work if you follow the instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,13 +23,13 @@ source activate dash_docs
 ```
 
 2. `pip install -r requirements.txt`
-3. `gunicorn tutorial.run:server`
+3. `gunicorn run:server`
 4. open http://127.0.0.1:8000 in your browser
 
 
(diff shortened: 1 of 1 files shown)


chriddyp:
The existing instructions work for me. Are you using Python 2 or Python 3? The instructions specify Python 2.

ned2:
run.py was indeed moved out of the tutorial folder in ddb2a7b16b0b4ba8fc76e6a10fa7f6e2ed10b132. So the docs do need updating to match the current layout. I actually couldn't see this at first because I still had a residual .pyc file from before the move, that the import was apparently still happy picking up -- which is slightly disconcerting.

chriddyp:
ah yeah, that's what must've been happening to me too. That's scary 🙈

PR #192, opened 2018-09-17
Added Dash DAQ Guage to Components Lists

jdamiba:
related to https://github.com/plotly/streambed/issues/11471

screen shot 2018-09-17 at 11 57 33 am

### DIFF
--- a/tutorial/core_components.py
+++ b/tutorial/core_components.py
@@ -438,5 +438,21 @@ confirm = dcc.ConfirmDialogProvider(
 
     html.Br(),
 
+    dcc.Markdown('***'),
+
+    html.A(
+        className="image-link",
+        href="http://dash-daq.netlify.com/#Gauge",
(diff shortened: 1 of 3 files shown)


tutorial/core_components.py
```diff
@@ -438,5 +438,21 @@
...
+    html.A(
+        className="image-link",
+        href="http://dash-daq.netlify.com/#Gauge",
+        children=html.Img(
+            src="https://s3-us-west-1.amazonaws.com/plotly-tutorials/static/marketing/dash/guage.png",
+            alt="Dash DAQ Guage Component"
+        )
+    ),
+
+    dcc.Link('More Guage Examples and Reference',
```
valentijnnieman: This link doesn't work currently - the `dcc.Link` component is used for single-page-app links, so it won't link outside of the docs app. Could you replace it with a `html.A` component?

Issue #172, opened 2018-09-07
Layout as a function.


T4rk1n:
I haven't seen this feature mentioned in the docs. Basically, we let users assign a function as layout, this function can then dynamically create the layout depending on some other variables. I think it's a great feature for more advanced users and we should document it somewhere.

There's a few caveats that should be mentioned when dealing with this feature.

- The layout function will be called before the first requests, meaning if you want to use `flask.requests` you will have to check if you actually have the request context.
- The initial layout should return the whole layout with all the components that are used in callbacks or set `app.config.suppress_callback_exceptions = True`.

rmarren1:
It is mentioned at the bottom of this: https://dash.plot.ly/live-updates

T4rk1n:
I see, I was expecting to see it in the dash layout section https://dash.plot.ly/getting-started.

I think it would still be good to include a warning about the callback exceptions.

ned2:
I could add an entry to the FAQ (#146)

Something like: "How can I create dynamic page layouts?"

perhaps?

michaelbabyn:
@T4rk1n , do you think this is addressed between https://dash.plot.ly/live-updates and the [new dash faq](https://dash.plot.ly/faqs) talking about callbacks in the layout?

> ## Callbacks require their Inputs, States, and Output to be present in the layout
> By default, Dash applies validation to your callbacks, which performs checks such as validating the types of callback arguments and checking to see whether the specified Input and Output components actually have the specified properties. For full validation, all components within your callback must therefore appear in the initial layout of your app, and you will see an error if they do not.
> However, in the case of more complex Dash apps that involve dynamic modification of the layout (such as multi-page apps), not every component appearing in your callbacks will be included in the initial layout. You can remove this restriction by disabling callback validation like this:
> app.config.supress_callback_exceptions = True

T4rk1n:
I think we should add a better example to the one in live-updates. Something that verify if it got the request context and return a whole layout if not.

I personally don't like to set `suppress_callback_exception` and prefer to not rely on it, so a valid pattern example where it isn't needed would be a good addition.

You can check if you have the request context with `flask.has_request_context()`.

michaelbabyn:
closed by #201

PR #201, opened 2018-09-25
add example demonstrating using a layout function to create a multi-

michaelbabyn:
page/tab app without suppress_callback_exceptions.

This PR address #172. I put the example under the other layout as a function example in live updates for now because I'm not sure where this would fit best.

@chriddyp, @cldougl maybe you could give me an idea of a better spot for this?


T4rk1n:
@chriddyp  There's no deploy ?

tutorial/live_updates.py
```diff
@@ -170,4 +170,110 @@ def serve_layout():
...
+import dash_html_components as html
+import dash_core_components as dcc
+from dash.dependencies import Input, Output, State
+
+import flask
+
+app = dash.Dash(__name__, external_stylesheets=['https://codepen.io/chriddyp/pen/bWLwgP.css'])
+
+layout_index = html.Div([
+    dcc.Tabs(id="tabs", value='tab-1', children=[
```
T4rk1n: I think an example with `dcc.Location` multi-page app would be better here, from what I've seen that pattern almost always suppress the exceptions. It is set in our multi-page examples https://dash.plot.ly/urls.

tutorial/live_updates.py
```diff
@@ -170,4 +170,110 @@ def serve_layout():
...
+               State('input-2-state', 'value')])
+def update_output(n_clicks, input1, input2):
+    return ('The Button has been pressed {} times,'
+            'Input 1 is "{}",'
+            'and Input 2 is "{}"').format(n_clicks, input1, input2)
+
+
+# Tab 2 Callbacks
+@app.callback(dash.dependencies.Output('tab-2-display-value', 'children'),
+              [dash.dependencies.Input('tab-2-dropdown', 'value')])
```
T4rk1n: Replace the long form Output/Input with the imported classes.

tutorial/live_updates.py
```diff
@@ -170,4 +170,110 @@ def serve_layout():
...
+For full validation, all components within your callback must therefore appear 
+in the initial layout of your app, and you will see an error if they do not.
+However, in the case of more complex Dash apps that involve dynamic modification 
+of the layout (such as multi-page apps), not every component appearing in your 
+callbacks will be included in the initial layout.
+
+Since this validation is done before flask has any request context, you can create 
+a layout function that checks `flask.has_request_context()` and returns a complete
+layout to the validator if there is no request context and returns the incomplete
+index layout otherwise:
```
T4rk1n: `:` typo ?

tutorial/live_updates.py
```diff
@@ -170,4 +170,110 @@ def serve_layout():
...
 a unique `layout` every hour or every day and serve the computed `layout`
 from memory in between.
-''')]
+'''),
+
+          dcc.Markdown('''
+
+***
+
+## Dynamically Create Layout for Multi-page App Validation
```
T4rk1n: Dynamically create a layout for multi-page app validation.

michaelbabyn:
@T4rk1n I updated this example to use dcc.Location instead of tabs to address your review

michaelbabyn:
Yeah, you're right @T4rk1n and now that it uses urls it makes sense for it to be with the other multi-page app documentation.

PR #204, opened 2018-09-26
Update plotly.py

### LINKED ISSUE #203: graph not rendering in /interactive-graphing
graph error in interactive-graphing

When I run the example code using my dash-docs venv it looks there is a unicode error in the traceback that is fixed when I upgrade plotly.py to 3.2.1

michaelbabyn:
This PR should fix #203


chriddyp:
💃

Issue #205, opened 2018-09-27
Create a component page out of date?


nicolaskruchten:
I was reading https://dash.plot.ly/plugins and it seems to me it's a bit out of date now that we generate materialized Python files and not JSON descriptors, right?

rmarren1:
I think this line can be updated:
> The Python component package includes the JSON file and the JavaScript bundle as extra files through the MANIFEST.in file.

It still contains the `metadata.json` file, but the package will include a Python class for each component.

nicolaskruchten:
I think it's actually worth stressing a bit that all Python code is autogenerated, so that you don't have to do it, but you *can* inspect it afterwards if you want. @VeraZab and I were trying to figure out how to confirm that once a prop was added to the React component, it had been picked up in the Python. The only ways I could think of were to check the generated Python (which isn't mentioned here) or to "try to use that prop in an app" but the former seems better :)

rmarren1:
What about modifying this paragraph? I think the word _dynamically_ is no longer relevant, and when talking about the Python classes we should let users know how to build them and where to find them.

> Dash reads this JSON file and ~dynamically~ **automatically** ~creates~ **generates** Python classes that subclass a core Dash component. These classes include argument validation, Python docstrings, types, and a basic set of methods. **You can inspect these files by checking in the folder with your component suite name (e.g. the `dash_html_components` folder) after running `npm run build:py`.** ~These classes are generated entirely automatically.~ A JavaScript developer does not need to write any Python in order to generate a component that can be used in the Dash ecosystem.

T4rk1n:
> Also, how does the build process work actually?

`metadata.json` is generated by the `extract-meta` js file, then generate_class use that to generate the python classes from the `metadata.json`.

Working to make the dash-component-boilerplate a cookiecutter currently. From my changes, we only need to run `build:all` and it will generate everything (js min/dev bundles, metadata, python classes). It's worth noting that dash needs to be installed in the python env to run the `build:py` command, I've added it to the boilerplate readme.

nicolaskruchten:
If the JSON file is only used to generate classes, is it even worth mentioning in this doc?

T4rk1n:
I think if someone wants to know how it works it should still be mentioned, but the process is now transparent with the `build:py` command and we don't read the metadata.json file at runtime anymore.

PR #215, opened 2018-10-04
some component creation clarifications

VeraZab:
Clears up: https://github.com/plotly/dash-docs/issues/205

@rmarren1 @T4rk1n @nicolaskruchten  Just thought I'd make these few modifications in the docs, because it helped me understand the system a bit better. Correct me if I'm wrong on anything or misunderstood any of your comments.

### DIFF
--- a/tutorial/plugins.py
+++ b/tutorial/plugins.py
@@ -33,14 +33,15 @@ compatible with the Dash ecosystem.
 
 On a high level, this is how that works:
 - Components in dash are serialized as [JSON](www.json.org).
-  To write a dash-compatible component, all of the properties
-  of the component must be serializable as JSON. For example,
-  JavaScript functions are not valid input arguments.
+  To write a dash-compatible component, all of the props
+  shared between the python code and the react code must be serializable as JSON.
+  Numbers, Strings, Booleans, or Arrays or Objects containing Numbers, Strings, Booleans.
+  For example, JavaScript functions are not valid input arguments.
   In fact, if you try to add a function as a prop to your Dash component, you
   will see that the generated Python code for your component will not include
   that prop as part of your component's accepted list of props.
-  (It's not going to be listed in the "Keyword arguments" enumeration or in the
-  self._prop_names array of the generated Python file).
+  (It's not going to be listed in the `Keyword arguments` enumeration or in the
+  `self._prop_names` array of the generated Python file for your component).
 - By annotating components with React docstrings (not required but helpful
   and encouraged), Dash extracts the information about the component's name,
   properties, and description through [React Docgen](https://github.com/reactjs/react-docgen).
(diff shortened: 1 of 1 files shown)


tutorial/plugins.py
```diff
@@ -36,18 +36,24 @@
   To write a dash-compatible component, all of the properties
   of the component must be serializable as JSON. For example,
   JavaScript functions are not valid input arguments.
-- By annotating components with React docstrings, Dash extracts
-  the information about the component's name, properties, and a description
-  of the components through [React Docgen](https://github.com/reactjs/react-docgen).
-  This is exported as a JSON file.
-- Dash reads this JSON file and dynamically creates Python classes that subclass
-  a core Dash component. These classes include argument validation,
+  In fact, if you try to add a function as a prop to your Dash component, you
```
VeraZab: I find that it's good to stress the importance of this. I find newbies don't necessarily notice this enough when building their app, and it causes some confusion.

tutorial/plugins.py
```diff
@@ -36,18 +36,24 @@
...
+  will see that the generated Python code for your component will not include
+  that prop as part of your component's accepted list of props.
+  (It's not going to be listed in the "Keyword arguments" enumeration or in the
+  self._prop_names array of the generated Python file).
```
chriddyp: can you update this to be markdown code? that is:
```
`self._prop_names`
```

tutorial/plugins.py
```diff
@@ -36,18 +36,24 @@
...
+  properties, and description through [React Docgen](https://github.com/reactjs/react-docgen).
+  This is exported as a JSON file (metadata.json).
+- At build time, Dash reads this JSON file and dynamically creates Python classes
+  that subclass a core Dash component. These classes include argument validation,
   Python docstrings, types, and a basic set of methods. _These classes are
   generated entirely automatically._ A JavaScript developer does not need to
   write _any_ Python in order to generate a component that can be used in the
   Dash ecosystem.
-- The Python component package includes the JSON file and the JavaScript bundle
-  as extra files through the `MANIFEST.in` file.
```
chriddyp: I'd like to keep `MANIFEST.in`. People frequently forget to put their files in this file and it's not immediately apparent because it only comes up when packaging the component.

tutorial/plugins.py
```diff
@@ -36,18 +36,24 @@
...
-- By annotating components with React docstrings, Dash extracts
-  the information about the component's name, properties, and a description
-  of the components through [React Docgen](https://github.com/reactjs/react-docgen).
-  This is exported as a JSON file.
-- Dash reads this JSON file and dynamically creates Python classes that subclass
-  a core Dash component. These classes include argument validation,
+  In fact, if you try to add a function as a prop to your Dash component, you
+  will see that the generated Python code for your component will not include
+  that prop as part of your component's accepted list of props.
+  (It's not going to be listed in the "Keyword arguments" enumeration or in the
```
T4rk1n: I would explain why instead: 

The shared props between the python code and react code must be pure json serializable attributes:

- Numbers
- Strings
- Boolean
- Arrays (of json props)
- Objects (of json props)

Any other prop type will not be included in the generated python component.

tutorial/plugins.py
chriddyp: We could make this clearer like:
> The files that are auto-generated by this build process are contained in the folder of your component. When you create your Python package, by default any non-Python files won't be included in the actual package. To include these files in the package, you must list them explicitly in `MANIFEST.in`. That is, `MANIFEST.in` needs to contain each JavaScript, JSON, and CSS file that you have included in your `my_dash_component/` folder.

VeraZab:
ah then would you make the updates in the docs for whatever changes please ? :)

tutorial/plugins.py
valentijnnieman: Nice, it's true that not all props *have* to be serializable, just the once that are shared. Nitpick: I feel "React" and "Python" should be capitalized. Also maybe "Arrays and Objects containing..." instead.

tutorial/plugins.py
```diff
@@ -33,21 +33,33 @@
...
   Dash ecosystem.
-- The Python component package includes the JSON file and the JavaScript bundle
-  as extra files through the `MANIFEST.in` file.
+- You will find all of the auto-generated files from the build process in the
+  folder named after your component. When you create your Python package, by default any
+  non-Python files won't be included in the actual package. To include these files
+  in the package, you must list them explicitly in `MANIFEST.in`. That is, `MANIFEST.in`
+  needs to contain each JavaScript, JSON, and CSS file that you have included in
+  your `my_dash_component/` folder. In the `dash-component-boilerplate` repository,
+  you can see that all the javascript for your react component is included in the
```
valentijnnieman: Similarly here, capitalize "javascript" and "react"

T4rk1n:
> ah then would you make the updates in the docs for whatever changes please ? :)

Of course.

PR #218, opened 2018-10-08
Bump `dash==0.28.1` -> `dash==0.28.2`.

rmarren1:
https://github.com/plotly/dash/pull/377


rmarren1:
Percy diffs were all dates and dash version :+1:

Issue #138, opened 2018-08-13
example of `n_clicks_timestamp` for button


chriddyp:
The button example:
![image](https://user-images.githubusercontent.com/1280389/44039649-80ebcd0a-9ee7-11e8-89de-66c24f1c83b9.png)

Could use a page of its own, like the rest of the components. In that page, we should include an example of `n_clicks_timestamp` (see the forum for examples)

michaelbabyn:
closed by #219

PR #219, opened 2018-10-09
add button examples page

bcdunbar:
addresses https://github.com/plotly/dash-docs/issues/138

- adds a dash-core-components/buttons page 
- more examples and reference
- adds `html.__file__` to `convert_props_to_table.py` for html props


michaelbabyn:
cc @cldougl

PR #228, opened 2018-10-17
dcc.Store documentation.

T4rk1n:
Add store examples and usage documentation.

### DIFF
--- a/run.py
+++ b/run.py
@@ -56,6 +56,10 @@ app.title = 'Dash User Guide and Documentation - Dash by Plotly'
 
 app.layout = html.Div(
     [
+        # Stores used by examples.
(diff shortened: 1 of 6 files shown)


T4rk1n:
@chriddyp 

I cannot get the callbacks in the examples to work when I run the app locally, but they work when deployed.

T4rk1n:
@chriddyp 

There is a circular import problem with the app, I can only get the callbacks if I import the app from `tutorial.tools`.

Also in componentWillMount, `setProps` is undefined. I added the Store components of my example to the initial layout and commented them in the example, it's now getting the setProps, so if a component is not in the initial layout it will not get setProps until later in the lifecycle.

PR #232, opened 2018-10-18
Dash DataTable Docs

chriddyp:
https://dash.plot.ly/datatable
:rocket:


chriddyp:
- [ ] Remove IDs from table

jonmmease:
### Part 2
 - [ ] 1) Top example is not responding, but the bottom example works well.

### Part 3
 - [ ] 1) This isn't a documentation thing, but I found it unintuitive that pressing "Enter" after typing a filter expression didn't apply the filter.  The filter wasn't applied until a click somewhere else to cause the cell to loose focus.
 - [ ] 2) Last example shows an empty table and no graph

### Part 4
 - [ ] 1) The last example doesn't show the squared values
screen shot 2018-11-01 at 9 58 52 am

### Part 5
 - [ ] 1) The dropdowns in the first example don't seem to respond (values are displayed but clicking on a different value doesn't change the cell value). But the second example works as I expected.

valentijnnieman:
This is perhaps more general docs related feedback, but I always find the margins to make the docs a bit harder to read than they could be. I know this is probably not a priority right now, but might be good to get this feedback in as long as you're working on it?

- [ ] `<h1>` tags have `margin-bottom: 2rem` but no `margin-top` - this makes the content above a h1 look too close to the title. I feel having `margin-top: 2rem` on there as well makes it look miles better.
- [ ] On the main DataTable page, the titles are `<li>` items, which don't have any margin, making them look squashed. I think either 
- [ ] The `<pre>` elements that hold the code examples have a margin of 10px - this could be 2rem for both top and bottom for me
- [ ] The `<p>` elements, same thing - currently has 0.75rem but could use 1 or even 2 rem on both top and bottom

tutorial/table/table_callbacks_chapter.py
```diff
@@ -0,0 +1,148 @@
...
+        className='example-container'
+    ),
+
+    section_title('Backend Paging with Filtering'),
+
+    dcc.Markdown(dedent('''
+    DataTable's front-end filtering has its own filtering expression
+    language.
+
+    Currently, backend filtering must parse the same filtering language.
```
cldougl: link to filtering language comprehensive list

tutorial/table/table_callbacks_chapter.py
```diff
@@ -0,0 +1,148 @@
...
+    If you write an expression that is not "valid" under the filtering
+    language, then it will not be passed to the backend.
+
+    This limitation will be removed in the future to allow you to
+    write your own expression query language.
+
+    In this example, we've written a Pandas backend for the filtering
+    language. It supports `eq`, `<`, and `>`. For example, try:
```
cldougl: should we note that you must tab or click out of the filtering cell to apply filter (`enter` tricked me 🙈 the 1st time)

tutorial/table/table_callbacks_chapter.py
```diff
@@ -104,6 +104,13 @@
     - Enter `> 5000` in the "gdpPercap" column
     - Enter `< 80` in the `lifeExp` column
 
+    > Note that unlike the front-end filtering, our backend filtering
+    > expression language doesn't require or support `num()` or wrapping
+    > items in double quotes (`"`).
+    > We will improve this syntax in the future,
```
cldougl: 🎉 that's more clear

tutorial/table/sizing_chapter.py
```diff
@@ -0,0 +1,620 @@
...
+        > dataframes that have different sizes and shapes. In particular,
+        > some of the dataframes have a large number of columns or have cells
+        > with long contents. If you'd like to follow along on your own
+        > machine, then open up the menu below to copy and paste
+        > the code behind these datasets.
+        '''
+        )),
+
+        html.Details(open=False, children=[
+            html.Summary('View the Datasets'),
```
cldougl: I love listing all the datasets here like this!

tutorial/table/sizing_chapter.py
```diff
@@ -0,0 +1,620 @@
...
+        > better accommodate wide content while keeping the table full-width
+        > and responsive. Subscribe to [plotly/dash-table#197](https://github.com/plotly/dash-table/issues/197) for more.
+        '''
+        )),
+
+        dcc.Markdown(dedent(
+        '''
+        ### Overflow Strategies - Multiple Lines
+
+        If your cells contain contain text with spaces, then you can overflow
```
cldougl: 👯‍♂️ double contain

tutorial/table/sizing_chapter.py
```diff
@@ -0,0 +1,620 @@
...
+        )
+        '''),
+
+        dcc.Markdown(dedent(
+        '''
+        Note how we haven't explicitly set the width of the individual columns
+        yet. The widths of the columns have been computed dynamically depending
+        on the width of the table and the width of the cells contents.
+        In the example above, by providing a scrollbar, we're effectively
+        giving the table as much width as needs in order to fit the entire
```
cldougl: as much width as it needs

tutorial/table/sizing_chapter.py
```diff
@@ -0,0 +1,620 @@
...
+layout = html.Div(
+    style={"marginLeft": "auto", "marginRight": "auto", "width": "80%"},
+    children=[
+
+        html.H1('DataTable Sizing'),
+
+        html.H3('Default Styles'),
+        dcc.Markdown(dedent(
+        '''
+        By default, the table will expand to the width of its container.
```
cldougl: do we always want to call this DataTable ?

PR #234, opened 2018-10-18
Add scale worker note

bcdunbar:
addresses https://github.com/plotly/streambed/issues/11411

- adds scaling `worker` content to `/dash-deployment-server/deployment`

@cldougl I'm not entirely convinced this is the best place for this. Open to suggestions if you think it is better in say Application Structure, Troubleshooting. etc..

### DIFF
--- a/tutorial/dash_deployment_server_examples.py
+++ b/tutorial/dash_deployment_server_examples.py
@@ -398,6 +398,22 @@ if __name__ == '__main__':
 
                     dcc.Markdown(s(
                     '''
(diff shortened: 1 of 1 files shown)


cldougl:
@bcdunbar I think this is a good spot but we should _also_ add a note in the troubleshooting section

cldougl:
@michaelbabyn can you take over this pr (just need to implement https://github.com/plotly/dash-docs/pull/234#issuecomment-436792210 )

PR #238, opened 2018-10-25
updating GTM tag

jdamiba:


tutorial/examples/external_css_and_js/custom-interpolate-string.py
```diff
@@ -11,8 +11,26 @@ def interpolate_index(self, **kwargs):
         <html>
             <head>
                 <title>My App</title>
+                <!-- Global site tag (gtag.js) - AdWords: 1009791370 -->
```
nicolaskruchten: this is an example, and should not contain our tags.

tutorial/examples/external_css_and_js/custom-index-string.py
```diff
@@ -13,8 +13,24 @@
         <title>{%title%}</title>
         {%favicon%}
         {%css%}
+        <!-- Global site tag (gtag.js) - AdWords: 1009791370 -->
```
nicolaskruchten: this is an example, and should not contain our tags.

PR #245, opened 2018-10-30
Add documentation for multiple file project structure

### LINKED ISSUE #244: Document multi-file structure for Dash projects
Issue moved from: https://github.com/plotly/dash-recipes/issues/19

The Dash guide shows how to create a Dash app in a single file, and a basic example of a [multi-page project with separate files](https://dash.plot.ly/urls). However, each app file quickly grows to several hundred files for a moderately sized dashboard (i.e. when layout and callbacks are combined in a single file). 

The question of how to organize Dash apps, so that callbacks are in a separate file, has been raised several times in the Dash community forum. 

Lets add a recipe for how to break down a Dash app into separate files, as can be seen in the Dash [slapdash boilerplate](https://github.com/ned2/slapdash/tree/master/src/slapdash). The Slapdash project is well-designed, but does not explain **why** the files are structured in such a way and **how** to create a simple Dash project with that file structure.

brylie:
Ready for review.

Closes #244

This PR is intended to document an example, or recommendation, for separating Dash apps into multiple files. E.g.

- index.py - main entrypoint
- app.py - app initialization and layout binding
- layout.py - layout definition
- callbacks.py - callbacks related to layout

This PR intentionally leaves out routing and some other code examples, to focus on clarity and simplicity of the text.

### DIFF
--- a/tutorial/chapter_index.py
+++ b/tutorial/chapter_index.py
@@ -17,6 +17,7 @@ from tutorial import introduction
 from tutorial import live_updates
 from tutorial import performance
 from tutorial import plugins
+from tutorial import project_structure
 from tutorial import search
 from tutorial import sharing_state
 from tutorial import state
@@ -195,6 +196,16 @@ chapters = {
                        '`dcc.Location`) that allow you to easily make ' \
                        'fast multipage apps using its own "Single Page ' \
                        'App (SPA)" design pattern.'
+    },
+
+    'project-structure': {
+        'url': '/project-structure',
+        'content': project_structure.layout,
+        'name': 'Structuring Dash apps for maintainability',
+        'description': 'As your Dash apps grow ' \
+                       'it becomes useful to separate the code into multiple files. ' \
+                       'This improves maintainability, ' \
+                       'by separating project code along conceptual and functional boundaries.'
     },
 
     'auth': {
--- a/tutorial/home.py
+++ b/tutorial/home.py
@@ -121,7 +121,10 @@ layout = html.Div(className='toc', children=[
                 chapters['external']['description']),
         Chapter(chapters['urls']['name'],
                 chapters['urls']['url'],
-                chapters['urls']['description'])
+                chapters['urls']['description']),
+        Chapter(chapters['project-structure']['name'],
+                chapters['project-structure']['url'],
+                chapters['project-structure']['description']),
     ]),
 
     Section('Production', [
--- a/tutorial/multiple_file_project_structure.py
+++ b/tutorial/multiple_file_project_structure.py
@@ -1,0 +1,1 @@
+# Placeholder
--- a/tutorial/multiple_file_project_structure.py
+++ b/tutorial/multiple_file_project_structure.py
@@ -1,1 +1,0 @@
-# Placeholder
--- a/tutorial/project_structure.py
+++ b/tutorial/project_structure.py
@@ -42,8 +42,9 @@ if __name__ == '__main__':
 
     dcc.Markdown("""
 ## Server
-`server.py` is where we actually define the Dash app instance, and possibly grab a reference to the Flask 
-server instance:"""),
+`server.py` is where we actually define the Dash app instance and attach the app layout. 
+
+    """),
 
     dcc.SyntaxHighlighter("""
 # Required import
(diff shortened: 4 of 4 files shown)


brylie:
@chriddyp this PR is ready for review.

ned2:
Awesome, this is a good start to get the discussion rolling. 

(Note: the relevant page is https://dash-docs-pr-245.herokuapp.com/project-structure)

The main thing I'm wondering is whether this example project layout needs to have its own separate section. It's just one way of laying out a Dash project alongside one other approach already presented in the [URLs](https://dash.plot.ly/urls) chapter under "Structuring a Multi-Page App". Users should be able to see both together as alternatives, so if we make a new project layout chapter, then I think the existing approach needs to be documented here also, but that would lead to duplication. 

I'm also not sure if we need quite as much commentary. I think the main architectural property of both approaches that is worth calling out is that `index.py` is the entry point of the app, and that `app.py` defines the Dash instance separately in order to avoid circular imports.

I'm leaning towards paring this down to another project layout similar to and alongside the existing one in "Structuring a Multi-Page App" and trying to make the new project layout as similar to the existing one as possible so as to reveal the differences between the two:
 * `app.py`
 * `index.py`
 * `layouts.py`
 * `callbacks.py`

We could then expand upon the existing prose, introducing the two as alternative approaches and explaining the organisational approach I mentioned above. 

@chriddyp, do you have any thoughts?

ned2:
Oh, speaking of circular imports, it's probably safer to do `app.layout = main_layout` inside `index.py`. While the current structure works now, if someone were to try to import from `app.py` in `layouts.py` they will get a circular import failure.

brylie:
> if someone were to try to import from app.py in layouts.py they will get a circular import failure.

Out of curiosity, why might you foresee someone importing from app.py in layouts?

ned2:
I do that in slapdash, in order to access the Flask instance, which is a
use case I've seen elsewhere.

https://github.com/ned2/slapdash/blob/master/src/slapdash/layouts.py

T4rk1n:
Importing global variables from modules is always bound to get circular imports. It's also dependent on the os, so YMMV. The best way to avoid that is to use functions that takes the variables as parameter instead.

brylie:
@ned couldn't you degine your config stuff in its own module, rather than in the app?

What is a simple milti-file approach to separating app, layout, and callbacks into separate modules, without a lot of boilerplate or circular imports?

I am hoping that this recipe will be simple to follow.

ned2:
@brylie, there's a range of approaches to avoiding circular imports with Flask apps. I'm only just starting to get my head around them now. The Flask [application context](http://flask.pocoo.org/docs/1.0/appcontext/)  for example helps with this, and is often used in conjunction with the [app factory pattern](http://flask.pocoo.org/docs/1.0/patterns/appfactories/) and [blueprints](http://flask.pocoo.org/docs/1.0/blueprints/). However I'm not sure if these can be used to get access to the `Dash` app instance, since they're specific to Flask. 

I feel like the approach that slapdash and the existing example in the Dash docs use, of organising your modules to minimise the chances of circular import (see also [this stack overflow answer](https://stackoverflow.com/a/48722390/77533) gives a pretty good compromise of simplicity while removing a large range of potential circular import errors. But @T4rk1n is right, if you want to make it bullet proof you need to resort to other approaches. 

I think this underscores why it might be good not to present a canonical recommended project structure as you do in your PR, because there's so many different ways one could do it, and it will depend on the needs of your project. But I do think adding another example layout in "Structuring a Multi-Page App" to give people more inspiration is a good idea.

brylie:
> I do think adding another example layout in "Structuring a Multi-Page App" to give people more inspiration is a good idea.

OK, sounds good. I would appreciate if someone opened a separate PR with an additional example. I don't have much more bandwidth to document alternatives, and would like to see something available in the documentation soon.

ned2:
Sure, I can put something along those lines together. Thanks for catalysing this!

By the way, I had a play around with other approaches for project structuring that takes advantage of Flask's application context. You can see my current approach in here: https://github.com/ned2/slapdash/blob/master/src/slapdash/app.py

brylie:
Thanks @ned2. I really appreciate your guidance on this issue 🕵

ned2:
Closing this in favour of #387

PR #255, opened 2018-11-06
Bump dash-table version to 3.1.3

Marc-Andre-Rivet:
Relates to https://github.com/plotly/dash-table/issues/217.


requirements.txt
```diff
@@ -19,7 +19,7 @@ Flask-Caching==1.4.0
 Flask-Compress==1.4.0
 Flask-Cors==3.0.3
 Flask-SeaSurf==0.2.2
-greenlet==0.4.13
+greenlet==0.4.14
```
Marc-Andre-Rivet: Fix version warning during pip install

PR #262, opened 2018-11-08
upgrade dash-core-components to 0.38.0

chriddyp:


chriddyp:
Also it looks like the CSS has changed in Percy. Do we know why that is?
image

valentijnnieman:
> Also it looks like the CSS has changed in Percy. Do we know why that is?
> image

Not a clue! Looks like the CSS works in the deployed app though.

PR #267, opened 2018-11-13
Add cookiecutter boilerplate usage documentation.

T4rk1n:
Added cookiecutter usage info in the react for python developers guide. I think that section should be renamed to **Develop your own dash components** or something and making it more language agnostic. With dashR coming, it make less sense to target it only for python developers.

### DIFF
--- a/tutorial/react_for_python_developers.py
+++ b/tutorial/react_for_python_developers.py
@@ -65,8 +65,106 @@ layout = html.Div([dcc.Markdown('''
   - Node.js will automatically install the Node Package Manager `npm` on your machine
   - Verify that node is installed by running: `node -v`
   - Verify that npm is installed by running: `npm -v`
+  
+  ## Python
+  
+  You will need python to generate your components classes to work with dash.
+  
+  Download python on the official website or through your os distribution package manager.
+  
+  - https://www.python.org/
+  - `apt-get install python`/`yum install python`
+  
+  ### Virtual environments
+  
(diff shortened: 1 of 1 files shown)


T4rk1n:
I've seen the `Writing your own component` section after writing in the react guide, maybe I should move the boilerplate info there ?

T4rk1n:
All percy diffs shows this:

bpostlethwaite:
Thanks @T4rk1n - @cldougl is this enough to work with (and possibly move into the component section)? 

Once we have it finalized we can take a look at those Percy diffs :)

cldougl:
yep this is great! I'll review here and then perhaps we can just add a link to the dev guide from the creating your own components section so we don't need to manage the same content in multiple places.

cldougl:
https://github.com/plotly/dash-docs/pull/267#issuecomment-438356945 that's a bit weird with the image diffs (it almost looks like a scrollbar)- but everything looks correct in the deployment 👍

cldougl:
The create a component section is pointing to the dash-component-boilerplate repo already so I don't think anything needs to be edited there.
screen shot 2018-11-13 at 5 33 06 pm

@T4rk1n your updates to the react guide for python devs is a sufficient update 👍

T4rk1n:
The widget in the percy diffs is sumo.

nicolaskruchten:
Can we remove sumo please? We’ve cancelled that service...

cldougl:
sumo removal issue logged: https://github.com/plotly/dash-docs/issues/271

PR #272, opened 2018-11-14
Update dash to 0.30.0, dash-renderer to 0.15.0

T4rk1n:
Hot-reload 🔥


chriddyp:
Let's follow up quickly with the documentation suggestions made in https://github.com/plotly/dash/pull/362#issuecomment-436799703

Issue #194, opened 2018-09-18
"React for Python Devs" - Feedback Thread


chriddyp:
This is a placeholder issue to get feedback from users about our forthcoming "React for Python Developers" guide. By following the instructions in that guide, you should feel comfortable creating your own React Dash components, even if it was your first time programming in JavaScript and React. 

How did it go? What did we leave out? Let us know in this issue!

JonathanChiang:
not really sure what cookie cutter was for and going through the boiler plate wasn't very clear, couldn't really get through the steps.

chriddyp:
Thanks for the feedback @JonathanChiang ! Sounds like we should clean up some of our language, as the boilerplate is now the same as the cookiecutter 😕 . 

> couldn't really get through the steps.

Could you describe in more detail which steps didn't work for you, or which steps were confusing?

Thanks for your feedback 🙇

JonathanChiang:
I tried it again today and got most of it working. I guess working on a few different technologies at once was a bit confusing at first. I think a two sentence description in the documentation of what cookie cutter is would've cleared things up a bit. Appreciate you following up. 

Love Dash, hope this project keeps growing!

emmanuelle:
I'm going through the tutorial at the moment and I'll try to suggest or implement some improvements when I'm done. At the moment I'm confused by the instruction to do ``npm run demo``. I followed the instructions setting up a directory with cookiecutter, but in package.json, there is no ``demo`` script so that ``npm run demo`` returns an error
```
emma@hopper:~/sandbox/dash_core_component$ npm run demo
npm ERR! missing script: demo

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/emma/.npm/_logs/2018-10-27T11_31_06_640Z-debug.log
```

emmanuelle:
OK I just saw on #239 that it's normal that ``npm run demo`` does not work and that cookiecutter instructions should be followed instead. What is actually the benefit of using cookiecutter here? I understand it is a very clean way of providing a template to be modified later on, but it also adds a new tool/abstraction to the cognitive load, so I wonder whether it would not be easier just to clone a repo which would then be modified.

emmanuelle:
Here a few suggestions for improvements to the tutorial. Please comment so that I can start working on a Pull Request.

**Major points**

[...]

PR #273, opened 2018-11-15
Some clarifications in react for Python developers tutorial

emmanuelle:
Following the discussion in #194.

### DIFF
--- a/tutorial/react_for_python_developers.py
+++ b/tutorial/react_for_python_developers.py
@@ -105,7 +105,7 @@ layout = html.Div([dcc.Markdown('''
   
   - Run the cookiecutter: `cookiecutter https://github.com/plotly/dash-component-boilerplate.git`
   - Answer the questions about the project:
-    - `project_name`: A display name for the project, can contain spaces and uppercase letters, for example `example_component`.
+    - `project_name`: A display name for the project, can contain spaces and uppercase letters, for example `Example Component`.
     - `project_shortname`: A variable derived from project_name without spaces and all lowercase letters.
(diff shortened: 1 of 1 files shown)


emmanuelle:
I think the CI error on python 3.7 is not related to the changes. Could someone with build permission restart the CI?

T4rk1n:
Restarted the workflow

tutorial/react_for_python_developers.py
```diff
@@ -105,7 +105,7 @@
   
   - Run the cookiecutter: `cookiecutter https://github.com/plotly/dash-component-boilerplate.git`
   - Answer the questions about the project:
-    - `project_name`: A display name for the project, can contain spaces and uppercase letters.
+    - `project_name`: A display name for the project, can contain spaces and uppercase letters, for example `example_component`.
```
cldougl: I'm not sure the for example part here adds any value. I'd suggest creating an example that clarifies the previous specification (can contain spaces and uppercase letters) or remove it (as the other steps here don't have examples)

emmanuelle: Yes, I should switch to "Example Component" instead (with spaces and uppercase letters). I added this part because the boilerplate README suggests to use "Dash Core Components" as the name, and in the tutorial the file where the component is defined is called ``ExampleComponent.react.js``.

cldougl: awesome :) I think that clarification is helpful!

emmanuelle:
Merging, thanks @cldougl

Issue #284, opened 2018-11-19
Broken image for lastodash


nicolaskruchten:
@matthewchan15 I think you posted this right?

![image](https://user-images.githubusercontent.com/203523/48740207-850d0180-ec24-11e8-919a-f013723cc61e.png)

matthewchan15:
Hey @nicolaskruchten I did post this https://github.com/plotly/dash-docs/pull/264/files. Looks like the `img_src` went stale!

PR #285, opened 2018-11-20
Fixed stale links in gallery.py by fetching images from assets folder

matthewchan15:
In this pull request I added a new directory in `assets` like so: `assets/image/gallery`. The folder `gallery` holds all the images for `gallery.py`. I also changed all the `img_src=< ... >` to fetch images from `assets/image/gallery` folder.

This PR fixes https://github.com/plotly/dash-docs/issues/284 by replacing the stale link in this issue with the above method.

I was not sure if I should delete the duplicate gallery pictures in the `images` folder located in the root, because they could be used else where - maybe @nicolaskruchten or @chriddyp could let me know if it's okay to delete them?


chriddyp:
> I was not sure if I should delete the duplicate gallery pictures in the images folder located in the root, because they could be used else where 

you can search through the repo to see if they are used. if they are not in the repo, then you can remove them.

PR #343, opened 2018-12-20
add ssh section to dds troubleshooting page

### LINKED ISSUE #313: add ssh debug command to troubleshooting guide
When SSH isn't set up correctly, the server will ask for a password. It can be helpful to run the ssh command in verbose mode:
```
GIT_SSH_COMMAND='ssh -vvvv' git push plotly master
```
This could be added to https://dash.plot.ly/dash-deployment-server/troubleshooting, as a new "SSH" section

michaelbabyn:
resolves #313

### DIFF
--- a/tutorial/dash_deployment_server_examples.py
+++ b/tutorial/dash_deployment_server_examples.py
@@ -2340,77 +2340,6 @@ Troubleshooting = html.Div(children=[
         '''))
     ]),
 
-dcc.Markdown(s('''
-    ***
-
-    #### SSH
-
-    ''')),
-
-    html.Details([
-        html.Summary("git push is asking for password."),
-
-        dcc.SyntaxHighlighter(
-        '''
-        $ git push multipage master
-        dokku@dash.local's password: 
-        ''',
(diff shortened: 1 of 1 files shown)


tutorial/dash_deployment_server_examples.py
```diff
@@ -2340,6 +2340,78 @@ def snapshot_page(n_clicks, url, wait_selector):
         '''))
     ]),
 
+dcc.Markdown(s('''
+    ***
+
+    #### SSH
```
cldougl: I think rather than adding a new section, this could fit under the Deployment Failing section below

tutorial/dash_deployment_server_examples.py
```diff
@@ -2340,6 +2340,78 @@ def snapshot_page(n_clicks, url, wait_selector):
...
+
+    ''')),
+
+    html.Details([
+        html.Summary("git push is asking for password."),
```
cldougl: then this could be something like ssh deploy: git push is asking for password.

PR #357, opened 2019-01-13
Add chapter on integrating Dash with other web apps

### LINKED ISSUE #246: Documenting strategies for embedding Dash apps in an existing Flask app
The community has been asking for a while about strategies for adding a Dash app to an existing Flask app (see plotly/dash#214). I think it would be good if we could get something into the documentation to help with this.

I'd be up for putting together a PR, but first I was hoping to get @plotly/dash's thoughts on the different strategies we could endorse, as well as some ideas about when the different approaches might be useful.

I'm currently aware of two distinct strategies:

1. Passing your `Dash` instance constructor the `server` kwarg that is an existing `Flask` instance as well as passing a value to the `url_base_pathname` kwarg so that multiple Dash apps can sit side by side without clashing. This is described very well in [a comment on the above mentioned issue](https://github.com/plotly/dash/issues/214#issuecomment-434655642).

2. Using Werkzeug's `DispatcherMiddleware` to mount the `Flask` instance of one or more existing `Dash` instances at specific routes of a parent `Flask` instance.  This is discussed in this [Dash community forum thread](https://community.plot.ly/t/hosting-multiple-dash-apps-with-uwsgi-nginx/6758).

My take is that the first option has less moving parts and so is a bit simpler, however the second has a bit more flexibility in that you can embed arbitrary existing Dash apps that can also be run independently (as opposed to requiring an existing flask instance to be passed in). 

Anyone have more strategies to add or thoughts about these two?

ned2:
This chapter provides examples of different approaches to integrating Dash apps with existing Dash apps:

- using a simple <iframe> in an HTML page
- mounting a single Dash app within an existing Flask app
- mounting multiple Dash apps within an existing Flask app
- Using `werkzeug.wsgi.DispatcherMiddleware` to mount multiple Dash apps with an existing WSGI app

Fix #246 and plotly/dash#214

### DIFF
--- a/tutorial/integrating_dash.py
+++ b/tutorial/integrating_dash.py
@@ -25,16 +25,16 @@ layout = html.Div([
     
     ## Using an `<iframe>`
 
-    Perhaps the simplest approach to embedding Dash in an existing web
-    application is to an include an `<iframe>` element in your HTML whose `src`
-    attribute points towards the address of an already running Dash
-    instance. This allows you to place your Dash app in a specific location
-    within an existing web page, with your desired dimensions:""")),
+    The simplest approach to embedding Dash in an existing web application is to
+    an include an `<iframe>` element in your HTML whose `src` attribute points
+    towards the address of an already running Dash instance. This allows you to
+    place your Dash app in a specific location within an existing web page, with
+    your desired dimensions:""")),
     Syntax(dedent("""\
     <iframe src="http://localhost:8050" width=700 height=600>
     """)),
     dcc.Markdown(dedent("""
-    ## Embedding a Dash app within an existing Flask app 
+    ## Embedding a Dash app within an Existing Flask App 
     
     As discussed in the [Deployment Chapter](/deployment), Dash uses the Flask
     web framework under the hood. It is possible to take advantage of this and
@@ -55,8 +55,7 @@ layout = html.Div([
     app = dash.Dash(
         __name__,
         server=server,
-        requests_pathname_prefix='/dash/',
-        routes_pathname_prefix='/dash/',
+        routes_pathname_prefix='/dash/'
     )
 
     app.layout = html.Div("My Dash app")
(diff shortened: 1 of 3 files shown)


ned2:
I've created some tests for these over in plotly/dash#535

tutorial/integrating_dash.py
```diff
@@ -0,0 +1,286 @@
...
+    server = flask.Flask(__name__)
+
+    @server.route('/')
+    def index():
+        return 'Hello Flask app'
+
+    app = dash.Dash(
+        __name__,
+        server=server,
+        requests_pathname_prefix='/dash/',
```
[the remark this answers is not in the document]
ned2: Ah yep, removed

tutorial/integrating_dash.py
```diff
@@ -0,0 +1,286 @@
...
+        __name__,
+        requests_pathname_prefix='/app2/'
+    )
+
+    app.layout = html.Div("Dash app 2")
+    """)),
+    html.Hr(),
+    dcc.Markdown("`wsgi.py`"),
+    Syntax(dedent("""\
+    from werkzeug.wsgi import DispatcherMiddleware
```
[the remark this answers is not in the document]
ned2: OK cool, also removed. Also good from the perspective of not provided an unnecessary amount of options.

ned2:
Thanks @T4rk1n, have made those fixes. Let me know if it looks good now

tutorial/integrating_dash.py
```diff
@@ -0,0 +1,198 @@
...
+    to combine one or more Dash apps with one more existing WSGI apps. With this
+    strategy, your existing web apps don't have to be built on Flask, but
+    instead can be any Python web application implementing the [WSGI
+    specification](https://wsgi.readthedocs.io).
+
+    The following example illustrates this approach. In this example, the
+    existing app being combined with two Dash apps is a Flask app, however this
+    approach enables the combination of [any WSGI
+    app](https://wsgi.readthedocs.io/en/latest/frameworks.html) with multiple
+    Dash apps (or any other WSGI apps for that matter).""")),
```
T4rk1n: Not too sure about the wording here, maybe it should be more focused on `dash`/`flask` apps as `WSGI` and just have a see more section with links.

ned2:
Thanks for the feedback @T4rk1n. I've made some improvements based on your comments. As well as a few other improvements in the content. Could you take another look and see if I've addressed those things?

As this this is adding a whole new chapter, would either @cldougl or @chriddyp be able to have a look?

tutorial/integrating_dash.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+    # Integrating Dash with Existing Web Apps
+
+    This section describes three different approaches to embedding a Dash app
+    within an existing web application.
+    
+    ## Using an `<iframe>`
+
+    The simplest approach to embedding Dash in an existing web application is to
+    an include an `<iframe>` element in your HTML whose `src` attribute points
```
cldougl: 🐐 extra an here

tutorial/integrating_dash.py
```diff
@@ -0,0 +1,219 @@
...
+    towards the address of an already running Dash instance. This allows you to
```
cldougl: maybe instead of: an already running Dash instance I would say deployed Dash app.

ned2: Yep, that works better, thanks

ned2:
Thanks @cldougl! Just waiting for an approving review now

PR #365, opened 2019-01-15
Remove meta tags to remove SEO.

shammamah:


jackparmer:
This is only for the alternate docs URL, correct? Official primary docs URL still needs to have these 

Envoyé de mon iPhone

shammamah:
@jackparmer Yes, this branch will not be merged into `master` and will be the branch from which we deploy to the alternate URL.

chriddyp:
there should also be the `noindex` tag: https://support.google.com/webmasters/answer/93710?hl=en

alexcjohnson:
@shammamah is this PR still relevant?

shammamah:
Nope! I'll close it right now.

PR #370, opened 2019-01-18
Add daq joystick

shammamah:
## About 
* Added examples for the joystick component. (This was initially broken when the PR was opened -- the angle wasn't updating -- but now appears to be fixed.) 
* Added additional examples for joystick label and size. 
* (Pending approval) Reformatted all daq code with `black`.


tutorial/daq.py
```diff
@@ -68,6 +68,10 @@
             'color': '\"#00cc96\"'
         }
     },
+    'Joystick': {
+        'description': '''A joystick that can be used to apply direction and \
+        force.'''
+    },
```
alexcjohnson: not for now since this works fine in the end and it's all over this file, but the string pattern here is weird - it ends up putting it all on one line with a big block of spaces where the line breaks were (which thankfully get ignored by `dcc.Markdown`, but the `dedent` it calls does nothing to remove these spaces because they're mid-line). Better would be something like:
```py
'Joystick': {
    'description': (
        'A joystick that can be used to apply direction and '
        'force.'
    )
}
```

Issue #359, opened 2019-01-13
CircleCI gives timeout error 


shammamah:
CI tests sometimes fail with the following message, even if the demo app on Heroku works fine (seen in #350): 

```
======================================================================
ERROR: test_docs (tests.test_integration.Tests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "tests/test_integration.py", line 126, in test_docs
    visit_and_snapshot(link)
  File "tests/test_integration.py", line 119, in visit_and_snapshot
    self.wait_for_element_by_id('wait-for-page-{}'.format(href))
  File "tests/test_integration.py", line 43, in wait_for_element_by_id
    wait_for(find_element_by_id)
  File "tests/utils.py", line 61, in wait_for
    raise WaitForTimeout(get_message())
WaitForTimeout

----------------------------------------------------------------------
Ran 1 test in 112.681s

FAILED (errors=1)
Exited with code 1
```

xhlulu:
Getting the same error. Could this have to do with Percy snapshots?

shammamah:
The demo apps (and the docs themselves) show as "Application error" for a while after being deployed, and then they start working again, so I think it's something to do with the docs themselves being too large to load quickly (especially since it started after adding Dash DAQ, which was a significant addition).

xhlulu:
This is interesting. Would it be better to schedule merging during low traffic hours in this case? Considering the amount of people visiting the docs we might want to minimize redeployment downtime, depending on how long this is.

cldougl:
closing in favour of https://github.com/plotly/dash-docs/issues/377

PR #375, opened 2019-01-23
rename daq slider example

cldougl:


tutorial/chapter_index.py
```diff
@@ -451,7 +451,7 @@
         'description': 'StopButton examples, properties, and reference.'
     },
 
-    'slider-examples': {
+    'daq-slider-examples': {
```
cldougl: this resolves the dcc.Slider doc that isn't rendering at this time 
screen shot 2019-01-23 at 1 41 29 am

tests/test_integration.py
```diff
@@ -115,11 +115,14 @@ def test_docs(self):
         ]
 
         def visit_and_snapshot(href):
-            self.driver.get('http://localhost:8050{}'.format(href))
-            self.wait_for_element_by_id('wait-for-page-{}'.format(href))
-            time.sleep(5)
-            self.snapshot(href)
-            self.driver.back()
+            if href == '/dash-daq':
```
cldougl: suggestion for resolving https://github.com/plotly/dash-docs/issues/359 as this is timing out on the snapshot of https://dash.plot.ly/dash-daq

Increasing the timeout (https://github.com/plotly/dash-docs/blob/master/tests/utils.py#L4) didn't seem to make a difference. 

@shammamah @chriddyp does anyone have any insight into loading issues with the daq page or why `wait_for_element_by_id()` is failing for this page?

cldougl:
leaning towards merging this asap to get tests back into a useful state + dcc slider reference back up and logging an issue for further investigation on the /dash-daq snapshot failure. 
Any objections? cc @Marc-Andre-Rivet

chriddyp:
> does anyone have any insight into loading issues with the daq page or why wait_for_element_by_id() is failing for this page?

To debug, I'd recommend removing `wait_for_`, adding `time.sleep(10)` and then taking a snapshot. Then, manually inspect the snapshot that got generated: will it be blank, or or partially loaded, or will there be an error?

chriddyp:
I'd also recommend running the tests locally: does the page load correctly and does the ID exist? Running the tests locally uses a different version of the bundles I think (locally vs on cdn), so the page will appear different than if you were to just run the app locally.

chriddyp:
> Any objections?

No objections from me.

shammamah:
💃 Good to have the tests back!

cldougl:
follow up issue https://github.com/plotly/dash-docs/issues/377

PR #379, opened 2019-01-24
added front end pagination

priyatharsan:
closes https://github.com/plotly/dash-table/issues/266


.gitignore
```diff
@@ -6,6 +6,7 @@ coverage/
 dash_core_components/metadata.json
 dash_core_components/bundle.js*
 node_modules/
+cache-directory/
```
cldougl: @priyatharsan why was this modified in this pr?

priyatharsan: It came into my attention that `cache-directory` I suppose a temporary folder popped up and realized that it may not be required in the repo. Thought I could fix it in this PR itself, but I shall revert that and create a new PR? @cldougl

cldougl: yes please :)

PR #382, opened 2019-01-30
it's `records` not `rows`

chriddyp:
rows “happens” to work because pandas only looks at the first letter
(“r”).
- docs here:
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFr
ame.to_dict.html
- community thread here:
https://community.plot.ly/t/data-table-quick-start-example-not-showing-d
ata/18734/2

:see_no_evil:


chriddyp:
Hm, that's odd. The records aren't showing up. In the community forum, this was apparently related to a bug in pandas, so perhaps we need to pin things down more in this repo.
image

eiriklid:
There is a bug in pandas 0.24.0, that will be fixed in the next iteration. Ref: https://github.com/pandas-dev/pandas/issues/24940

The example is working with pandas 0.23.4

PR #384, opened 2019-01-30
Make data mapping examples use absolute paths

### LINKED ISSUE #383: Mapping local directories doc should use absolute path not a relative one.
In this doc https://dash.plot.ly/dash-deployment-server/map-local-directories

michaelbabyn:
resolves #383


michaelbabyn:
cc @cldougl

PR #387, opened 2019-02-03
Add alternative multi-page app project layout and comment about project layout

### LINKED ISSUE #244: Document multi-file structure for Dash projects

ned2:
This adds a second project layout for structuring multi-page apps. Rather than breaking the project up into multiple apps, it breaks up into files for callbacks and layouts, providing some inspiration for alternative project layout strategies. 

It also adds a comment explaining the motivation for the structure that is common to both examples.

Fixes #244

### DIFF
--- a/tutorial/urls.py
+++ b/tutorial/urls.py
@@ -406,5 +406,137 @@ def display_page(pathname):
 
 if __name__ == '__main__':
     app.run_server(debug=True)
-''', language='python', customStyle=styles.code_container)
+''', language='python', customStyle=styles.code_container),
+          dcc.Markdown('''
+
+***
+
+Alternatively, you may prefer a flat project layout with callbacks and layouts
+separated into different files:
+
+```
+- app.py
+- index.py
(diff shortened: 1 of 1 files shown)


CaioEuzebio:
Thank you ned.

PR #388, opened 2019-02-05
Update dash & co

T4rk1n:


tests/IntegrationTests.py
```diff
@@ -59,8 +59,8 @@ def startServer(self, app, path='/'):
         def run():
             # Use CDN so that we don't have to reconfigure percy to find the
             # component assets
```
chriddyp: Let's ⚡️ this comment now. I'm surprised that this is working with percy!

PR #394, opened 2019-02-08
Update dds privacy and authentication

### LINKED ISSUE #390: dds - update privacy and sharing section based on ui additions
https://dash.plot.ly/dash-deployment-server/privacy, https://dash.plot.ly/dash-deployment-server/app-authentication and https://dash.plot.ly/authentication should be updated to showcase: 
- dash auth is no longer needed for on-premise applications
- the dds GUI settings in privacy section:
screen shot 2019-02-05 at 5 28 22 pm

michaelbabyn:
This resolves #390 by simplifying `/dds/privacy` to use the DDS 3.0 UI. 

Since `/dds/app-authentication` is currently mostly about how to edit your app to make `dash-auth` work which is not necessary anymore, I changed `/app-authentication` to be about how to integrate `dash-enterprise-auth` methods into dash apps so they can access user authentication data.

@cldougl, I was also wondering if I should mention the version of DDS and maybe somehow keep a reference to the older docs?


cldougl:
> @cldougl, I was also wondering if I should mention the version of DDS and maybe somehow keep a reference to the older docs?

Yep - let's mention dds 3.0.0+ but say contact support for questions about previous versions.

tutorial/dash_deployment_server_examples.py
```diff
@@ -1539,9 +1539,10 @@ def update_graph(value):
     dcc.Markdown(s('''
     &nbsp;
 
-    You can restrict who is able to view your app from your its
-    management page on the Dash Deployment Server. Find a list
-    of links to these pages for your apps at `https://<your-dash-deployment-server>.com/Manager/apps`.
+    Starting in Version 3.0.0 of Dash Deployment Server, you can restrict who is able to view your app
+    from the app's management page. Find a list of links to these pages for your apps at 
+    `https://<your-dash-deployment-server>.com/Manager/apps`. [Contact support](mailto:onpremise.support@plot.ly)
```
cldougl: I would just say 'contact support' or link to the support section https://dash.plot.ly/dash-deployment-server/support

mailto links aren't always super user-friendly

tutorial/chapter_index.py
```diff
@@ -551,7 +551,7 @@
         'url': '/dash-deployment-server/app-authentication',
         'content': dds_examples.Authentication,
         'name': 'Dash App Authentication',
-        'description': 'Adding Authentication to your Dash App'
+        'description': 'Using Authentication Data in your Dash App'
```
cldougl: I can't see if I previously commented here but why is is Authentication Data in the description? 
I think it might make more since to say something like:
How to modify the authentication settings of your Dash App

michaelbabyn: `dash_enterprise_auth` doesn't modify the authentication settings of your Dash app though it replaces the functionality (i.e the ability to access authentication data) that `dash_auth` users would lose by switching to the builtin DDS authentication model.

cldougl: ok got it - I didn't realize which section this was, in that case I don't think the name: Dash App Authentication is still appropriate. Then it would be worth updating the name here: https://github.com/plotly/dash-docs/pull/394/files#diff-22b19cff76c056678f322a80a608581aR70 as well
Maybe: Dash Enterprise Auth Features 

Also do you think it's worth it to qualify the authentication data with User Authentication data as most of this will be based on user correct?

PR #396, opened 2019-02-08
Import callbacks so that they will be called (Structuring a Multi-Page App flat project layout example)

### LINKED ISSUE #395: Callbacks are not called in Structuring a Multi-Page App flat project layout example
After selecting a dropdown value from either App1 or App2, the selected value is not displayed as intended.

constantlycoding:
Fixes #395 

Currently, callbacks are not called because they are not imported yet.

### DIFF
--- a/tutorial/urls.py
+++ b/tutorial/urls.py
@@ -433,6 +433,8 @@ external_stylesheets = ['https://codepen.io/chriddyp/pen/bWLwgP.css']
 app = dash.Dash(__name__, external_stylesheets=external_stylesheets)
 server = app.server
 app.config.suppress_callback_exceptions = True
+
+import callbacks
 ''', language='python', customStyle=styles.code_container),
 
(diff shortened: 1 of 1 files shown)


ned2:
Thanks so much for the PR!

constantlycoding:
Moved import to `index.py`. Thank you for sorting out the project structure for us!

constantlycoding:
oops

tutorial/urls.py
```diff
@@ -386,7 +386,7 @@ def display_value(value):
 
 from app import app
 from apps import app1, app2
-
+import callbacks
```
ned2: This shouldn't be added to this project structure, just the second flat project structure. Can you remove this one?

constantlycoding: Removed. Was careless.

ned2:
Thanks @constantlycoding!

PR #403, opened 2019-02-18
specify workers via env variables

chriddyp:
- app.json is used by review apps: https://devcenter.heroku.com/articles/github-integration-review-apps#the-app-json-file
- Specify number of workers via env variables instead of hard-coding. `WEB_CONCURRENCY` is used by `gunicorn`: http://docs.gunicorn.org/en/stable/settings.html#workers
- Supply a ENV variable in the heroku UI for production, setting workers to 8

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,1 @@
-web: gunicorn --workers 8 run:server
+web: gunicorn run:server
--- a/app.json
+++ b/app.json
@@ -5,7 +5,8 @@
   "env": {
     "accesstoken": {
       "required": true
-    }
+    },
+    "WEB_CONCURRENCY": 2
   },
   "formation": {
   },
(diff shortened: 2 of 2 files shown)


chriddyp:
Looks like it works
Production on the left, this review app on the right. Production is at ~1GB and this review app is at ~256MB. The review apps are using 2 workers now instead of 8, shrinking the memory by 1/4.

chriddyp:
The only thing to look out for will be whether the new env variable (`WEB_CONCURRENCY: 2`) ends up getting used in the production app. I'll monitor that when I merge and a follow up PR may be necessary.

chriddyp:
looks good on production - `WEB_CONCURRENCY` remains at 8
image

xhlulu:
Does that mean that the number of workers used in prod is specified using the GUI on heroku.com, instead of hardcoded in the procfile?

Thanks for figuring this out!

PR #404, opened 2019-02-18
WIP - DashR organization

chriddyp:
basic architecture for DashR documentation.


dashr/utils.R
```diff
@@ -0,0 +1,42 @@
+library(dashCoreComponents)
+source('app.R')
+
+LoadExampleCode <- function(filename) {
+  # Take a self-contained DashR example filename,
+  # eval it, and return that example's `layout`
+  # and the source code.
+  example.file.as.string <- readChar(filename, file.info(filename)$size);
```
rpkyle: It's a minor item, but you can omit the semicolon from this line.

dashr/utils.R
```diff
@@ -0,0 +1,42 @@
...
+
+  # modify the example code so that it can run within
+  # the context of an already running app
+  example.ready.for.eval <- example.file.as.string
```
rpkyle: I don't know if it's a style guide thing, but I usually avoid naming variables and objects using periods as separators/word boundaries. It's pretty clear that the objects here are not functions, but just the same I'd rather reserve this syntax for S3 methods:

http://adv-r.had.co.nz/S3.html

dashr/utils.R
```diff
@@ -0,0 +1,42 @@
...
+  eval(parse(text=example.ready.for.eval))
+
+  return(list(
+    layout=htmlDiv(className='example-container', children=layout),
+    source_code=htmlDiv(
+      children=coreSyntaxHighlighter(example.file.as.string),
+      className='code-container'
+    )
+  ))
+}
```
rpkyle: @alexcjohnson convinced me that inserting a single blank line at the end of R scripts is a good idea; I'm going to try to get into the habit.

index.R
```diff
@@ -0,0 +1,87 @@
...
+chapters.callbacks <- new.env()
+source('dashr/chapters/callbacks/index.R', local=chapters.callbacks)
+
+
+# Temporary workaround until `assets/` are loaded
+# This will serve the CSS files in `assets` from a separate
+# webserver.
+# In a separate terminal, run
+# python2 -m SimpleHTTPServer 8000
+css.files = list.files('assets')
```
rpkyle: I'd use `<-` instead of `=` here.

dashr/utils.R
chriddyp: > I'm going to try to get into the habit.

This is a configurable setting for most editors - "Add newline to end of file on save"

dashr/utils.R
chriddyp: OK cool, I'll use camelCase instead? Took the `.` separated variable names from the Google R styleguide:
```
variable.name is preferred, variableName is accepted
              
 GOOD: avg.clicks
              
 OK: avgClicks
              
 BAD: avg_Clicks
```

rpkyle: I might be in the minority on this topic, but I’ve also noticed that the Tidyverse types seem to use snake case for some function names:

https://style.tidyverse.org/syntax.html

Ha! Until now I’d not read their style guide closely, but they also agree on the dot separator thing. Take that, Google.

>Base R uses dots in function names (`contrib.url()`) and class names (`data.frame`), but it’s better to reserve dots exclusively for the S3 object system. In S3, methods are given the name `function.class`; if you also use . in function and class names, you end up with confusing methods like `as.data.frame.data.frame()`.

dashr/utils.R
rpkyle: Good point. I’ll go ahead and enable this in Atom/Pycharm/RStudio.

dashr/utils.R
rpkyle: Hmm. If we’re considering adding `lintr` checks to the R repos, maybe we should follow the Tidyverse style guide. They started with the Google guide, but addressed a few other topics it didn’t cover.

dashr/utils.R
TahiriNadia: I’ll go also configure my environment.

dashr/utils.R
TahiriNadia: The strategy of the variable is mostly underscoring if I see `R Style Guide` by `tidyverse`. 
 There are 4-5 naming conventions in R asked in R communauty). But `.` is confusing (object oriented programming languages). So the variable is not so clear define. We need to decide? I think `_` will be the best choice.

TahiriNadia: The debate: https://www.r-bloggers.com/consistent-naming-conventions-in-r/

index.R
TahiriNadia: We need also to use `"` instead of `'`  see https://style.tidyverse.org/syntax.html#assignment

TahiriNadia: For assignment use `<-` instead of `=` except on the function def 
`PredictCTR <- function(query, property, num.days, show.plot = TRUE)`

dashr/chapters/installation/index.R
```diff
@@ -0,0 +1,63 @@
...
+You can see what version of R you have by entering version in the R CLI. 
+CRAN is the easiest place to download the latest R version.
+"),
+  
+dccMarkdown("
+```{r}
+library(devtools) # devtools: Tools to Make Developing R Packages Easier  
+install_github('plotly/dashR') # The core dash backend
+install_github('plotly/dash-html-components') # HTML components   
+install_github('plotly/dash-dcc-components') # Supercharged components
```
alexcjohnson: `install_github('plotly/dash-core-components')`

PR #407, opened 2019-02-22
Fix typo

ycaokris:


ycaokris:
@chriddyp  would you please review it? Thanks

Issue #400, opened 2019-02-13
Dash User Guide - "Announcement Essay" chapter link is not a `target="_blank"` 


byronz:
Since this href is an external link to a medium page, the experience is not consistent as other "Chapters components".  

this should be rendered as <a> with `target="_blank"`, but I found out that the `ddc.Link` doesn't expose this like `html.A`.

PR #412, opened 2019-02-25
fix the json.org url

byronz:
without `http`, the site will be rendered as local path. 
but all these external links in markdown will have similar issues as mentioned in #400


byronz:
verified in staging, the fix is effective

tutorial/plugins.py
```diff
@@ -32,7 +32,7 @@
 compatible with the Dash ecosystem.
 
 On a high level, this is how that works:
-- Components in dash are serialized as [JSON](www.json.org).
+- Components in dash are serialized as [JSON](http://www.json.org/).
```
cldougl: can we use the https link? https://www.json.org/

byronz: yes, I just copied from browser, both https and http work, and https is better

PR #414, opened 2019-02-25
update "which input has changed" for callback_context

alexcjohnson:
Docs update for https://github.com/plotly/dash/pull/608
screen shot 2019-02-25 at 3 00 15 pm

### DIFF
--- a/tutorial/examples/faqs/last_clicked_button.py
+++ b/tutorial/examples/faqs/last_clicked_button.py
@@ -1,3 +1,5 @@
+import json
+
 import dash
 import dash_html_components as html
 from dash.dependencies import Input, Output
@@ -5,31 +7,38 @@ from dash.dependencies import Input, Output
 app = dash.Dash(__name__)
 
 app.layout = html.Div([
-    html.Button('Button 1', id='btn-1', n_clicks_timestamp='0'),
-    html.Button('Button 2', id='btn-2', n_clicks_timestamp='0'),
-    html.Button('Button 3', id='btn-3', n_clicks_timestamp='0'),
+    html.Button('Button 1', id='btn-1'),
+    html.Button('Button 2', id='btn-2'),
+    html.Button('Button 3', id='btn-3'),
     html.Div(id='container')
 ])
 
 
 @app.callback(Output('container', 'children'),
-              [Input('btn-1', 'n_clicks_timestamp'),
-               Input('btn-2', 'n_clicks_timestamp'),
-               Input('btn-3', 'n_clicks_timestamp')])
+              [Input('btn-1', 'n_clicks'),
+               Input('btn-2', 'n_clicks'),
+               Input('btn-3', 'n_clicks')])
 def display(btn1, btn2, btn3):
-    if int(btn1) > int(btn2) and int(btn1) > int(btn3):
-        msg = 'Button 1 was most recently clicked'
-    elif int(btn2) > int(btn1) and int(btn2) > int(btn3):
-        msg = 'Button 2 was most recently clicked'
-    elif int(btn3) > int(btn1) and int(btn3) > int(btn2):
-        msg = 'Button 3 was most recently clicked'
-    else:
+    ctx = dash.callback_context
+
+    if not ctx.triggered:
         msg = 'None of the buttons have been clicked yet'
+    else:
+        button_num = ctx.triggered[0]['prop_id'].split('.')[0].split('-')[1]
+        msg = 'Button {} was most recently clicked'.format(button_num)
+
+    ctx_msg = json.dumps({
+        'states': ctx.states,
+        'triggered': ctx.triggered,
+        'inputs': ctx.inputs
+    }, indent=2)
+
     return html.Div([
         html.Div('btn1: {}'.format(btn1)),
         html.Div('btn2: {}'.format(btn2)),
         html.Div('btn3: {}'.format(btn3)),
-        html.Div(msg)
+        html.Div(msg),
+        html.Pre(ctx_msg)
     ])
 
 
--- a/tutorial/examples/faqs/last_clicked_button.py
+++ b/tutorial/examples/faqs/last_clicked_button.py
@@ -22,10 +22,9 @@ def display(btn1, btn2, btn3):
     ctx = dash.callback_context
 
     if not ctx.triggered:
-        msg = 'None of the buttons have been clicked yet'
+        button_id = 'No clicks yet'
     else:
-        button_num = ctx.triggered[0]['prop_id'].split('.')[0].split('-')[1]
-        msg = 'Button {} was most recently clicked'.format(button_num)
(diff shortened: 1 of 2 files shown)


tutorial/examples/faqs/last_clicked_button.py
```diff
@@ -1,35 +1,44 @@
...
-        msg = 'Button 2 was most recently clicked'
-    elif int(btn3) > int(btn1) and int(btn3) > int(btn2):
-        msg = 'Button 3 was most recently clicked'
-    else:
+    ctx = dash.callback_context
+
+    if not ctx.triggered:
         msg = 'None of the buttons have been clicked yet'
+    else:
+        button_num = ctx.triggered[0]['prop_id'].split('.')[0].split('-')[1]
```
T4rk1n: Just get the last character:

```suggestion
        button_num = ctx.triggered[0]['prop_id'][-1]
```

tutorial/faqs.py
```diff
@@ -62,42 +62,48 @@ class names of your components. Both `dash-html-components` and
...
-    times a component has been clicked), all `dash-html-components` have an
-    `n_clicks_timestamp` property, which records the time that the component was
-    last clicked. This provides a convenient way for detecting which
-    `html.Button` was clicked in order to trigger the current callback. Here's
-    an example of how this can be done:''')),
+    **A:** *New in v0.38.0!* In addition to event properties like `n_clicks`
+    that change whenever an event happens (in this case a click), there is a
+    global variable `dash.callback_context`, available only inside a callback.
+    It has properties:
+    - `triggered`: list of changed properties. Usually this will be empty on
```
T4rk1n: Not usually, always empty with the initial none callbacks, after that it's a list of size 1+.

tutorial/faqs.py
```diff
@@ -62,42 +62,48 @@ class names of your components. Both `dash-html-components` and
...
+      initial load, and a length-1 list later. With more complex callback
+      chains it's possible that multiple intermediate properties change and
+      trigger a callback.
```
T4rk1n: Maybe some more information is needed for that since the callback request is initiated with `setProps`, the number of properties that changes are dependent on the call and the subscribed inputs. Most of the times where that happens are with the timestamps.

tutorial/faqs.py
```diff
@@ -62,42 +62,48 @@ class names of your components. Both `dash-html-components` and
...
+    - `inputs` and `states`: allow you to access the callback params
+      by id and prop instead of through the function args.
+
```
T4rk1n: I would add the syntax for the `prop_id`: `component_id.prop_name`.

tutorial/faqs.py
```diff
@@ -62,42 +62,48 @@ class names of your components. Both `dash-html-components` and
...
 
-    Note that `n_clicks` is the only property that has this timestamp
-    property. We will add general support for "determining which input changed"
-    in the future, you can track our progress in this [GitHub
-    Issue](https://github.com/plotly/dash/issues/291).
-    
+    Prior to v0.38.0, you needed to compare timestamp properties like
+    `n_clicks_timestamp` to find the most recent click. While existing uses of
+    `*_timestamp` continue to work for now, this approach is deprecated, and
+    may be removed in a future update.
```
T4rk1n: The `modified_timestamp` from the Store component should be excluded from the deprecation notice.

tutorial/examples/faqs/last_clicked_button.py
alexcjohnson: That doesn't quite work, the number is in the middle. I suppose I could do
`button_num = ctx.triggered[0]['prop_id'].split('.')[0][-1]`
but that only saves a few characters, at the expense of generality in case someone extends this to 10+ buttons

tutorial/faqs.py
alexcjohnson: Except if the inputs are themselves outputs of some other callback - I don't quite understand this behavior, but this is a case where I thought we might get length-2, but instead we get just one of the two inputs:
```py
app.layout = html.Div([
    html.Button('a1', id='a1'),
    html.Button('a2', id='a2'),
    html.Button('b1', id='b1'),
    html.Button('b2', id='b2'),
    html.Div(id='a'),
    html.Div(id='b'),
    html.Div(id='c')
])

@app.callback(Output('a', 'children'), [Input('a1', 'n_clicks'), Input('a2', 'n_clicks')])
def a(a1, a2):
    return str((a1 or 0) + (a2 or 0))

@app.callback(Output('b', 'children'), [Input('b1', 'n_clicks'), Input('b2', 'n_clicks')])
def b(b1, b2):
    return str((b1 or 0) + (b2 or 0))

@app.callback(Output('c', 'children'), [Input('a', 'children'), Input('b', 'children')])
def c(a, b):
    return 'a: {}, b: {}, triggered: {}'.format(a, b, json.dumps(dash.callback_context.triggered))
```
The initial render looks like:
screen shot 2019-02-25 at 5 27 25 pm
<hr>
In fact if I tweak callback `b` to depend on `a1` and `b2`, clicking button `a1` updates both `a` and `b`, but I still only get one (always `a.children`) listed as the trigger for `c`. Is that as expected? I would have thought that this would be the case for a length-2 `triggered`.

tutorial/examples/faqs/last_clicked_button.py
byronz: regex way might be cleaner. 'btn-(\d+)\.', but I feel you can instead of getting the button number, just render the first part of  `btn-3.n_clicks` would be enough for demonstration.

tutorial/faqs.py
byronz: do we using the `warnings` module for that purpose?

alexcjohnson: We haven't gotten that far, AFAIK. I'm tempted to say once we release v1.0 we just treat any removal as a breaking change requiring a major bump, which means `warnings` would only be useful within the 0.x series, leaving us a (hopefully) very short time to employ it!

tutorial/examples/faqs/last_clicked_button.py
alexcjohnson: > just render the first part of `btn-3.n_clicks` would be enough for demonstration.

I like that - the id `btn-3` is what's most useful for developers -> https://github.com/plotly/dash-docs/pull/414/commits/0c5d158632fc833b994f374f4b39328045e468f0

alexcjohnson: (also made a pretty little table in that commit ^^)

alexcjohnson:
@T4rk1n see what you think of my updated descriptions in https://github.com/plotly/dash-docs/pull/414/commits/6263ee2932656bc15a1ec6bcf96f1c1c3f68814b. I didn't go into detail about `setProps`, on the theory that at this point in someone's dash education they probably won't know what that means, and anyway as discussed offline we may want to change this behavior in the future.

PR #415, opened 2019-02-25
Delete a bunch of obsolete files

alexcjohnson:
noticed because they reference `Event()` - @Marc-Andre-Rivet had me worried that we were still somehow trying to use `Event`, and we're not, but that means all these files referencing them are unused and can be removed.


byronz:
image
@alexcjohnson  that's a good sign!

PR #416, opened 2019-02-26
Update dash-table documentation with typing/validation/formatting/presentation

Marc-Andre-Rivet:
This is a sibling PR for https://github.com/plotly/dash-table/pull/377.

Adds documentation for typing, validation, formatting and presentation.
Also modifies the existing dropdown examples so they work with the new presentation API.

These items need to be completed before merging:
- [x] https://github.com/plotly/dash-table/pull/377 is merged
- [x] ...and released
- [x] requirements.txt updated with new version of dash / rc versions removed
- [x] rc versions removed from the PR


tutorial/table/table_typing_chapter.py
```diff
@@ -0,0 +1,132 @@
...
+        className='example-container'
+    ),
+
+    section_title('DataTable with formatting'),
+    dcc.Markdown(dedent(
+    '''
+    This table contains columns with type `numeric` and `datetime`. The "max" columns have the default
+    behavior and will not allow for invalid data to be passed in. The "min" columns are more permissive.
+    The "Min Temperature (F)" column will default invalid entries to `None` and display "N/A". The "Min
+    Temperature (Date)" column will validate the data but let invalid input go through as-is.
```
cldougl: >will validate the data but let invalid input go through as-is.

this may warrant a bit more explanation - what's the point of validating if you let it go ie does it return a warning or something?

tutorial/table/table_typing_chapter.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Both temperature columns are using the Format helper object to create the desired formatting. The
+    equivalent manual configuration is shown below. One can always see the resulting configuration for a
```
cldougl: are the manual configurations the commented out parts? If so I would say:
The equivalent manual configurations are shown as comments in the code below.

tutorial/examples/table/typing_formatting.py
```diff
@@ -0,0 +1,88 @@
...
+            'id': 'max',
+            'name': u'Max Temperature (˚F)',
+            'type': 'numeric',
+            'format': Format(
+                precision=0,
+                scheme=Scheme.fixed,
+                symbol=Symbol.yes,
+                symbol_suffix=u'˚F'
+            ),
+            # 'format': {
```
cldougl: might be helpful to include inline `# example manual configuration` on these comments - just for those readers that are more focused on the code examples rather than text explanation.

run.py
```diff
@@ -294,6 +294,7 @@ def display_content(pathname):
         <footer>
             {%config%}
             {%scripts%}
+            {%renderer%}
```
cldougl: @Marc-Andre-Rivet is this now needed to do the update in renderer?

Marc-Andre-Rivet: Yes. `external_css_and_js.py` has similar variables, wondering if the docs need to be updated there

Marc-Andre-Rivet: https://github.com/plotly/dash-docs/blob/770913d043d4259abb3df9eea19f89e4cfc2bd7f/tutorial/external_css_and_js.py#L319

I guess it does :)

Marc-Andre-Rivet: @cldougl The custom hook PR does contain a usage update -- would defer to that PR for a complete docs update of the feature. https://github.com/plotly/dash-docs/pull/419/files#diff-918f458c6a61b531c2522e31e2d0f9abR379

PR #418, opened 2019-02-26
Prevent slider marks where no data exists

timesler:
The default slider behaviour is to put a step at every year, rather than where data to plot actually exists. As a result, people completing this tutorial may think they have made an coding error when they see the data disappear when the slider is moved.


timesler:
Not sure who is best to review this, sorry for for the nag @chriddyp

PR #419, opened 2019-03-01
Docs for custom hooks and standalone dash-renderer

valentijnnieman:
This will add a section to the `external-resources` page about using your own `DashRenderer` in Dash by providing it in your custom index html, and how to set the `hooks` object for that to provide request hooks.


tutorial/external_css_and_js.py
```diff
@@ -410,6 +426,39 @@
...
+    dcc.SyntaxHighlighter(
+        examples['custom-dash-renderer-hooks'],
+        language='python',
+        customStyle=styles.code_container
+    ),
+
+    dcc.Markdown(s('''
+    Notice the `request_pre` function takes the payload of the request being sent as its argument, and the `request_post` fuction takes both the payload and the response of the server
+    as arguments. In the example above, the `request_pre` function is fired before each server call, and in the case of this example, it will `console.log()` the request parameter. The
+    `request_post` function will fire __after__ each server call, and in our example will also print out the response parameter.
```
alexcjohnson: We should say something here about being able to modify the payload (in `request_pre`) or the response (in `request_post` - you could modify the payload there too but it's a little late for that to matter!).

It would be nice if we could give this some practical context... @chriddyp can you shed any light here? The only things I can think of are:
- `request_pre` might add extra info to the payload, like, I don't know... some sort of extra auth managed by the front end? What timezone the user is in?
- The callback might be a simple passthrough to some other non-dash API - so `request_pre` would transform the payload into the format needed by that other API, and `request_post` could transform the response to what dash-renderer is expecting. Why you wouldn't just do this on the back end I don't know...

Even better would be if we could actually *use* this capability in the example, to do something more than just inspect the values and pass them through.

valentijnnieman: Yeah I was having a hard time thinking of examples for this too. I also figured that perhaps having a very basic example of how to set up the hooks would be good - without any other code in the actual hook functions - to not make the example more complicated than it should be. If a user wants to use this, then they probably have an idea of what they would want to do with it?

alexcjohnson: If we can't come up with a simple example then fine, the existing example will do. We should definitely describe the ability to mutate the args though.

tutorial/examples/external_css_and_js/custom-dash-renderer.py
```diff
@@ -0,0 +1,35 @@
...
+    </head>
+    <body>
+        <div>Testing custom DashRenderer</div>
+        {%app_entry%}
+        <footer>
+            {%config%}
+            {%scripts%}
+            <script id="_dash-renderer" type="application/javascript">
+                var renderer = new DashRenderer();
+            </script>
```
alexcjohnson: I don't know that this example adds much. But it could be nice to show overriding the renderer script *without* overriding the whole index.

tutorial/external_css_and_js.py
chriddyp: Yeah, a simple example is tough. Of the use cases that the sponsor was using this for was trimming down the request payload by modifying the props on either end. Think dash datatable and the `data` property: suppose you only needed a single column's worth of data instead of all 10. If you were OK with writing some JS and digging deep into the request, you could modify the datatable's `data` property before it was sent to the server and then modify on its way back.

This is pretty deep behaviour though and I don't think that we really want to encourage it in the community as the support burden will be high. It's more of an "escape hatch".

tutorial/examples/external_css_and_js/custom-index-string.py
```diff
@@ -20,6 +20,7 @@
         <footer>
             {%config%}
             {%scripts%}
+            {%renderer%}
```
Marc-Andre-Rivet: Some of this is already in https://github.com/plotly/dash-docs/pull/424 and https://github.com/plotly/dash-docs/pull/416 as I ran into these issues while updating other docs.

tutorial/external_css_and_js.py
```diff
@@ -399,6 +414,7 @@
     'favicon': '',
     'metas': '<meta charset="UTF-8"/>',
     'scripts': '<script src="https://unpkg.com/react@15.4.2/dist/react.min.js"></script>\\n<script src="https://unpkg.com/react-dom@15.4.2/dist/react-dom.min.js"></script>\\n<script src="https://unpkg.com/dash-html-components@0.11.0/dash_html_components/bundle.js"></script>\\n<script src="https://unpkg.com/dash-renderer@0.13.0/dash_renderer/bundle.js"></script>',
```
Marc-Andre-Rivet: While updating the docs in this section, bump up the versions for html and renderer to latest!

Issue #417, opened 2019-02-26
spike - improvement on visual regression


byronz:
the current *percy* visual regression has several issues in terms of execution time and false alert(noise):

1. the snapshot is mostly taken at the top level of each chapter route, which produces a super long snapshot image. it's not obvious to catch the minor diff
2. there are some hardcoded `time.sleep(5)` which should be replaced by implicit_wait 
3. there is no easy way to ignore certain flakey area by either css selector or fixed region coordinates, need to investigate more on percy docs, or trial on **applitools**
4. lack of IE testings, @cldougl mentioned ideally we should test at least the last two *edge* versions

@plotly/dash-core comments are welcome

- [x] change the datetime.now() case in demo   #423 #425 
- [x] reduce the execution time with `driver.implicitly_wait(2)  #427

alexcjohnson:
most of the flaky areas, as far as I've seen, come from using `datetime.now()`. We should just replace that with a fixed value like `datetime(2019, 2, 26, 12, 0, 0, 0)`. More verbose but at least deterministic!

xhlulu:
Percy is currenly unable to render canvas element, since it is not part of the DOM. Enabling JS when taking the snapshot does not seem to solve the problem. If you visit the actual page, you can see that there's a network where the blank box is: https://dash.plot.ly/cytoscape

Here's what the Percy snapshot looks like:
![image](https://storage.googleapis.com/percy-images/03d3ecab93251bccc005c60b5d6fb838dc41e25c34744cb5072f5c3cf6e31506)

wbrgss:
Re: canvas elements, this is what the Percy Dev told me (excepted from a longer conversation...):

> 02:06 PM | Robert from Percy: Just wanted to let you know I've been throwing stuff at the wall over here with my python to try and get something working and I haven't quite yet either. 
> 
> 02:07 PM | Will Burgess: OK, thanks, I really appreciate it. Unfortunately the product (Dash) uses Python/React to render content and its method of serving assets (with Flask) is a bit complicated
> 
> 02:09 PM | Will Burgess: I am close to getting something that's "good enough", but two main problems I have are 1.) nested DOM elements are rendered out of order and 2.) html <canvas> elements don't render properly
> 
> 02:11 PM | Robert from Percy: Ohhh, nested DOM elements being rendered out of order is really interesting. Is there any CSS (flexbox) that changes their order visually? 
> 
> 02:11 PM | Will Burgess: yeah for sure
> 
[...]

PR #423, opened 2019-03-04
use first dash date to avoid visual diff

byronz:
one minor improvement proposed in  #417

### DIFF
--- a/tutorial/core_component_examples.py
+++ b/tutorial/core_component_examples.py
@@ -650,7 +650,7 @@ DatePickerRange = html.Div(children=[
 from datetime import datetime as dt
 
 dcc.DatePickerRange(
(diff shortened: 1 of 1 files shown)


byronz:
> This is a silly place to be testing something like that, these are the docs so we should be using the components more "correctly." But we should modify one or two of the dcc tests to include a nonzero time component to fill in this gap. Would you please make either an issue or a PR for this?

you mean sth like dt(2017,6,21,23,21)?

alexcjohnson:
> you mean sth like dt(2017,6,21,23,21)?

Yes, in fact keep going. dt(2027,6,21,23,21,53,748912) - the time part should be ignored, just want to make sure it leads to the correct date and doesn’t throw an error.

Issue #425, opened 2019-03-05
DatePicker examples with edges cases and more explanation on date


byronz:
this issue is created following the discussion in  #423.

we will need to add few edge cases as static date and add better explanation for date fields

PR #426, opened 2019-03-05
#425 add edge cases and docs

byronz:
- add few edges case like the hour boundaries, leap year, string date
- the single and range has slightly different explanation regarding the date, and I think we should recommend users to use `datetime.date` object as it's purely about date. I also like the simple string solution as 'YYYY-MM-DD'.  this will need a fix in DCC repo

```
Specifies the starting date for the component, best practice is to pass value via datetime object

Specifies the starting date for the component. Accepts datetime.datetime objects or strings in the format 'YYYY-MM-DD'
```

### DIFF
--- a/tutorial/core_component_examples.py
+++ b/tutorial/core_component_examples.py
@@ -793,14 +793,12 @@ DatePickerSingle = html.Div(children=[
     html.Hr(),
     html.H3("Simple DatePickerSingle Example"),
     dcc.Markdown("This is a simple example of a `DatePickerSingle` \
-                component tied to a callback. The 'date` property is recommended \
-                to use python `datetime.date` or `datetime.datetime` objects.\
-                The `min_date_allowed` and \
-                `max_date_allowed` properties define the minimum and \
-                maximum selectable \
-                dates on the calendar while `initial_visible_month` defines \
-                the calendar month that is first displayed when the \
-                `DatePickerSingle` component is opened."),
+                component tied to a callback. The `date` property is recommended \
+                to use python `datetime.date` or `datetime.datetime` objects. \
+                The `min_date_allowed` and `max_date_allowed` properties \
+                define the minimum and maximum selectable dates on the calendar \
+                while `initial_visible_month` defines the calendar month that is first \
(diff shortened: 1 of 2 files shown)


alexcjohnson:
I guess it's useful for people to know they can include time in their `datetime` and we'll ignore it - but let's describe this behavior when more than a pure date is provided.

Re: `datetime` objects vs strings - this I'm a bit conflicted about actually. Callbacks (currently anyway) always send back strings for the dates you choose. So do we really want to say it's best practice to provide a value that's not what you're going to get back later? Of course we could figure out a way to have the return value automatically converted to a `datetime` - though this would be a substantial breaking change - but barring that self-consistency seems like it argues for strings being the "best practice" (thoughts @chriddyp?)

chriddyp:
Yeah that's a good point. For all the `value*` properties, they will surely be passed back into the callback and so I'd agree with you that `str(datetime.datetime.now())` would probably be the best way to specify those arguments.

byronz:
@alexcjohnson  see if the changes reflect above discussions

tutorial/core_component_examples.py
```diff
@@ -793,12 +793,13 @@
...
-                 component tied to a callback. The `min_date_allowed` and \
-                 `max_date_allowed` properties define the minimum and \
-                 maximum selectable \
-                 dates on the calendar while `initial_visible_month` defines \
-                 the calendar month that is first displayed when the \
-                 `DatePickerSingle` component is opened."),
+        component tied to a callback. It is safe to use the string \
+        format of `datetime.date` or `datetime.datetime` objects for the \
+        `date` property, be aware of the extra time in `datetime` object will \
+        be ignored. The `min_date_allowed` and `max_date_allowed` properties \
```
alexcjohnson: Let's be a little more explicit, how about:

You can use either date objects (`datetime.date` or `datetime.datetime`) or strings in the form 'YYYY-MM-DD' to provide dates to Dash components. Strings are preferred because that's the form dates take as callback arguments. Be aware that any time information included in a `datetime` object or string will be ignored.

PR #427, opened 2019-03-05
#417 Replace unnecessary hard-code wait with implicitly_wait at driver level

byronz:
the changes has been tested several times both locally and with percy snapshots. 

- the circleci jobs are twice faster than before now

image

- percy duration drop from ~7m20s to ~2m12s

image


alexcjohnson:
@byronz I hadn't looked at the percy diffs for this PR before, but one of them - search-dropdown - I think may actually need the longer delay. The text typed in the search box (dropdown) doesn't match the text that was actually searched (drop, as you can see from the italics and extra search results that don't show up in the previous screenshot)
screen shot 2019-03-07 at 12 41 10 pm

PR #428, opened 2019-03-06
[Dash 0.41] Update react-markdown in dcc, docs impact

Marc-Andre-Rivet:
Relates to https://github.com/plotly/dash-core-components/pull/492

Running against a version of dcc using the latest version of react-markdown (4.0.6) and making things match as much as possible. Do look at the percy diffs. Some slight alignment changes and URLs styling mostly.


tutorial/getting_started_part_1.py
```diff
@@ -72,52 +72,53 @@
...
-    Give it a try: change the title "Hello Dash" in your application or change the `x` or the `y` data. Your app should auto-refresh with your change.
-
-    > Don't like hot-reloading? You can turn this off with `app.run_server(dev_tools_hot_reload=False)`. 
-    > Learn more in [Dash Dev Tools documentation](/devtools) Questions? See the [community forum hot reloading discussion](https://community.plot.ly/t/announcing-hot-reload/14177).
-
-    #### More about HTML
-
-    The `dash_html_components` library contains a component class for every
-    HTML tag as well as keyword arguments for all of the HTML arguments.
+    Note:
```
alexcjohnson: This is still using `.replace('    ', '')` - so the result is probably collapsed quite a lot more to the left than intended.

Let's keep the indentation deeper and more consistent, then switch the `replace` to `dedent` - which this file already imports as `from textwrap import dedent as s`

alexcjohnson: I'm tempted in fact to rip out *all* the `.replace('    ', '')` in this repo, since it doesn't do what it was supposed to do (and `dedent` does)

tutorial/react_for_python_developers.py
```diff
@@ -2,745 +2,749 @@
 import dash_html_components as html
 
 layout = html.Div([dcc.Markdown('''
-  # React for Python Developers: a primer
+# React for Python Developers: a primer
```
alexcjohnson: Did anything change here besides indentation? I actually like the unindented text here, since this one string is the whole file.

alexcjohnson:
This markdown block (in `/dash-html-components`) was supposed to be shown as an app output, but it now looks like part of the prose.
Screen Shot 2019-03-19 at 2 57 32 PM

alexcjohnson:
This one (from `DatePickerSingle`) was not intended to change. Not a huge deal, but it seems square brackets are being interpreted as delimiting a code block? is that something we can turn off?
Screen Shot 2019-03-19 at 3 06 20 PM

alexcjohnson:
Same thing happened in `DatePickerRange` but those are the only two. NBD. I definitely appreciate the automatic URL detection!

Marc-Andre-Rivet:
https://github.com/plotly/dash-docs/pull/428#issuecomment-474525816 --> Ah! Didn't realize this was intentional. I thought it was a bad interpretation.

Not sure about the indentation change. Maybe VSCode auto formatted and I didn't realize.

In all cases, checking.

tutorial/react_for_python_developers.py
Marc-Andre-Rivet: Indentation only.

You can use the diff settings above to help filter those out:
image

Marc-Andre-Rivet:
After a long hiatus, coming back to this clean up task..

Marc-Andre-Rivet:
- This is running against patched versions of dash, renderer and dcc -- will need to make a patch release + update this PR to be truly mergeable
- One significant diff left but it's not there when running the app locally. Looks like a timing issue from the snapshot.
image

alexcjohnson:
> One significant diff

My guess https://github.com/plotly/dash-core-components/pull/516#issuecomment-481427777

> kinda weird percy diffs - just a bunch of new whitespace at the bottom of the page. I'm going to guess it's just percy doing something weird with the extra internal structure of a plotly.js graph added in plotly/plotly.js#3589... looks fine locally, I approved the diff.

alexcjohnson:
https://github.com/plotly/dash-docs/pull/428#issuecomment-474525816 (lost app block in dash-html-components) is still open. I'll open an issue about removing `.replace('    ', '')` repo-wide.

alexcjohnson:
subsumed into #477

PR #429, opened 2019-03-07
add extra wait for dropdown search #427

byronz:


byronz:
@alexcjohnson  I tested with 1s and 2s delay, and switch back to 1s. 

also can you see if #426 is good to :dancer: ?

PR #438, opened 2019-03-13
Multi output docs

### LINKED ISSUE #432: Document Multi-Output callbacks
In particular:

* Getting Started Part 2 now incorrectly states that callbacks can only update a single `Output` and provides the old solution of defining multiple callbacks
* The FAQs also has a gotcha that states, "Callbacks can only target a single Output component/property pair". I think we can just remove this?

alexcjohnson:
Closes #432 - did I miss anything?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,5 +18,6 @@ cache
 bundle.js
 .idea
 .env
+.mypy_cache/
 
 ignore/indicators.csv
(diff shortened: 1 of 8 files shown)


ned2:
Looks pretty good!

ned2:
:dancer:

The comment warning about the downsides of too many Outputs is a good addition!

PR #439, opened 2019-03-15
[Dash 0.40] Removing 'setProps undefined` from gotchas and react-for-python-devs

Marc-Andre-Rivet:
With https://github.com/plotly/dash-renderer/pull/126, the gotchas and additional instructions are moot.

To be pushed when Dash 0.40.0 is released.

### DIFF
--- a/tutorial/faqs.py
+++ b/tutorial/faqs.py
@@ -217,6 +217,21 @@ layout = html.Div([
     `Inputs` or `States` is determined by a user's input, then you must
     pre-define up front every permutation of callback that a user can
     potentially trigger. For an example of how this can be done programmatically
-    using the `callback` decorator, see this [Dash Community forum
-    post](https://community.plot.ly/t/callback-for-dynamically-created-graph/5511). '''))
+  using the `callback` decorator, see this [Dash Community forum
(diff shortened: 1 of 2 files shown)


alexcjohnson:
For both of these pieces - the usage gotcha and the component developer's note - I'm a little worried that by simply having this info disappear we'll leave existing devs confused. We'll describe the simplification in the forum of course, but I wonder if it would be worthwhile having a short note where these problems used to be described, along the lines of:

> Note: prior to dash v0.40.0 it was important to connect all core components to callbacks. That issue is now resolved.

and something similar in the component dev section. Could be removed after a few months, just seems like it could avoid confusion in the short term.

Thoughts @chriddyp ?

chriddyp:
Yeah, I agree. We could even link to the forum discussion, something like:

> Note: Prior to v0.40.0, `setProps` was only defined if the component was connected to a callback. This required complex state management within the component [like this](https://github.com/plotly/dash-core-components/blob/63be1c258c15b419a82fb66366e1a31e66fbfaef/src/components/Input.react.js#L51-L59). Now, `setProps` is _always_ defined which should simplify your component's state management. [Learn more in this community forum topic](...)

PR #440, opened 2019-03-15
add anchors to section titles on dash gallery

jdamiba:


chriddyp:
Note that if you are doing this in order to provide direct links to a certain part of the page, this won't work. Dash loads its content client-side and so the content isn't visible on the initial page load when the browser looks for the ID to scroll to.

alexcjohnson:
Would be an awesome feature to add though! Might need to be careful to only scroll on initial draw when the element is added, not if it gets added by a user interaction.

PR #442, opened 2019-03-19
Update Username Field Type

blozano824:
This helps the Input field render properly since its type was previous not specified and is necessary for the correct CSS to be applied to the element..


alexcjohnson:
Good catch @blozano824, thanks! This changes it from unstyled:
Screen Shot 2019-03-19 at 1 22 36 AM
to styled:
Screen Shot 2019-03-19 at 1 21 59 AM
@chriddyp we could probably do this more generally by adding `input:not([type])` anywhere we have `input[type="text"]` in [base.css](https://github.com/plotly/dash-docs/blob/master/assets/base.css). Alternatively we could make `'text'` the default for the `type` prop of `dcc.Input`, which is what a missing `type` means anyway, it just doesn't match css selectors that look for an explicit type. Thoughts?

PR #444, opened 2019-03-19
[Dash 0.39] Remove explicit html/dcc/table/renderer installation

Marc-Andre-Rivet:
Fixes https://github.com/plotly/dash-docs/issues/443.

With Dash version locking https://github.com/plotly/dash-core/issues/24 the documentation is now out of line with the preferred practice of only installing `dash` instead of installing all packages individually. Removing explicit installs for dcc, html, renderer and table.

### DIFF
--- a/tutorial/installation.py
+++ b/tutorial/installation.py
@@ -23,14 +23,16 @@ layout = html.Div([
 
     dcc.SyntaxHighlighter('''pip install dash=={}  # The core dash backend
         pip install dash-daq=={}  # DAQ components (newly open-sourced!)
-
-        **Note**: starting with dash 0.37.0, dash automatically installs dash-renderer, dash-core-components,
-        dash-html-components, and dash-table, using known-compatible versions of each. You need not and
-        should not install these separately any longer, only dash itself.
     '''.replace('    ', '').format(
         dash.__version__,
         dash_daq.__version__
     ), customStyle=styles.code_container),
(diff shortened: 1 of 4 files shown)


tutorial/installation.py
```diff
@@ -22,15 +22,9 @@
     Python 2 and 3 are supported.'''.replace('    ', '')),
 
     dcc.SyntaxHighlighter('''pip install dash=={}  # The core dash backend
-        pip install dash-html-components=={}  # HTML components
-        pip install dash-core-components=={}  # Supercharged components
-        pip install dash-table=={}  # Interactive DataTable component (new!)
```
alexcjohnson: It would be nice to add a note here about how we used to do this and why (and when) it changed. Something like

*Note: starting with dash 0.37.0, dash automatically installs dash-renderer, dash-core-components, dash-html-components, and dash-table, using known-compatible versions of each. You need not and should not install these separately any longer, only dash itself.*

Marc-Andre-Rivet: Good point.

alexcjohnson: looks funny in the code block - I'd either put it as regular text outside the code, or if you want to leave it in the code, make it a comment.

chriddyp: We've been using the quote markup (`>` ) for these types of "aside remarks" too

PR #455, opened 2019-04-01
New chapter on dash-canvas: images annotations and processing

emmanuelle:
This is a new chapter of the documentation on the ``dash-canvas`` module. 

For those who have access to the playground server, it is deployed on https://dash-playground.plotly.host/emma-dash-docs/ (component libraries section). 

For the deployment I had to switch to python 3.6.7. Was it on purpose that the doc was deployed using on Python 2.7? I could revert to Python 2.7 with some workaround if needed. 

I hope it is ready for review.


emmanuelle:
I saw in the existing documentation that functions exist in the repo to generate the html documentation of the API for Dash components (via a table of their props). Do similar functions exist for other kinds of objects like functions? In ``dash-canvas`` there is the ``DashCanvas`` component but also a lot of helper functions for manipulating annotations, and their docstrings should appear somewhere. For another package I would just use sphinx and put the reference API online like on pythonhosted but what would you recommend for a package based on dash?

shammamah:
@emmanuelle The docs do run on Python 2.7 (https://github.com/plotly/dash-docs/blob/master/runtime.txt). What kinds of workarounds did you need to implement?

tutorial/canvas.py
```diff
@@ -0,0 +1,158 @@
...
+
+    dcc.Markdown('''
+    You can draw inside the object with the freehand tool, and use the tool
+    buttons to draw lines, zoom in and out, pan, select objects and move them
+    inside the canvas.
+
+    ``DashCanvas`` comes with a set of properties which can be adjusted to
+    control the geometry of the canvas, the default tool and its properties.
+    You can pass a background image either as a filename (``filename``
+    property) or as a data string (``image_content`` property, more examples
```
shammamah:
```suggestion
    property) or as a data string (``image_content`` property); more examples
    
```

tutorial/canvas.py
```diff
@@ -0,0 +1,158 @@
...
+    below).
+    '''.replace('  ', '')),
```
shammamah: I think you can actually use `textwrap.dedent` here (https://docs.python.org/2/library/textwrap.html#textwrap.dedent)-- it's also used in the `dash-core-components` documentation (https://github.com/plotly/dash-docs/blob/78313a8aa4793b0935e594cdf12e93e916252db9/tutorial/core_components.py#L5).

tutorial/canvas.py
```diff
@@ -0,0 +1,158 @@
...
+    ),
+    html.Div(examples[2][1], className='example-container'),
+    dcc.Markdown('''
+    In the example above, a slider ``dcc.Slider`` and a color picker
+    ``daq.ColorPicker`` are used to adjust the width and color of the
+    drawing brush. We just created an image coloring tool in a few lines
+    of code! Also note that the set of available buttons has been restricted
+    through the ``hide_buttons`` properties, in order to keep the app design
+    simple. You can learn more about available components in the
+    [Component libraries](https://dash.plot.ly/) section of the Dash
```
shammamah:
```suggestion
    [component libraries](https://dash.plot.ly/) section of the Dash
```

tutorial/canvas.py
```diff
@@ -0,0 +1,158 @@
...
+          customStyle=styles.code_container
+    ),
+    html.Div(examples[2][1], className='example-container'),
+    dcc.Markdown('''
+    In the example above, a slider ``dcc.Slider`` and a color picker
+    ``daq.ColorPicker`` are used to adjust the width and color of the
+    drawing brush. We just created an image coloring tool in a few lines
+    of code! Also note that the set of available buttons has been restricted
+    through the ``hide_buttons`` properties, in order to keep the app design
+    simple. You can learn more about available components in the
```
shammamah: You could probably move this sentence so that it's right after the sentence in which you mention `dcc.Slider` and `daq.ColorPicker`!

tutorial/canvas.py
```diff
@@ -0,0 +1,158 @@
...
+    [Component libraries](https://dash.plot.ly/) section of the Dash
+    documentation.
+
+    ### Retrieving the geometry of annotations and using utility functions
+
+    The geometry of annotations can be retrieved by pressing the bottom-right
+    button of the ``DashCanvas``. This button is called "Save" by default,
```
shammamah:
```suggestion
    button of the ``DashCanvas``. This button is called "Save" by default;
```

tutorial/canvas.py
```diff
@@ -0,0 +1,158 @@
...
+    dcc.SyntaxHighlighter(
+          examples[3][0],
+          language='python',
+          customStyle=styles.code_container
+    ),
+    html.Div(examples[3][1], className='example-container'),
+
+    dcc.Markdown('''
+    You can either write custom functions to parse the JSON string, or
+    use utility functions of DashCanvas. In particular, `
```
shammamah:
```suggestion
    use the utility functions included in the `dash_canvas` package. In particular, `
```

emmanuelle:
Thank you @shammamah for your review! I've implemented most changes, I'm still struggling with the textwrap suggestion (tests failing with python 2.7). 

The reason why I moved to Python 3 was some problem with some scikit-image function, probably because new versions of scikit-image are now compatible with Python 3 only. I should investigate more and make sure dash-canvas is compatible with Python 2.7 and Python 3.6, but I will have to make a new release of dash-canvas for this.

emmanuelle:
@shammamah I fixed the python 2 problem with a release of dash-canvas and reverted the deployment to python 2.7 as it was before. I think I have now addressed all your suggestions. Thanks a lot!

emmanuelle:
Merging, thank you @shammamah

PR #458, opened 2019-04-04
V1

TahiriNadia:


README.md
```diff
@@ -1,41 +1,46 @@
-# Dash Userguide
+# Deploying to Heroku
```
alexcjohnson: Currently README is replacing the Python one with R - we certainly don't want to lose the Python one, do we want a single file with two sections? Or two separate files?

app.R
```diff
@@ -0,0 +1,13 @@
...
+
+css_list <- list("https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css",
+                 "https://dash.plot.ly/assets/base.css?m=1552356970.0",
+                 "https://dash.plot.ly/assets/daq.css?m=1552356970.0",
+                 "https://dash.plot.ly/assets/dosis.css?m=1552356970.0",
+                 "https://dash.plot.ly/assets/local-css-example.css?m=1552356970.0",
+                 "https://dash.plot.ly/assets/open-sans.css?m=1552356970.0",
+                 "https://dash.plot.ly/assets/override.css?m=1552356970.0",
+                 "https://dash.plot.ly/assets/pdf-docs.css?m=1552356970.0",
+                 "https://dash.plot.ly/assets/tabs-styled-with-classes.css?m=1552356970.0")
```
alexcjohnson: Can `/assets` be removed here, since we're serving them automatically?

assets/open-sans.css
```diff
@@ -1,3 +1,5 @@
+external_stylesheets
+
```
alexcjohnson: 🔪 ?

PR #462, opened 2019-04-09
object_hook parameter in json.loads function

TahiriNadia:
I'm trying to generate props docs using the idea of python version
Here it's snipped the lines
https://github.com/plotly/dash-docs/blob/833fb677a0488243d477aaf0a2a10ad5fdad6bec/tutorial/utils/convert_props_to_table.py#L134-L180
If I understand clearly about `object_hook`, the goal is to convert `json` format to `dict` format in Python, it's true or it's something I missed?

Here the small example of `object_hook` in `Python`:

```
import json
from collections import namedtuple

data = '{"name": "John Smith", "hometown": {"name": "New York", "id": 123}}'

# Parse JSON into an object with attributes corresponding to dict keys.
x = json.loads(data, object_hook=lambda d: namedtuple('X', d.keys())(*d.values()))
print x.name, x.hometown.name, x.hometown.id
```

The result is:
```
python object_hook.py
John Smith New York 123

Process finished with exit code 0
```


rpkyle:
Discussed this offline with @alexcjohnson, I believe the comment above relates to generation of props tables on the R side. 

Following @chriddyp's request to do this via parsing of R's online help pages, I wrote a function entitled `props_to_list` which serves a similar purpose to `convert_props_to_table`. 

I'll open a PR to merge `props_to_list` into the R utilities we've been using to construct the `dash-docs`, link here, and then close this PR if it doesn't have any content we've not already updated.

PR #464, opened 2019-04-11
Revert "Update iv tracer screenshot"

chriddyp:
Reverts plotly/dash-docs#450

@ycaokris  - It appears that this PR caused the main documentation site to go down today. I have rolled back the deploy on heroku and will now revert this PR officially on master so that heroku & github remain in sync.

Then, in a separate PR, we can investigate what might've caused the docs to fail.


ycaokris:
@chriddyp  Thanks! not sure why this happens, I'll open another branch to work on.

PR #466, opened 2019-04-11
preload flag

chriddyp:


Procfile
```diff
@@ -1 +1 @@
-web: gunicorn run:server
+web: gunicorn run:server --preload
```
josegonzalez: The flag needs to go before the module being executed, so the correct form is:

```
gunicorn --preload run:server
```

PR #473, opened 2019-04-17
Gallery app maintainence

ycaokris:
- Fix invalid app link for dash-daq LED control
- Update lastodash screencast


ycaokris:
@chriddyp  would you please proceed with deployment? Thanks

PR #477, opened 2019-04-22
Dev Tools prererelease with markdown fix

chriddyp:
https://github.com/plotly/dash-docs/pull/476 with https://github.com/plotly/dash-docs/pull/428 merged into it.


chriddyp:
This pretty much looks good to me.

Looks like dropdowns are slightly more narrow - I'm fine with that.
Screen Shot 2019-04-22 at 2 56 23 PM

Missing borders around this codeblock. I'll fix that now.
Screen Shot 2019-04-22 at 2 54 12 PM

chriddyp:
The border around the markdown example should get fixed by https://github.com/plotly/dash-core-components/pull/539

tutorial/examples/canvas_color.py
```diff
@@ -29,7 +29,7 @@
                 min=2,
                 max=40,
                 step=1,
-                value=[5]
+                value=5
```
byronz: :cat2:

PR #479, opened 2019-04-22
replace props table with component docstring

chriddyp:
Previously, our documentation constructed it’s own table derived from the component’s `metadata.json`.

This resulted in a lot of duplicate code: we do the same `metadata.json` parsing when we generate our components.

Rather than maintaining two separate `metadata.json` parsers, I’d rather just maintain one, in `dash` or `dash-component-system`, and make that system really good.

This was prompted by a recent change in `dash` that needed to add support for the `exact` prop type (https://github.com/plotly/dash/pull/692).

Besides maintenance, the table format for prop types doesn’t scale well for nested properties.

### DIFF
--- a/tutorial/utils/convert_props_to_table.py
+++ b/tutorial/utils/convert_props_to_table.py
@@ -6,227 +6,29 @@ import dash_core_components as dcc
 import dash_cytoscape
 import dash_table
 import pandas as pd
-
-def js_to_py_type(type_object):
-    js_type_name = type_object if (type(type_object) is str) else type_object['name']
-
-    # wrapping everything in lambda to prevent immediate execution
-    js_to_py_types = {
-        'array': lambda: 'list',
-        'bool': lambda: 'boolean',
-        'func': lambda: 'func',
-        'number': lambda: 'number',
-        'string': lambda: 'string',
-        'component_name': lambda: 'component_name',
-        'object': lambda: 'dict',
-
-        'any': lambda: 'boolean | number | component_name | dict | list',
-        'element': lambda: 'dash component',
-        'node': lambda: (
-            'a list of or a singular dash component, component_name or number'
-        ),
-
-        # React's PropTypes.oneOf
-        'enum': lambda: 'one of: {}'.format(', '.join([
-            '{}'.format(str(t['value'])) for t in type_object['value']
-        ])),
-
-        # React's PropTypes.oneOfType
-        'union': lambda: '{}'.format(' | '.join([
-            '{}'.format(js_to_py_type(subType))
-            for subType in type_object['value'] if js_to_py_type(subType) != ''
-        ])),
-
-        # React's PropTypes.arrayOf
-        'arrayOf': lambda: 'list {}'.format(
-            'of {}'.format(js_to_py_type(type_object['value']))
-            if (js_to_py_type(type_object['value']) != '')
-            else ''
-        ),
-
-        # React's PropTypes.objectOf
-        'objectOf': lambda: (
-            'dict with component_names as keys and values of type {}'
-        ).format(js_to_py_type(type_object['value'])),
-
-
-        # React's PropTypes.shape
-        'shape': lambda: (
-            'dict containing key(s): {}\n{}'.format(
-                ', '.join(
-                    ["'{}'".format(t) for t in list(type_object['value'].keys())]
-                ),
-                '\n. Those keys have the following types: \n{}'.format(
-                    '\n'.join([
-                        '  - ' + (argument_doc(
-                            prop_name,
-                            prop,
-                            '' if (type(prop) is str) else prop.get('description', '')
-                        )) for
-                        prop_name, prop in list(type_object['value'].items())
-                    ])
-                )
-            )
-        )
-    }
-
-    if 'computed' in type_object and type_object['computed']:
-        return ''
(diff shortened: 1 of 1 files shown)


alexcjohnson:
@chriddyp @shammamah on merging master this PR fails because dash-bio was using `js_to_py_type` that this PR removed. Do we still want to continue with this strategy?

```
  File "/home/circleci/project/tutorial/utils/dashbio_docs.py", line 12, in <module>
    from tutorial.utils.convert_props_to_table import js_to_py_type
ImportError: cannot import name 'js_to_py_type'
```

chriddyp:
Yeah - I think we should continue with this strategy and future effort into the docstring generation so that it's consistent across all environments.

alexcjohnson:
OK good - @shammamah can you take over this PR and update the `dash-bio` parts to match?

shammamah:
@alexcjohnson @chriddyp This has been taken care of in 997e53f. I'm a little concerned that the default values aren't shown in this new description (I've personally found them really helpful in the past). 

Could we also maybe rename the function to `generate_prop_info`? 

It could also be helpful to format each prop name, description, type, etc. in a structured way that isn't necessarily a table (maybe a bullet-point list), so that people can find information faster. This was the main reason I would use the docs instead of looking at the docstrings myself. What do you think?

alexcjohnson:
> I'm a little concerned that the default values aren't shown in this new description (I've personally found them really helpful in the past).

We're pretty scattered about how these defaults are handled right now. Lots of them are useless (do we really need to tell people the default for an optional object is `{}`? But if we were to *always* include defaults I wouldn't mind), lots of them are missing (when `undefined` substitutes for a falsy default) but I agree with you that it's really nice to have meaningful defaults in the docs. But I think per the original motivation for this PR, the right solution is to include defaults in the docstring and make sure the formatting looks good there. Perhaps instead of including the word "optional" we just put the default? So required params wouldn't change, they'd stay like:
- label (string | number; required): The checkbox's label

but optional items would say:
- disabled (boolean; default False): If true, ...

> Could we also maybe rename the function to `generate_prop_info`?

Yes please, good catch!

> It could also be helpful to format each prop name, description, type, etc. in a structured way that isn't necessarily a table (maybe a bullet-point list), so that people can find information faster.

Isn't that (a bullet list) what we have with this PR? I wouldn't mind doing a little bit of touchup here though, like maybe bold code for prop names, bold for type & default?
- **`disabled` (boolean; default False):** If true, ...

> This was the main reason I would use the docs instead of looking at the docstrings myself.

Ideally people are primarily coming here to get examples, or if they simply *prefer* not to use docstrings, not because the docstrings are insufficient. Per the new function:
> Access this documentation in your Python terminal with:
> ```
> >>> help({}.{})
> ```

shammamah:
> But I think per the original motivation for this PR, the right solution is to include defaults in the docstring and make sure the formatting looks good there. Perhaps instead of including the word "optional" we just put the default? 

I think that would be a good idea! 

>Isn't that (a bullet list) what we have with this PR? I wouldn't mind doing a little bit of touchup here though, like maybe bold code for prop names, bold for type & default?

Yes, that's what I meant -- sorry for not being more clear about this. 

>Ideally people are primarily coming here to get examples, or if they simply prefer not to use docstrings, not because the docstrings are insufficient. 

Fair enough! 


I can take care of these items (as well as renaming the function), if nobody else wants to.

shammamah:
Small note - I'm currently adding specific formatting for default values, prop names, etc. in tandem with the following: https://github.com/plotly/dash/pull/782. This has not yet been merged, so the changes won't be quite reflected yet.

chriddyp:
Just noticed that our existing props-to-table isn't showing info about objects inside arrays like `dash_table.DataTable.columns`:
image


So, it'd be great to get this in at least just for this page!

shammamah:
@alexcjohnson @Marc-Andre-Rivet I've added in the tarball of the Dash Table build after https://github.com/plotly/dash-table/pull/534 was merged in, and the results are here: https://dash-docs-pr-479.herokuapp.com/datatable/reference

tutorial/utils/convert_props_to_list.py
```diff
@@ -9,7 +9,7 @@ def generate_prop_info(component_name, lib=dcc):
     component = getattr(lib, component_name)
     component_doc = component.__doc__
 
-    regex = r'''^(.*?)\s*\(([^;]*);*\s*([\w\s'"().{}:,;#-\[\]]*)\):\s*(.*?)\s*$'''
+    regex = r'''^([^\(]*)\s*\(([^;]*);\s*([\W\w]+?)\):\s*(.*?)\s*$'''
```
alexcjohnson: is there a difference between `[\W\w]` and `.`? According to https://docs.python.org/3/library/re.html:
```
\W
    Matches any character which is not a word character. This is the opposite of \w.
```

shammamah: There's not really a difference. I've switched to `.` in a9c1f55 because it looks simpler and is easier to understand.

cclauss:
These changes created an _undefined name_ in the following code.
```
./dash_docs/tutorial/utils/dashbio_docs.py:491:25: F821 undefined name 'js_to_py_type'
            prop_type = js_to_py_type(props[prop]['type'])
                        ^
```

PR #496, opened 2019-05-09
Update markdown for store example accordingly to #493

StefanTippelt:
Post-merge checklist:

- [x] I understand

### DIFF
--- a/tutorial/core_component_examples.py
+++ b/tutorial/core_component_examples.py
@@ -1223,10 +1223,11 @@ ConfirmDialogProvider = html.Div([
 
 Store = html.Div([
     html.H1('Store component'),
-    dcc.Markdown(s('''
+    dcc.Markdown(
+        s('''
     Store json data in the browser.
 
-    ## limitations.
+    ## limitations
 
     - The store will not work properly if there is no callback associated.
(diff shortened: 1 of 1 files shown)


alexcjohnson:
Ah thanks! But this begs the question, why doesn't this part use the pattern elsewhere of generating the syntax block from the same example file?

https://github.com/plotly/dash-docs/blob/eaa0cbe97bcf3614137013d198a683ff8c6cea20/tutorial/core_component_examples.py#L1218-L1219

Probably the answer is just that we wrote this example before we thought of `tools.load_example`, but @StefanTippelt would you like to have a go of getting rid of this duplication?

StefanTippelt:
Thanks @alexcjohnson for the hint, I updated it accordingly.

alexcjohnson:
@StefanTippelt apologies for letting this one slip through. I see what was going on, `Store` components need to be in the layout from the beginning, so the actual executed code needs to not have these, but the syntax needs them or it won't make sense. I'm going to close this PR and clean this up in a new PR. Thanks for the original fix #493 and for pointing out this follow-on!

StefanTippelt:
Thanks @alexcjohnson for the follow up!

PR #504, opened 2019-05-23
Add mobile meta tag example + comments to explain what they do

wbrgss:
I wanted to have an example of a (perhaps overlooked) meta tag for mobile support to send to Dash users. I also explained what the other meta tags in this example do.

Post-merge checklist:

- [x] I understand


wbrgss:
Percy diffs look like typical noise, except for [the first one](https://percy.io/plotly/dash-docs/builds/1924388/view/125227661/1280?mode=diff&browser=firefox&snapshot=125227661) which is kind of confusing...could be that the baseline is outdated.

PR #513, opened 2019-06-05
Make examples idiomatic

z267xu:
DashR docs (except chapter 5)

- fix docs typo
- make example code idiomatic

TODO: 

- make sure all links work well
- In chapter 6, **Why global variables will break your app** part needs to rewrite, or at least we need to find a alternative `Gunicore` in `R`; "Example 3" and "Example 4" are hidden, since `flask_cashing` is not available so far.
- In chapter 7, I need to have a double check, remove some `python` components
- dashTable


rpkyle:
#513 was duplicate of #514; closed.

PR #520, opened 2019-06-07
adding dash testing - MERGE it after 1.0 is released

byronz:
Post-merge checklist:

- [x] I understand


tutorial/dash_test_sample.py
```diff
@@ -0,0 +1,29 @@
...
+
+	# 6. most of the time, you can use #6 to replace #5
+	assert dash_duo.find_element('#nully-wrapper').text == '0'
+
+	# 7. use python builtin assert statement, and describe your
+    # criteria as an assertion message after the comma.
+	assert dash_duo.get_logs() == [], "browser console contain no error"
+
+	# 8. visual testing with percy snapshot
+	dash_duo.percy_snapshot("bsly001-layout")
```
alexcjohnson: hard tabs? 🤔

tutorial/dash_test_sample.py
```diff
@@ -0,0 +1,29 @@
...
+
+    # 3. define your app inside the test function
+    app = dash.Dash(__name__)
+    app.layout = html.Div(id='nully-wrapper', children=0)
+
+	# 4. host the app locally in a thread
+    dash_duo.start_server(app)
+
+    # 5. use wait_for_* only when it's necessary
+    # e.g. the test result is flakey and an explicitly wait might be helpful
```
alexcjohnson: honestly I think this is much more common than "if the test is flaky" - basically any time the value you're looking for is the result of a callback - even in the initial callback chain - I would like us to recommend `wait_for_*`. If you're checking a *second* thing that is expected to have been updated along with or prior to this one (multi-output, or several attrs/elements that got returned by the same callback), that can use form # 6

byronz: agree on the callback thing, one thing I would avoid here is to make people feel they should place the wait_for_* everywhere, and/or check unnecessary text compare which they can simple do with find_element, as we have a 2 seconds margin on each action. I will change the comments here

tutorial/dash_test_sample.py
```diff
@@ -0,0 +1,29 @@
...
+
+    # 3. define your app inside the test function
+    app = dash.Dash(__name__)
+    app.layout = html.Div(id="nully-wrapper", children=0)
+
+    # 4. host the app locally in a thread
+    dash_duo.start_server(app)
+
+    # 5. use wait_for_* if your target element is the result of a callback,
+    # keep in mind even the initial rendering is an return value of a callback
```
alexcjohnson:
```suggestion
    # keep in mind even the initial rendering can trigger callbacks
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    Syntax(code),
+    dcc.Markdown(dedent("""
+
+    **Notes**:
+
+    * #1 For most test scenarios, you don't need to import any modules for
+    the test, just import what you need for the dash app itself.
+
+    * #2 A test case is a regular python function, the function name follows
+    this pattern `test_{tcid}_{test tile}`. The `tcid` (test case ID) is
```
```suggestion
    this pattern `test_{tcid}_{test title}`. The `tcid` (test case ID) is
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    statement to do assertions. It's good practice to expose your
+    acceptance criteria directly in the test case rather than wrapping the
+    assert inside another helper API. By looking at the test case, reviewers
+    and maintainers should be able easily figure out the purpose of the test by
+    the test title, the app definition, the actions and the check points.
+
+    * #8 We use [Percy](https://percy.io/) as our *Visual Regression Testing*
+    tool, it's a good alternative besides assertions when your check point is
+    about a graph diagram and/or an app layout. We integrate the percy service
+    with a `PERCY_TOKEN` varaible,  so the regression result is only
```
```suggestion
    with a `PERCY_TOKEN` variable,  so the regression result is only
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+        2. A selenium webdriver is initialized and it navigates to the
+        local server with the `server_url`
+        3. It waits until the page has loaded in the browser
+
+    * #5 A test case is composed of preparation, actions and check points.
+    Both #5 and #6 are doing the same check in this example, we are expecting
+    that the defined `Div` component's text is identical to the `children`'s
+    content. #5 will wait the expected state to reach within an explicitly 4
+    seconds timeout. It's a safer way to use when you are doing an element
+    check related to callbacks, as it happens at lot under dash context,
```
shammamah:
```suggestion
    check related to callbacks, as it is usually performed in the Dash context;
```

(not sure if "usually" is what you meant by "a lot")

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    the element is already present in the DOM, but not necessarily the props.
+
+    * #6 The `find_element` API call has an implicitly global timeout set at
```
shammamah:
```suggestion
    * #6 The `find_element` API call has an implicitly global timeout of two seconds set at the
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    driver level, i.e. the driver waits at most 2 seconds to find the element
```
shammamah:
```suggestion
    driver level, i.e. the driver waits at most two seconds to find the element
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    by the locator, **BUT** it will compare the texts as soon as the driver
```
shammamah:
```suggestion
    by the selector. **Howver**, it will compare the text as soon as the driver
```

(not sure of you meant "selector" by "locator")

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    returns an element.  Also note that the `find_element('#nully-wrapper')`
```
shammamah:
```suggestion
    returns an element.  Also note that the function `find_element('#nully-wrapper')`
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    * #7 Unlike `unittest`, `pytest` uses the native python
+    [`assert`](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement)
+    statement to do assertions. It's good practice to expose your
+    acceptance criteria directly in the test case rather than wrapping the
+    assert inside another helper API. By looking at the test case, reviewers
+    and maintainers should be able easily figure out the purpose of the test by
+    the test title, the app definition, the actions and the check points.
+
+    * #8 We use [Percy](https://percy.io/) as our *Visual Regression Testing*
+    tool, it's a good alternative besides assertions when your check point is
```
shammamah:
```suggestion
    tool. It's a good alternative to assertions when your check point is
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    about a graph diagram and/or an app layout. We integrate the percy service
```
shammamah:
```suggestion
    related to graphical aspects of the app, such as the layout or the appearance of any graphs. We integrate the percy service
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    available in our circleci setup.
+
+    ## Fixtures
+
+    To avoid accidental name collision with other pytest plugins,
+    all dash test fixtures start with the prefix `dash_`
+
+    - dash_duo
+
+    The default fixture for dash integration tests, it contains a
```
shammamah:
```suggestion
    The default fixture for dash integration tests. It contains a
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    thread_server and a webdriver wrapped with high level dash testing APIs
```
shammamah:
```suggestion
    `thread_server` and a webdriver wrapped with high-level Dash testing APIs
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+
+    - dash_br
+
+    A standalone webdriver wrapped with high-level dash tesing APIs. this is
```
shammamah:
```suggestion
    A standalone webdriver wrapped with high-level dash testing APIs. This is
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    suitable for testing a dash app in a deployed environment, i.e. your dash
```
shammamah:
```suggestion
    suitable for testing a dash app in a deployed environment, i.e. one in which your Dash
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    app is accessible from a URL
+
+    - dash_thread_server
+
+    Start your dash app locally in a python `threading.Thread`, lighter and
```
shammamah:
```suggestion
    Start your dash app locally in a python `threading.Thread`, which is lighter and
```

tutorial/testing.py
shammamah: It might be a good idea to end all of these with periods, too (since some of them contain complete sentences already)!

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    faster than a process
+
+    - dash_process_server
+
+    Start your dash app with `waitress` in a python `subprocess`, this is
```
shammamah:
```suggestion
    Start your dash app with `waitress` in a python `subprocess`. This is
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    close to your production/deployed environment.  **Note:**  *you need to
+    configure you PYTHONPATH so that your dash app source file is
```
shammamah:
```suggestion
    configure your `PYTHONPATH` so that your Dash app source file is
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    directly importable*
+
+    ## APIs
+
+    ### Selenium Overview
+
+    Both `dash_duo` and `dash_br` expose the  selenium webdriver via the
+    property `driver`, e.g. dash_duo.driver, which you have the full access to
```
shammamah:
```suggestion
    property `driver`, e.g. dash_duo.driver, which gives you full access to
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    the [Python Selenium API](https://selenium-python.readthedocs.io/api.html).
+    (*Note that this is not the official selenium documentation site, but has
+    somehow become the defacto python community reference*)
```
shammamah: 👍 for "somehow" :P

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+
+    One of the core routine in selenium testing is to find the **web element**
```
shammamah:
```suggestion
    One of the core components of selenium testing is finding the **web element**
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    with a `locator`, and perform some actions like `click`, `send_keys` on it,
```
shammamah:
```suggestion
    with a `locator`, performing some actions like `click` or `send_keys` on it,
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    wait and verify if the expected state is met after actions, the check is
```
shammamah:
```suggestion
    and waiting to verify if the expected state is met after those actions. The check is
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    considered as an acceptance criteria, which you can write in built-in
```
shammamah:
```suggestion
    considered as an acceptance criterion, for which you can write in a built-in
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    python `assert` statement.
+
+    #### Element Locators
+
+    There are several strategies to
+    [locate elements](https://selenium-python.readthedocs.io/locating-elements.html#locating-elements),
```
shammamah:
```suggestion
    [locate elements](https://selenium-python.readthedocs.io/locating-elements.html#locating-elements);
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    css selector and xpath are two versatile ways. We recommend to use
```
shammamah:
```suggestion
    CSS selectors and xpath are two versatile ways. We recommend using
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    **CSS Selector** in most of the case due to its
```
shammamah:
```suggestion
    **CSS Selectors** in most cases due to their
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    [performance and robustness](http://elementalselenium.com/tips/34-xpath-vs-css-revisited-2) across browsers.
+
+    If you are new at using CSS Selector, these
```
shammamah:
```suggestion
    If you are new at using CSS Selectors, these
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    [Saucelab tips](https://saucelabs.com/resources/articles/selenium-tips-css-selectors)
+    is a great start, and remember that
```
shammamah:
```suggestion
    are a great start. Also, remember that
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    [Chrome Dev Tools Console](https://developers.google.com/web/tools/chrome-devtools/console/utilities)
+    is always your good friend and playground.
+
+    ![dev tools](https://user-images.githubusercontent.com/1394467/59371148-3505a180-8d12-11e9-9ad6-7dce223b6019.png)
+
+    #### Waits
+
+    [This link](https://selenium-python.readthedocs.io/waits.html) has a
```
shammamah:
```suggestion
    [This link](https://selenium-python.readthedocs.io/waits.html) covers
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    nice coverage on this topic. For impatient readers, a quick take away is
```
shammamah:
```suggestion
    this topic nicely. For impatient readers, a quick take away is
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    quoted as follow:
```
shammamah:
```suggestion
    quoted as follows:
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+
+    Selenium Webdriver provides two types of waits
```
shammamah:
```suggestion
    The Selenium Webdriver provides two types of waits:
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+
+    - **explicit wait**
+        Makes webdriver wait for a certain condition to occur before
+        proceeding further with execution. All our APIs with `wait_for_*`
+        falls into this category.
+    - **implicit wait**
+        Makes webdriver poll the DOM for a certain amount of time when trying
+        to locate an element. We set a global 2 seconds timeout at
```
shammamah:
```suggestion
        to locate an element. We set a global two-second timeout at
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+        `driver` level.
```
shammamah:
```suggestion
        the `driver` level.
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    | `session_id` | shortcut to `driver.session_id` |
+    | `server_url` | set the server_url as setter so the selenium is aware of the local server port, it also implicitly calls `wait_for_page`. return the server_url as property |
+
+
+    ## Debugging
+
+    ### Verify your test environment
+
+    If you run the integration in a virtual environment, make sure you are
+    getting the latest commit in **master** branch from each component, and
```
shammamah:
```suggestion
    getting the latest commit in the **master** branch from each component, and
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    the installed `pip` versions are correct.
```
shammamah:
```suggestion
    that the installed `pip` versions are correct.
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+
+    note: We have some enhancement initiatives tracking in this [issue](https://github.com/plotly/dash/issues/759)
+
+    ### Run the CI job locally
+
+    The [CircleCI Local Cli](https://circleci.com/docs/2.0/local-cli/) is a
```
shammamah:
```suggestion
    The [CircleCI Local CLI](https://circleci.com/docs/2.0/local-cli/) is a
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    handy tool to run all the jobs locally. It gives you an earlier warning
+    before even pushing your commits to remote,  which leaves no chance of
+    making an embarrasing public exposure. The environment is identical to the
```
shammamah:
```suggestion
    making an embarrassing public mistake. The environment is identical to the
```

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+    remote one except the percy snapshot and test reports are not functional
```
shammamah:
```suggestion
    remote one, except the percy snapshot and test reports are not functional
```

tutorial/testing.py
byronz: I like the *graphical aspects of the app*, but  it seems long and with repeated graphs-ish in the following `such as`

tutorial/testing.py
byronz: yes, I need to open my grammarly extension next time :)

tutorial/testing.py
```diff
@@ -0,0 +1,295 @@
...
+        `threading.Thread`
+        2. A selenium webdriver is initialized and it navigates to the
+        local server with the `server_url`
+        3. It waits until the page has loaded in the browser
+
+    * #5 A test case is composed of preparation, actions and check points.
+    Both #5 and #6 are doing the same check in this example, we are expecting
+    that the defined `Div` component's text is identical to the `children`'s
+    content. #5 will wait the expected state to reach within an explicitly 4
+    seconds timeout. It's a safer way to use when you are doing an element
```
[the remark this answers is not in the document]
byronz: write actions

tutorial/testing.py
byronz: HOWEVER ?  🖌

tutorial/testing.py
byronz: why CSS selector needs a plural form here?

shammamah: I'm looking at it again and I think it might be better to say "the CSS selector" instead!

tutorial/chapter_index.py
```diff
@@ -995,6 +996,13 @@
         'description': 'Dash dev tools reference'
     },
 
+    'testing': {
+        'url': '/testing',
+        'content': testing.layout,
+        'name': 'Dash Testing',
+        'description': '(New in Dash 1.0!) An introduction on testing your dash app with selenium'
```
mkcor:
```suggestion
        'description': '(New in Dash 1.0!) An introduction to testing your dash app with selenium'
```

tutorial/dash_test_sample.py
```diff
@@ -0,0 +1,29 @@
+# 1. imports of your dash app
+import dash_html_components as html
+import dash
```
mkcor: Import dash first ;)

tutorial/dash_test_sample.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+# 2. give each testcase a tcid, and pass the fixture
+# as a function argument, less boilerplate
+def test_bsly001_falsy_child(dash_duo):
```
mkcor: For my own learning and understanding, why do we call it dash_*duo*?

tutorial/dash_test_sample.py
```diff
@@ -0,0 +1,29 @@
...
+
+    # 5. use wait_for_* if your target element is the result of a callback,
+    # keep in mind even the initial rendering can trigger callbacks
+    dash_duo.wait_for_text_to_equal("#nully-wrapper", "0", timeout=4)
+
+    # 6. use this form if its present is expected at the action point
+    assert dash_duo.find_element("#nully-wrapper").text == "0"
+
+    # 7. use python builtin assert statement, and describe your
+    # criteria as an assertion message after the comma.
```
mkcor: This is incorrect, it's a *failure* message you want to write.
References:
https://wiki.python.org/moin/UsingAssertionsEffectively
https://docs.python.org/3/reference/simple_stmts.html#grammar-token-assert-stmt

tutorial/dash_test_sample.py
```diff
@@ -0,0 +1,29 @@
...
+    assert dash_duo.get_logs() == [], "browser console contain no error"
```
mkcor:
```suggestion
    assert dash_duo.get_logs() == [], "browser console contains error(s)"
```

tutorial/testing.py
```diff
@@ -0,0 +1,297 @@
+from textwrap import dedent
+
+import dash_html_components as html
+import dash_core_components as dcc
```
mkcor: please import in alphabetical order ;)

tutorial/testing.py
```diff
@@ -0,0 +1,297 @@
...
+    But the fixture allows you to choose another browser from the command line,
+    e.g. `pytest --webdriver Firefox -k bsly001`.
+
+    **Notes**:
+
+    * The *Gecko(Marionette)* driver from Mozilla is not fully compatible with
+    selenium specifications. Some features may not work as expected.
+
+    * We only include *Chrome* and *Firefox* in the supported list for now,
+    but other popular webdrivers can be included if there are popular demands.
```
mkcor:
```suggestion
    but other popular webdrivers may be included based on popular demand.
```

tutorial/testing.py
```diff
@@ -0,0 +1,297 @@
...
+    Syntax(code),
+    dcc.Markdown(dedent("""
+
+    ### Notes
+
+    * #1 For most test scenarios, you don't need to import any modules for
+    the test; just import what you need for the Dash app itself.
+
+    * #2 A test case is a regular python function. The function name follows
+    this pattern: `test_{tcid}_{test title}`. The `tcid` (test case ID) is
```
mkcor:
```suggestion
    this pattern: `test_{tcid}_{function or feature}`. The `tcid` (test case ID) is
```

tutorial/testing.py
shammamah: I agree that it sounds a little odd. What about "related to the appearance of the app's layout and/or graphs"?

mkcor:
@shammamah wrote:

>     * please change all "circleci" to "CircleCI" 
>     * please change all "dash" to "Dash"

I would tend to agree, but (confessing my own inconsistency) I'm not sure because then why not capitalizing "selenium" and "python" everywhere as well?... Lower case is cute :p

shammamah:
@mkcor Good catch -- those should be capitalized too. I think it makes more sense to have either "Dash" or `dash` (but not "dash"). Super nitpicky, I know!

tutorial/dash_test_sample.py
byronz: it's a composite of selenium browser and a server fixture, so dash_combo, dash_comp, dash_duo were in the list

tutorial/dash_test_sample.py
byronz: good point, but the real reference should be here https://docs.pytest.org/en/latest/assert.html. And I will be more specific on the assert, it's like python native assert, but pytest overwrites all the asserts behavior in test collection files with some black magic.

mkcor: Oh, thanks for the actual reference! Good to know -- but it would have been really surprising (and wrong) to have the opposite convention for pytest's assert (pass vs. failure message).

tutorial/testing.py
```diff
@@ -0,0 +1,297 @@
...
+    criteria directly in the test case rather than wrapping the assert inside
+    another helper API. By looking at the test case, reviewers and maintainers
+    should be able to easily figure out the purpose of the test by
+    the test title, the app definition, the actions, and the checkpoints.
+
+    * #8 We use [Percy](https://percy.io/) as our *Visual Regression Testing*
+    tool. It's a good alternative to assertions when your checkpoint is
+    about the graphical aspects of a Dash app, such as the whole layout or a
+    `dcc.Graph` component. We integrate the Percy service
+    with a `PERCY_TOKEN` variable,  so the regression result is only
```
mkcor:
```suggestion
    with a `PERCY_TOKEN` variable, so the regression result is only
```

tutorial/testing.py
```diff
@@ -0,0 +1,297 @@
...
+    available in Plotly CircleCI setup.
```
mkcor:
```suggestion
    available in Plotly's CircleCI setup.
```

tutorial/testing.py
```diff
@@ -0,0 +1,297 @@
...
+    - dash_duo
+
+    The default fixture for Dash integration tests, it contains a
+    `thread_server` and a webdriver wrapped with high-level Dash testing APIs.
+
+    - dash_br
+
+    A standalone webdriver wrapped with high-level Dash testing APIs. This is
+    suitable for testing a Dash app in a deployed environment, i.e. one in
+    which your Dash app is accessible from a URL.
```
mkcor: http://localhost:8050/ is a URL, so an adjective is missing ("from a ... URL"); I don't think it's "public" because the app could be deployed on an intranet...

tutorial/testing.py
```diff
@@ -0,0 +1,297 @@
...
+
+    - dash_duo
+
+    The default fixture for Dash integration tests, it contains a
+    `thread_server` and a webdriver wrapped with high-level Dash testing APIs.
+
+    - dash_br
+
+    A standalone webdriver wrapped with high-level Dash testing APIs. This is
+    suitable for testing a Dash app in a deployed environment, i.e. one in
```
mkcor:
```suggestion
    suitable for testing a Dash app in a deployed environment, i.e., when
```
removing "which" afterwards

tutorial/testing.py
```diff
@@ -0,0 +1,297 @@
...
+
+    #### Element Locators
+
+    There are several strategies to
+    [locate elements](https://selenium-python.readthedocs.io/locating-elements.html#locating-elements);
+    CSS selector and XPATH are the two most versatile ways. We recommend using
+    the **CSS Selector** in most cases due to its
+    [performance and robustness](http://elementalselenium.com/tips/34-xpath-vs-css-revisited-2) across browsers.
+
+    If you are new at using CSS Selectors, these
```
mkcor:
```suggestion
    If you are new to using CSS Selectors, these
```

tutorial/dash_test_sample.py
alexcjohnson: For my part, I don't think of it as either pass or fail messages; the clearest is to write these messages using should/should not (as in the pytest reference), rather than is/is not. That way the message itself makes clear BOTH a criterion AND whether that criterion is a pass or a fail situation.

mkcor: Pardon my elliptic style (maybe confusing wording): By "failure message" I meant that the message gets printed upon failure of the assertion (by "you want to write" I didn't mean to express a preference or suggestion in terms of message style, I meant "you want to write the message that will show when the assertion fails").

I agree that using "should" makes the message more helpful, and thanks for pointing out it's standard practice (I hadn't paid attention to this so far). So, I'll delete my suggestion below; I guess "browser console should contain no error" would be the way to go then.

tutorial/dash_test_sample.py
mkcor: Right, you told me before about this composite. Wonderful, thanks!

tutorial/testing.py
byronz: not agree on `function or feature` here,  the rest of the function name should be a short summary of the test, the feature is the folder name that groups several files, like `devtools`

tutorial/dash_test_sample.py
byronz: it's always good to know if someone gets confused or have a different understanding. that means there is a way to improve on the doc side, the discussion is still beneficial.

tutorial/dash_test_sample.py
```diff
@@ -0,0 +1,30 @@
...
+
+    # 5. use wait_for_* if your target element is the result of a callback,
+    # keep in mind even the initial rendering can trigger callbacks
+    dash_duo.wait_for_text_to_equal("#nully-wrapper", "0", timeout=4)
+
+    # 6. use this form if its present is expected at the action point
+    assert dash_duo.find_element("#nully-wrapper").text == "0"
+
+    # 7. to make the checkpoint more readable, you can describe the
+    # acceptance criteria as an assert message after the comma.
```
mkcor:
```suggestion
    # acceptance criterion as an assert message after the comma.
```

tutorial/dash_test_sample.py
mkcor: @byronz absolutely! (suggested singular "criterion" meanwhile)

tutorial/testing.py
mkcor: definitely, it should be a short summary of the test, and I thought "function or feature" could express that (as in "test a given feature," e.g., `test_{tcid}_molecule2d_update()`), but it may be confusing and your structure is better (deleting suggestion)

mkcor: wait, maybe I'll just resolve because deleting could delete the thread as well (not sure)

PR #524, opened 2019-06-07
Add Dash Molecule2dViewer and update docs.

shammamah:
## About
* Added Dash Bio Molecule2dViewer to docs. 
* Reworked example code in docs to use the new `dash_bio_utils` package.
* Linted code.

Post-merge checklist:

- [x] I understand


tutorial/dashbio_component_examples.py
```diff
@@ -55,10 +64,17 @@
             'param_name': 'Show/hide Barplots',
             'description': 'Enable or disable the secondary bar plots for gaps and conservation.',
             'code': '''from six.moves.urllib import request as urlreq
+from six import PY3
 import dash_bio as dashbio
 
 
-data = urlreq.urlopen("https://raw.githubusercontent.com/plotly/dash-bio/master/tests/dashbio_demos/sample_data/alignment_viewer_p53.fasta").read().decode('utf-8')
+data = urlreq.urlopen(
+    "https://raw.githubusercontent.com/plotly/dash-bio-docs-files/master/" +
```
mkcor:
```suggestion
    'https://raw.githubusercontent.com/plotly/dash-bio-docs-files/master/' +
```

tutorial/dashbio_component_examples.py
```diff
@@ -55,10 +64,17 @@
...
+    "alignment_viewer_p53.fasta"
```
mkcor:
```suggestion
    'alignment_viewer_p53.fasta'
```

tutorial/dashbio_component_examples.py
```diff
@@ -71,10 +87,17 @@
             'param_name': 'Tile Size',
             'description': 'Change the height and/or width of the tiles.',
             'code': '''from six.moves.urllib import request as urlreq
+from six import PY3
 import dash_bio as dashbio
 
 
-data = urlreq.urlopen("https://raw.githubusercontent.com/plotly/dash-bio/master/tests/dashbio_demos/sample_data/alignment_viewer_p53.fasta").read().decode('utf-8')
+data = urlreq.urlopen(
+    "https://raw.githubusercontent.com/plotly/dash-bio-docs-files/master/" +
```
mkcor:
```suggestion
    'https://raw.githubusercontent.com/plotly/dash-bio-docs-files/master/' +
```

tutorial/dashbio_component_examples.py
```diff
@@ -71,10 +87,17 @@
...
+    "alignment_viewer_p53.fasta"
```
mkcor:
```suggestion
    'alignment_viewer_p53.fasta'
```

tutorial/dashbio_component_examples.py
```diff
@@ -410,6 +465,57 @@
...
+
+model_data = urlreq.urlopen(
+    'https://raw.githubusercontent.com/plotly/dash-bio-docs-files/master/' +
+    'mol2d_buckminsterfullerene.json'
+).read()
+
+dashbio.Molecule2dViewer(
+    id='molecule2d-selectedatomids',
+    modelData=json.loads(model_data),
+    selectedAtomIds=[i for i in range(10)]
```
mkcor:
```suggestion
    selectedAtomIds=list(range(10))
```
functional <3

tutorial/dashbio_component_examples.py
```diff
@@ -468,11 +588,18 @@
...
 
 
-data = urlreq.urlopen("https://raw.githubusercontent.com/plotly/dash-bio/master/tests/dashbio_demos/sample_data/needle_PIK3CA.json").read().decode("utf-8")
+data = urlreq.urlopen(
+    'https://raw.githubusercontent.com/plotly/dash-bio-docs-files/master/' +
+    'needle_PIK3CA.json'
+).read()
+
+if PY3:
+    data = data.decode("utf-8")
```
mkcor:
```suggestion
    data = data.decode('utf-8')
```

tutorial/dashbio_component_examples.py
```diff
@@ -493,11 +620,18 @@
...
```
mkcor:
```suggestion
    data = data.decode('utf-8')
```

tutorial/dashbio_component_examples.py
```diff
@@ -721,15 +930,18 @@
         },
         {
             'param_name': 'Point sizes and line widths',
-            'description': 'Change the sizeo f the points on the scatter plot, and the widths of the effect \
-            lines and the genome-wide line.',
+            'description': 'Change the sizeo f the points on the scatter plot, \
```
mkcor:
```suggestion
            'description': 'Change the size of the points on the scatter plot, \
```

tutorial/dashbio_component_examples.py
```diff
@@ -721,15 +930,18 @@
...
+            and the widths of the effect lines and the genome-wide line.',
```
mkcor:
```suggestion
            and the widths of the effect lines and genome-wide line.',
```
if I understand correctly, i.e., "change the width of the effect lines and the width of the genome-wide line" (as is, the meaning is "change ... and change the genome-wide line" which doesn't seem fine, because we don't know which feature of the genome-wide line is getting changed).

tutorial/examples/dashbio_components/molecule_2d_viewer.py
```diff
@@ -0,0 +1,45 @@
...
+])
+
+
+@app.callback(
+    dash.dependencies.Output('molecule2d-output', 'children'),
+    [dash.dependencies.Input('my-dashbio-molecule2d', 'selectedAtomIds')]
+)
+def update_selected_atoms(ids):
+    if ids is None or len(ids) == 0:
+        return "No atoms have been selected. Select an atom by clicking on it."
```
mkcor:
```suggestion
        return "No atom has been selected. Select atoms by clicking on them."
```

tutorial/examples/dashbio_components/molecule_3d_viewer.py
```diff
@@ -33,7 +45,8 @@
 )
 def show_selected_atoms(atom_ids):
     if atom_ids is None or len(atom_ids) == 0:
-        return 'No atoms have been selected. Click somewhere on the molecular structure to select an atom.'
+        return 'No atoms have been selected. Click somewhere on the molecular \
```
mkcor:
```suggestion
        return 'No atom has been selected. Click somewhere on the molecular \
```

tutorial/utils/dashbio_docs.py
```diff
@@ -138,7 +138,7 @@ def generate_component_example(
 
     # location of all sample data
     DATA_LOCATION_PREFIX = '''https://raw.githubusercontent.com/plotly/\
-dash-bio/master/tests/dashbio_demos/sample_data/'''
+dash-bio-docs-files/master/'''
```
mkcor: I would like to discuss this change. From a data provenance perspective, this looks like a bad change. This new *ad hoc* repo https://github.com/plotly/dash-bio-docs-files does not contain any metadata and it's clearly an example of repurposing public GitHub repos. I would argue that by fetching sample data from `dash-bio/master/tests/dashbio_demos/sample_data/`, we point to an existing source (instead of creating one *ex nihilo*) which contains (in the rest of the repo) at least *some* data lineage information.

tutorial/dashbio_component_examples.py
```diff
@@ -38,10 +40,17 @@
             'param_name': 'Color Scales',
             'description': 'Change the colors used for the heatmap.',
             'code': '''from six.moves.urllib import request as urlreq
+from six import PY3
 import dash_bio as dashbio
 
 
-data = urlreq.urlopen("https://raw.githubusercontent.com/plotly/dash-bio/master/tests/dashbio_demos/sample_data/alignment_viewer_p53.fasta").read().decode('utf-8')
+data = urlreq.urlopen(
+    "https://raw.githubusercontent.com/plotly/dash-bio-docs-files/master/" +
```
mkcor:
```suggestion
    'https://raw.githubusercontent.com/plotly/dash-bio-docs-files/master/' +
```
because we have single quotes from line 117 onwards

tutorial/dashbio_component_examples.py
```diff
@@ -38,10 +40,17 @@
...
+    "alignment_viewer_p53.fasta"
```
mkcor:
```suggestion
    'alignment_viewer_p53.fasta'
```

mkcor:
Thanks for this update! Sorry, I don't have a Percy account to validate the visual diffs.
I realize that when `dash_bio_utils` version `0.0.3-rc3` ships (https://github.com/plotly/dash-bio-utils/pull/3), I'll have to update not only `dash-bio` but also `dash-docs`.

tutorial/utils/dashbio_docs.py
shammamah: I don't necessarily disagree here, but in the interest of being consistent with the Plotly datasets (https://github.com/plotly/datasets), we can leave it in for now and change both in the future (if that is the decision that's made).

mkcor: Yes, that's an interesting topic to discuss and an important decision to make. I'm aware that many of these data files contain metadata, and why not having a clear policy on how to use this GitHub repo dedicated to data storage (e.g., append-only)...

For now, I'm wondering then, why couldn't these Dash Bio sample datasets go directly into https://github.com/plotly/datasets?

shammamah: We can definitely do that, too. I kept it separate since so many of the datasets have specialized file formats that are specific to this library only.

mkcor:
Can you poke someone to take care of the Percy diffs?

tutorial/utils/dashbio_docs.py
mkcor: I agree they are much more specialized. But if we could use the variables

Dataset name | Best for | Origin | Chart example | Download | Open in Plotly
-- | -- | -- | -- | -- | --

with new values for those domain-specific datasets (haven't investigated as of yet), I think they would fit fine in the list https://plotly.github.io/datasets/ :)

shammamah: The description for the repo is also 

"**CSV datasets** used in **Plotly API examples** https://plotly.github.io/datasets"

which doesn't really apply to our datasets. If we are adding the Dash Bio datasets into this repo, the description should also be updated accordingly. I'd personally prefer using the `dash-bio-docs-files` repo for that reason. I could also update the README for that repo to look more similar to the one for the datasets repo. What do you think?

mkcor: I can see that there are also a few JSON files in https://github.com/plotly/datasets, but fair enough.

I'm missing one element to get the full picture. Since you want to parallel the "Plotly datasets" (https://github.com/plotly/datasets, used in Plotly API examples), what is the main repo which makes use of these datasets? Basically, what's the equivalent of https://github.com/plotly/dash-docs, in a picture where https://github.com/plotly/datasets is the equivalent of https://github.com/plotly/dash-bio-docs-files?

> I'd personally prefer using the dash-bio-docs-files repo for that reason. I could also update the README for that repo to look more similar to the one for the datasets repo. What do you think?

I think the repo's name `dash-bio-docs-files` isn't very clear. If we intend to store data files, the name should contain at least "data" or "datasets." Do you envision storing other kinds of files?

nicolaskruchten: I just updated the "description" of that repo to remove the word "API"

nicolaskruchten: let's just use `datasets` for every dataset, there's no real need to scope things per library IMO

mkcor:
> Can you poke someone to take care of the Percy diffs?

I noticed that the visual diffs were approved, so that's great.

I haven't approved the PR yet, because I'm still hesitating about the dataset-only repo; I'll see what kind of feedback we can get from the team.

tutorial/utils/dashbio_docs.py
bpostlethwaite: If we want more metadata we could add a BIO folder into Datasets with its own README

mkcor: @nicolaskruchten thanks (why bother with yet another dataset repo, indeed)

> I just updated the "description" of that repo to remove the word "API"

Great! Do you mind removing "CSV" as well? ;)

shammamah:
@mkcor If you approve this PR, I'm happy to open up another one to take care of that particular problem. I just don't think it belongs in this particular discussion, since it applies to all of the components (not just this one).

tutorial/utils/dashbio_docs.py
mkcor: @bpostlethwaite very true

mkcor:
@shammamah the problem is that, as is, this PR introduces links to this new `dash-bio-docs-files` repo, which I don't think should exist. The logic would be to revert the specific changes pertaining to the location of sample datasets, approve, merge, and then address the issue of storing new data into `datasets` as a separate PR.

So, indeed, the point is not to require extra work in this PR, but a PR review should ensure that nothing undesirable lands into `master`.

mkcor:
> revert the specific changes pertaining to the location of sample datasets

`git revert e5b69a33746` should take care of it (alternatively, I'm fine with you rebasing)

requirements.txt
```diff
@@ -6,7 +6,8 @@ click==6.7
 configparser==3.5.0
 dash==0.43.0
 dash-auth==1.2.0
-dash-bio==0.0.10
+dash-bio==0.1.2rc3
+dash-bio-utils==0.0.3rc2
```
mkcor: It's not clear to me that `dash-bio-utils` should become an explicit dependency. Doesn't it get installed as a dependency of `dash-bio` when the latter gets installed? Plus, "overwriting" (over-determining, as done here) this nested dependency creates a second, parallel set of dependency versions.
I mean, `dash-bio` version `n` depends on `dash-bio-utils` version `m` and this is uniquely specified. It should remain so, while here we introduce the possibility of running `dash-bio` version `n` and `dash-bio-utils` version `m'`... Do you see the conflict?

tutorial/dashbio.py
```diff
@@ -97,9 +98,14 @@
             'width': '600'
         },
         'component_wrap': 'dcc.Graph(figure=_[0])',
-        'setup_code': '''df = pd.read_csv('https://raw.githubusercontent.com/plotly/dash-bio/master/tests/dashbio_demos/sample_data/clustergram_mtcars.tsv',
-        sep='\t', skiprows=4).set_index('model')
-data = df.values''',
+        'setup_code': '''df = pd.read_csv(
+    'https://raw.githubusercontent.com/plotly/dash-bio-docs-files' +
```
mkcor: Do not introduce this new, ad hoc repo.

requirements.txt
shammamah: The `dash-bio-utils` package isn't listed as a dependency of the `dash-bio` package: https://github.com/plotly/dash-bio/blob/22e6a157514b8749ea3a164c285a41083f6d5677/setup.py#L26

You might be confused because it is in the repository's `requirements.txt`. This does not refer to requirements for the actual package, but rather requirements for the demo apps in `tests/dashbio_demos`.

mkcor: Oh, right, I didn't double-check the actual package's dependencies. Thanks for clarifying. Well, I'll definitely take a mental note for it, because we probably want to use the same version pair `(n, m)` across demo apps, docs, gallery examples... Resolving this conversation insofar as this PR is concerned.

tutorial/dashbio.py
```diff
@@ -144,8 +169,16 @@
             ['json', 'json'],
             ['urllib.request', 'urlreq']
         ],
-        'setup_code': '''model_data = urlreq.urlopen('https://raw.githubusercontent.com/plotly/dash-bio-docs-files/master/mol3d/model_data.js').read()
-styles_data = urlreq.urlopen('https://raw.githubusercontent.com/plotly/dash-bio-docs-files/master/mol3d/styles_data.js').read()
+        'setup_code': '''
+model_data = urlreq.urlopen(
+    'https://raw.githubusercontent.com/plotly/dash-bio-docs-files/' +
```
mkcor: some left-over pointer ;)

tutorial/dashbio.py
```diff
@@ -144,8 +169,16 @@
...
+    'master/mol3d/model_data.js'
+).read()
+styles_data = urlreq.urlopen(
+    'https://raw.githubusercontent.com/plotly/dash-bio-docs-files/' +
```
mkcor: here as well

tutorial/dashbio.py
shammamah: Those URLs were already there -- the only thing I did was separate them into two lines, so we're not *introducing* any undesirable changes. I can put them back onto one line, to make it look as if the code were untouched.

mkcor: Oh, I see: Current `master` already contains instances of `dash-bio-docs-files`, which I didn't realize. It looks like they were introduced when adding Molecule3D examples, so it should have sparked a debate back then (it's not innocent to create, add, or mix up data sources).

PR #525, opened 2019-06-08
Extend multi page section with example of using flask blueprints

SaskiaKeil:
In this PR I add another way of having a multi-page setup which is - in my eyes - a bit cleaner than the current examples.
It uses a main flask app and then registers the dash apps via blueprints.

Happy to hear your feedback :)

----------------------

Post-merge checklist:

- [x] I understand

### DIFF
--- a/tutorial/urls.py
+++ b/tutorial/urls.py
@@ -540,4 +540,85 @@ however if this were imported from `index.py`, the initial loading of `index.py`
 would ultimately require itself to be already imported, which cannot be
 satisfied.
 '''),
+
+          dcc.Markdown('''***
+# Build a Multi-Page App with Flask
+
+Instead of using a conditional to serve different apps you can also use Flask's
+inbuilt routing via blueprints.
+To do so you have to combine different dash apps in one parent flask app.
+
+File structure:
+```
+- app.py
+- page_1.py
+- page_2.py
+```
+
+***
+
+`app.py`
+'''),
+
+          dcc.SyntaxHighlighter("""from flask import Flask, redirect, url_for
+
+# create application
+app = Flask(__name__)
+
+# Import dash apps after the definition of the flask app, so they can actually use it
+import page_1
+import page_2
+
+# Forward to Page 1 if no path is given
+@app.route('/')
+def index():
+    return redirect(url_for('/page-1/'))
+
+""", language='python', customStyle=styles.code_container),
+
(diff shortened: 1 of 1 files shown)


alexcjohnson:
@SaskiaKeil I agree, this is a really nice pattern. There are various pros and cons of this compared to the other methods we show, both in performance and in functionality, and we should describe the key distinctions. But this is certainly a technique people should be aware of!

SaskiaKeil:
@alexcjohnson Fair point. I added two lines with the pros and cons I could think about. Let me know if you have more in mind ;)

tutorial/urls.py
```diff
@@ -548,6 +548,13 @@ def display_page(pathname):
 inbuilt routing via blueprints.
 To do so you have to combine different dash apps in one parent flask app.
 
+The solution above using a callback is easy to setup and understand.
+Once you want to have a couple of pages it gets a bit messy in code though.
+Also it can have a small impact on performance as you have to evaluate the decision tree on every request.
+
+On the other hand, setting up another flask app to combine all dash apps adds a bit more complexity to your code,
```
alexcjohnson: I don't think the decision tree is a big deal - it's only for requests with a modified url, and you were to make the tree so big that the time became noticeable you could turn the long `if/elif/...` into a fast dict lookup.

Here are the factors that occur to me - @chriddyp any you'd like to add?

As mentioned at the top of this file:
> Dash renders web applications as a "single-page app". This means that the application does not completely reload when the user navigates the application, making browsing very fast.

Making each page a separate `dash.Dash` app reverses that. Changing pages this way results in a full reload, rather than just an ajax call and a rerender. That'll probably be slower, but in exchange each page may be simpler, so a faster initial load if for example you have one page that needs a large component that none of the other pages needs.

This way you also get a simpler callback graph - a separate graph for each page - which can simplify development because you don't need to worry about ID collisions (see https://github.com/plotly/dash/issues/637) and may allow you to avoid `suppress_callback_exceptions = True`. And a very large callback graph that you're only using a small part of at a time may have loading and dynamic performance costs. But it also means sharing state between pages becomes more cumbersome - the single-app way you can make a `dcc.Store`, or leave some dropdowns or other input elements at the top or side of the page, and have all the sub-pages share the info there. You can do this with a `local` or `session` `dcc.Store` if either of those degrees of permanence fits your use case.

PR #529, opened 2019-06-10
Remove .RData and .Rhistory and add both to .gitignore

rpkyle:
The docs repo should not contain `.RData` or `.Rhistory` files; these have been purged, and the filenames added to `.gitignore`.


nicolaskruchten:
💃

PR #530, opened 2019-06-11
Make Store examples' syntax match their code

alexcjohnson:
... by extending `load_example` to ignore lines with "# no-run" comments. Stores are "special" components that can't (at least for now) be added outside the original layout, so we create them in the base `run.py`, and we have to strip them out of the examples before running them within the greater docs app. Would be nice to remove that restriction, but until then this at least keeps the two pieces in sync.

This is an updated version of #496 by @StefanTippelt 

Post-merge checklist:

- [X] I understand

### DIFF
--- a/tutorial/examples/core_components/store_clicks.py
+++ b/tutorial/examples/core_components/store_clicks.py
@@ -1,7 +1,7 @@
 import dash
 
 import dash_html_components as html
-import dash_core_components as dcc  # no-run
+import dash_core_components as dcc  # no-exec
 from dash.dependencies import Output, Input, State
 from dash.exceptions import PreventUpdate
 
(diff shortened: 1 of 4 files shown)


tutorial/examples/core_components/store_clicks.py
```diff
@@ -11,8 +11,15 @@
 app = dash.Dash(__name__, external_stylesheets=external_stylesheets)
 
 app.layout = html.Div([
-
-
+    # The memory store will always get the default on page refreshes
```
byronz: page refresh?

tutorial/tools.py
```diff
@@ -1,15 +1,14 @@
...
     def wrapper(path):
         try:
             return func(path)
         except Exception as e:
-            print('\nError running {}\n{}'.format(
-                path,
-                ('======================================' +
-                 '======================================')
-            ))
+            print('\nError running {}\n{}'.format(path, '=' * 76))
```
byronz: 👍

tutorial/tools.py
```diff
@@ -51,12 +50,26 @@ def load_example(path):
             'print("Running")\n    # app.run_server'
         )
 
+        # if there are lines that should be included in the syntax but
+        # not executed, simply put this comment at the end of the line.
+        no_run = '# no-run'
```
byronz: Shall we use not-exec or no-py-exec? I didn't get it in the comments until here

alexcjohnson: sure, I can change it to `no-exec` if you think that's clearer.

alexcjohnson:
Buttons moved into the table:
Screen Shot 2019-06-11 at 9 53 16 PM

PR #534, opened 2019-06-13
DCC landing page, dropdown, slider, rangeslider,input, textarea. CSS …

KPhans:
…and proptable modifications will come after.

Post-merge checklist:

- [ ] I understand


dashr/chapters/dash-core-components/Input/examples/blur.R
```diff
@@ -0,0 +1,28 @@
...
+app$callback(
+  output(id = 'output-keypress', property='children'),
+  params=list(input(id='input-1-submit', property='n_submit'),
+              input(id = 'input-1-submit', property = 'n_blur'),
+              input(id='input-2-submit', property='n_submit'),
+              input(id = 'input-2-submit', property = 'n_blur'),
+              state('input-1-submit', 'value'),
+              state('input-2-submit', 'value')),
+  function(ns1, nb1, ns2, nb2, input1, input2) {
+    sprintf('Input 1 is %s and input 2 is %s', input1,input2)
```
rpkyle: insert space between `input1,` and `input2`

dashr/chapters/dash-core-components/Checklist/examples/checklistproptable.R
```diff
@@ -0,0 +1,31 @@
+
+
+test =utils$props_to_list('dccChecklist')
```
rpkyle: :hocho: test?

dashr/chapters/dash-core-components/Checklist/examples/checklistproptable.R
```diff
@@ -0,0 +1,31 @@
...
+df <- data.frame(matrix(unlist(test), nrow=length(test), byrow=T))
+names(df) = c('Attribute', 'Description')
```
rpkyle: I apologize, it's a bit confusing with different terms across languages, but I think we actually want to use _Argument_ instead of _Attribute_ here, because the former is what's in the R online help.

dashr/chapters/dash-core-components/Checklist/examples/checklistproptable.R
```diff
@@ -0,0 +1,31 @@
...
+        align = c('left', rep('center', ncol(df))),
+        line = list(width = 1, color = 'white'),
+        fill = list(color = 'white'),
+        font = list(family = "Arial", size = 14, color = "#264e86")
+      ),
+      cells = list(
+        values = rbind(
+          rownames(df), 
+          t(as.matrix(unname(df)))
+        )[c(2,3),],
```
rpkyle: It looks like the content of `cells` is essentially restructuring the output of `propsAsList` into a single row, in which the column names are the `arguments` themselves, and the vector elements are the string contents. Is that right?

I'll post a follow-up comment with my take on this if so, my concern here is the `[c(2,3),]` bit; I think matching on character arguments is safer in general, since the result is invariant to position. 

Alternatively, I have also used `match` to identify the column or row position of a character name, which I then use to subset.

dashr/chapters/dash-core-components/Input/examples/inputproptable.R
```diff
@@ -0,0 +1,30 @@
+
+test =utils$props_to_list('dccInput')
```
rpkyle: same comment as above, :hocho: test and rename to something like `propsList`?

dashr/chapters/dash-core-components/Input/index.R
```diff
@@ -0,0 +1,37 @@
...
+  
+  htmlH3('Update Value on Keypress'),
+  
+  examples$keypress$source,
+  examples$keypress$layout,
+  htmlH3('Update Value on Enter/Blur'),
+  
+  dccMarkdown("
+`dccInput` has properties `n_submit`, which updates when the enter button is pressed, and `n_blur`
+ ,which updates when the component loses focus (e.g tab is pressed or the user clicks away).
```
rpkyle: insert space within `,which`

rpkyle: period missing in `e.g`

rpkyle: instead of "the user clicks away" I might write "the user clicks outside of the input field"

dashr/chapters/dash-core-components/Checklist/examples/checklistproptable.R
KPhans: a name i assigned to the list. it is the same for all. dont think it matters

PR #536, opened 2019-06-14
add documentation for the CHECKS file in DDS

michaelbabyn:
resolves /plotly/streambed#11907

Post-merge checklist:

- [x] I understand


tutorial/dash_deployment_server_examples.py
```diff
@@ -641,7 +642,13 @@ def display_instructions_deploy(method):
     ```server = app.server```
 
     ***
+    `CHECKS`
 
+    This optional file allows you to define custom checks to be performed on your app upon deployment.
```
cldougl: I think we should note the DDS version here so something like:
As of DDS version 3.2.0, this optional file allows you to define custom checks to be performed on your app upon deployment.

michaelbabyn: Aren't the checks part of all versions DDS but mostly we've just been using the default 10 second test for failure? I've been seeing this message on deployment as far back as I can remember.

```
-----> Attempting pre-flight checks
       For more efficient zero downtime deployments, create a file CHECKS.
       See http://dokku.viewdocs.io/dokku/deployment/zero-downtime-deploys/ for examples
       CHECKS file not found in container: Running simple container check...
-----> Waiting for 10 seconds ...
-----> Default container check successful!

```

cldougl: @michaelbabyn you're right! The additional changes have been postponed (https://github.com/plotly/streambed/issues/11907#issuecomment-484910822 ) so this will work as is.
💃

PR #540, opened 2019-06-19
Remove references to SyntaxHighlighter and highlight with dcc.Markdown

wbrgss:
Removes all references to `SyntaxHighlighter`.

Hopefully Percy will catch if I missed manually wrapping any code blocks with ` ``` ``` `.

The new highlighter will color Python bools a different way, and will also highlight inline code, e.g. the JS in `tutorial/react_for_python_developers.py`, which was previously un-highlighted.

In many cases, formatting/indentation/references to dedent could be improved to make the docs look prettier, but it should be functional.

Post-merge checklist:

- [x] I understand


tutorial/tools.py
```diff
@@ -68,7 +68,7 @@ def load_example(path):
         exec(_example, scope)
 
     return (
-        _source,
+        '```python ' + _source + '```',
```
alexcjohnson: Do we want a newline ````'```python\n' + _source + '```'````?

tutorial/utils/component_block.py
```diff
@@ -26,10 +26,9 @@ def ComponentBlock(example_string, **kwargs):
 
         raise e
     return html.Div([
-        dcc.SyntaxHighlighter(
-            example_string,
-            language='python',
-            customStyle=styles.code_container
+        dcc.Markdown(
+            '``` python ' + example_string + '  \n```',
```
alexcjohnson: newline here too?

tutorial/utils/dashbio_docs.py
```diff
@@ -483,7 +482,7 @@ def generate_prop_table(
             html.Tr([html.Td(dcc.Markdown(prop_name)),
                      html.Td(dcc.Markdown(prop_desc)),
                      html.Td(dcc.Markdown(prop_type)),
-                     html.Td(dcc.SyntaxHighlighter(prop_default))])
+                     html.Td(dcc.Markdown(prop_default))])
```
alexcjohnson: backticks?

contributing-dds.md
```diff
@@ -51,7 +51,7 @@ NewChapter = html.Div(children=[
 ##### Best Practices
 
 When adding **text** use: ```dcc.Markdown(s())```    
-When adding **code** use: ```dcc.SyntaxHighlighter(s())```     
+When adding **code** use: ```dcc.Markdown(s())```
```
alexcjohnson:
````suggestion
When adding **code** use: ```dcc.Markdown(s())``` with triple backticks.
````

File: tutorial/auth.py [hunk 5]
```diff
@@ -63,67 +63,76 @@
...
+                    'l': 20,
+                    'b': 20,
+                    'r': 10,
+                    't': 60
+                }
+            },
+            'data': [{'x': [1, 2, 3], 'y': [4, 1, 2]}]
+        }
+
+    app.scripts.config.serve_locally = True
```
alexcjohnson: This is the default now - so any `serve_locally = True` lines can be removed.

tutorial/tools.py
wbrgss: I found this wasn't necessary, but I did it in [other places](https://github.com/plotly/dash-docs/pull/540/files#diff-88e1faf7d97d8d38640d1fcd2b683a89R14), so I'll change this for consistency.

Markdown seems to like `  \n` (i.e. two spaces before the newline), so I'm inclined to do `'```python  \n' + _source + '  \n```'` but for code blocks I'm not sure it matters in the end.

alexcjohnson: huh, I hadn't encountered a need for spaces before newline - and I would hope we don't need this, as that would be the only place we need to allow trailing whitespace, which most of us have our editors set to automatically trim. Anyway not a big deal, for now we just need to get it to work. If it works consistently the way you have it then fine.

wbrgss: Yeah, on second thought, I think this is actually for regular Markdown syntax (at least in the [original spec](https://daringfireball.net/projects/markdown/syntax#p)), not Markdown code syntax

> When you do want to insert a <br /> break tag using Markdown, you end a line with two or more spaces, then type return.

alexcjohnson: oh right, I do remember seeing that. Super annoying that the spec uses trailing spaces that way. I can think of one kind of hacky way to write that in an editor that removes trailing spaces:
```
'''
my line ending in spaces.  ''' '''
and another line
'''
```

tutorial/core_component_examples.py
```diff
@@ -514,23 +514,40 @@
     html.Hr(),
     html.H3("Inline Code"),
     html.P("Any block of text surrounded by ` ` will rendered as inline-code. "),
+    # Don't use ComponentBlock for markdown block quotes... too complicated to
+    # get all the nested quotes right!
+    dcc.Markdown("""
+    ````py
```
byronz: needs 4 here ?

alexcjohnson: yes, because this block is documenting how to use 3 backticks, so needs to be wrapped in more!

This has triple backticks nested in triple single-quotes nested in quadruple backticks nested in triple double-quotes... but who's counting 😏

byronz: 👽

Issue #551, opened 2019-06-26
Not all callbacks functional in Part 5 of Dash Cytoscape docs


ohaanika:
# Problem

Not all callbacks appear to be functional in [**Part 5. Event Callbacks for User Interactions**](https://dash.plot.ly/cytoscape/events), specifically in the first two sections, **Simple callback construction** and **Click, tap and hover**.

# Potential Cause

Since all three `cytoscape` graphs in Part 5 have the same `id='cytoscape-event-callbacks'`, the callbacks divided between the three sections are retrieving input data from the most recently declared graph (i.e. the third graph) instead of the intended graphs. 

# Potential Solution

Relabel each `cytoscape` graph and respective callbacks to have `id='cytoscape-event-callbacks-i'` for `i={1,2,3}`.

ohaanika:
Solution implemented, tested and pushed to branch: https://github.com/plotly/dash-docs/tree/dash-cytoscape/tutorial

rpkyle:
#552 merged (thanks @ohaanika!) so closing this stale issue.

Issue #550, opened 2019-06-26
Syntax styling issue for Dash Cytoscape docs 


ohaanika:
# Summary

When converting from `dcc.SyntaxHighlighter` to `dcc.Markdown` for Dash Cytoscape docs, the changes don't seem to be reflected in the docs hosted online but appear as they should when run locally as shown below for an example from [**Part 1: Elements**](https://dash.plot.ly/cytoscape/elements).

Hosted online:
<img width="956" src="https://user-images.githubusercontent.com/23238764/60152579-032f1900-97af-11e9-9870-8fde9b0dfe16.png">

Local:
<img width="956" src="https://user-images.githubusercontent.com/23238764/60152575-00ccbf00-97af-11e9-8178-498f6951fadd.png">

# Potential Cause/Solution

This appears to be the case for all Parts 1-6 where `Display(...)` is used as shown in the [code for the above example](https://github.com/plotly/dash-docs/blob/master/tutorial/cytoscape/elements_chapter.py):

```
Display = CreateDisplay({
    'cyto': cyto,
    'my_stylesheet': my_stylesheet
})

...

Display('''
    cyto.Cytoscape(
        id='cytoscape-elements-boolean',
        layout={'name': 'preset'},
        style={'width': '100%', 'height': '400px'},
        elements=[...]
    )
    ''')
```

This may be better addressed by referring the code for `CreateDisplay` in [utils.py](https://github.com/plotly/dash-docs/blob/master/tutorial/cytoscape/utils.py).

ohaanika:
Turns out the error occurs locally as well! I might have had a different version of Dash installed before?

Nonetheless, solution implemented, tested and pushed to branch `dash-cytoscape`: https://github.com/plotly/dash-docs/tree/dash-cytoscape/tutorial

rpkyle:
Branch has already been merged (thanks @ohaanika!) so closing this stale issue 🙂

PR #552, opened 2019-06-26
Dash Cytoscape updates

ohaanika:
Relevant Issues: #551, #550 

Post-merge checklist:

- [x] I understand


alexcjohnson:
Nice! Great job debugging the callback problems.

I wonder if we can take it a step further though: Now that `dcc.Markdown` automatically handles syntax highlighting and dedenting, can we get rid of `PythonSnippet` entirely? Replace:
```py
    PythonSnippet('''
    @app.callback(Output('cytoscape-callbacks-1', 'layout'),
                  [Input('dropdown-callbacks-1', 'value')])
        def update_layout(layout):
            return {'name': layout}
    '''),
```

With:
````py
    dcc.Markdown('''
    ```py
    @app.callback(Output('cytoscape-callbacks-1', 'layout'),
                  [Input('dropdown-callbacks-1', 'value')])
        def update_layout(layout):
            return {'name': layout}
    ```
    ''')
````

Which to my mind anyway is much clearer because you don't have to wonder what's going on in `PythonSnippet`.

And then it turns out a lot of the previous `PythonSnippet` components are preceded or followed by other `dcc.Markdown` components so they can be merged (after removing the remaining `dedent` calls) - see https://github.com/plotly/dash-docs/pull/549 for an example.

**Comment by ohaanika** [comment]
@alexcjohnson Apologies for taking long to get back to this PR! 

Since I see that `dcc.Markdown()` is used every where else in the docs, I did as you suggested to keep things consistent. To be specific:

1. I had to rewrite `PythonSnippet('''...''')` as `dcc.Markdown(dedent(''' ```py ... ``` '''), style=styles.code_container)` to maintain the styling that have we for code snippets (i.e. grey left border). 
2. I did not merge `dcc.Markdown` components, as the `dcc.Markdown` component capturing code snippets had different styling which had to specified (as mentioned above). 

Come to think of it... I just remembered with Cytoscape R docs, I remember using the equivalent of `Display` instead of the equivalent of `PythonSnippet` for such code snippets, since it happened by chance that the code snippets that we want to show wouldn't output anything to begin with. Just a side note!

File: tutorial/cytoscape/applications_chapter.py [hunk 1]
```diff
@@ -37,10 +36,12 @@
     
     ''')),
 
-    PythonSnippet('''
+    dcc.Markdown(dedent('''
```
alexcjohnson: `dedent` is no longer necessary as of Dash 1.0 https://github.com/plotly/dash-core-components/pull/569

ohaanika: Oh is this change yet to be implemented in the rest of the docs too?
e.g. https://github.com/plotly/dash-docs/blob/master/tutorial/table/sizing_chapter.py

alexcjohnson: right, we haven't stripped it out everywhere yet, that's going to take a while... but we can at least avoid adding new uses.

ohaanika: In that case I'll remove ALL `dedent()` for Cytoscape docs (i.e. including non-code snippets).

ohaanika: Removed!

alexcjohnson: Great! now if `dedent` isn't used anymore in these files, can you :hocho: `from textwrap import dedent`?

ohaanika: Good point, just committed 👍

PR #556, opened 2019-06-27
remove "Alpha" description from table docs

alexcjohnson:
@chriddyp are you still happy to have your monicker on this revised note?

Post-merge checklist:

- [x] I understand


alexcjohnson:
While I was here, I added a note about the demise of table's `content_style` prop to the migration guide, which we missed previously.

alexcjohnson:
using this PR to make the percy tests more robust so we don't need to approve so many partially-rendered pages... turns out by giving the pages a little longer to load, and because we run CI in dev mode, this turned up an error in the `DatePickerSingle` main example
Screen Shot 2019-06-27 at 11 25 42 AM
fixed in https://github.com/plotly/dash-docs/pull/556/commits/07ff36e7006d8c117ac4bc72106022fa202120ae

PR #579, opened 2019-07-03
add return before example code in dcc markdown

### LINKED ISSUE #575: library(dash) not showing up in examples using Dash
For example, in all of the examples from https://dashr.plot.ly/getting-started only the below libraries are called which leads to an `error: object 'Dash' not found`.

```
library(dashCoreComponents)
library(dashHtmlComponents)
```

michaelbabyn:
resolves #575 by adding a return in example code gen function so that the first line doesn't always get ignored by dcc markdown. cc @cldougl, @rpkyle 

Post-merge checklist:

- [x] I understand


cldougl:
awesome @michaelbabyn !!! 
looks good to me! 

💃 

Screen Shot 2019-07-03 at 6 25 48 PM

PR #586, opened 2019-07-05
Fix DashR doc urls, etc.

ohaanika:
Touching up on the following:
- [x] Handle parsing issue for Dash DataTable example (part 5, example 2)
- [x] Change R urls to match Python urls
- [x] Implement reusable components in Dash DataTable docs to fix styling 
- [x] Add reference page for Dash DataTable
- [x] Add grey border to `code-container` and space under `example-container`
- [x] Change function generating reference table from `generate_props_table` to `generate_table`


rpkyle:
🎉 🚀 💃

PR #610, opened 2019-07-23
:pencil: add section for dashR testing

### LINKED ISSUE #609: add documents for dashR testing
complement to https://github.com/plotly/dash/pull/827

byronz:
Closes #609 

Post-merge checklist:

- [x] I understand


tutorial/testing.py
```diff
@@ -145,8 +147,8 @@ def test_bsly001_falsy_child(dash_duo):
 
     ## Fixtures
 
-    To avoid accidental name collision with other pytest plugins,
-    all Dash test fixtures start with the prefix `dash_`.
```
alexcjohnson: Still seems like one sentence - and perhaps we could just drop the `_`:

> To avoid accidental name collision with other pytest plugins,
> all Dash test fixtures start with the prefix `dash`.

Also, I might reorganize a little: `dash_br` isn't just a Dash for Python fixture any more, so we could say something like:

- `dash_br`: A standalone WebDriver wrapped with high-level Dash testing APIs. Used by `dash_duo` for testing Python Dash apps, and by `dashr` for testing R Dash apps. You can also use this for testing a Dash app in a deployed environment, i.e. when your Dash app is accessible from a URL.

### Dash for Python fixtures:
- `dash_duo` ...
- `dash_thread_server` ...
- `dash_process_server` ...

### Dash for R fixtures
- `dashr` ...
- `dashr_server` ...

tutorial/testing.py
```diff
@@ -260,6 +277,41 @@ def test_bsly001_falsy_child(dash_duo):
...
 
+    ## Dash for R testing
+
+    We released [Dash for R](https://medium.com/@plotlygraphs/announcing-dash-for-r-82dce99bae13)
+    in July 2019. As a follow-up to this initiative, we extend `dash.testing`
+    so it can support integration with Dash for R Apps. The core change is
+    achieved in less than 50 lines of [python code](https://github.com/plotly/dash/blob/339141a1196c4a8af96fd7147b23e07c39866003/dash/testing/application_runners.py#L228-L277).
+    With the new `dashr` fixture, we can have a harmonized test development
+    experience in Dash for R. Here is a simple example runnable with `pytest`, with
+    a test target app written in R language.
```
alexcjohnson: 50 LOC is certainly impressive, and speaks volumes about how clean the testing framework is; but I don't think that really belongs in the tutorial. Perhaps a blog post titled something like "Testing R apps from Python?" extolling how easy it is to support new languages this way... but here I'd just highlight that we're writing the test code in Python with the R code as a string or a separate file.

How about:

> We released [Dash for R](https://medium.com/@plotlygraphs/announcing-dash-for-r-82dce99bae13) in July 2019. To facilitate testing it, we extended the *Python* package `dash.testing` to support Dash for R apps. The new `dashr` fixture gives us a test development experience in Dash for R that's nearly identical to the `dash_duo` fixture in Dash for Python. Here is a simple example runnable with `pytest`. The R app is written as a string inside the Python code (`dashr` can also accept a path to a `.R` file), and we use the same Selenium interface to control and query it as we do Dash for Python apps.

tutorial/testing.py
```diff
@@ -274,20 +326,18 @@ def test_bsly001_falsy_child(dash_duo):
...
     The [CircleCI Local CLI](https://circleci.com/docs/2.0/local-cli/) is a
-    handy tool to run all the jobs locally. It gives you an earlier warning
-    before even pushing your commits to remote, which leaves no chance of
-    making an embarrassing public mistake. The environment is identical to the
-    remote one, except the Percy snapshot and test reports are not functional
-    locally.
+    handy tool to execute some jobs locally. It gives you an earlier warning
+    before even pushing your commits to remote. For example, it's always
+    recommended to pass lint and unit tests job first on your local machine. So
+    we can make sure there is no simple mistakes in the commit.
```
```suggestion
    we can make sure there are no simple mistakes in the commit.
```

tutorial/testing.py
```diff
@@ -8,6 +8,8 @@
 
     *New in Dash v1.0*
 
+    *support Dash for R testing in v1.1.0*
```
```suggestion
    *support Dash for R testing added in v1.1.0*
```

tutorial/testing.py
```diff
@@ -145,19 +147,19 @@ def test_bsly001_falsy_child(dash_duo):
...
-    `thread_server` and a WebDriver wrapped with high-level Dash testing APIs.
+    To avoid accidental name collision with other pytest plugins, all Dash test
+    fixtures start with the prefix `dash` or `dashr`.
 
     - dash_br
 
     A standalone WebDriver wrapped with high-level Dash testing APIs. This is
-    suitable for testing a Dash App in a deployed environment, i.e. when
-    your Dash App is accessible from a URL.
+    suitable for testing a Dash App in a deployed environment (Dash for Python
```
byronz: @alexcjohnson  the dash_br was used by neither dash_duo nor dashr.  it's a direct instantiation from the browser class. but good point about the categories.

tutorial/testing.py
```diff
@@ -260,6 +277,43 @@ def test_bsly001_falsy_child(dash_duo):
...
+    ## Dash for R testing
+
+    We released [Dash for R](https://medium.com/@plotlygraphs/announcing-dash-for-r-82dce99bae13)
+    in July 2019.  To facilitate testing it, we extended the *Python* package
+    `dash.testing` to support Dash for R Apps. The new `dashr` fixture gives
+    us a test development experience in Dash for R that's nearly identical to
+    the `dash_duo` fixture in Dash for Python. Here is a simple example
+    runnable with `pytest`. The R App is written as a string inside the Python
+    code (the `app` argument can also be a valid path to a `.R` file), and we
+    can use the same set of APIs calls to test it exactly as we do Dash testing
```
```suggestion
    can use the same set of API calls to test it exactly as we do Dash testing
```

tutorial/testing.py
```diff
@@ -171,6 +173,21 @@ def test_bsly001_falsy_child(dash_duo):
     configure your `PYTHONPATH` so that the Dash app source file is
     directly importable*.
 
+    And `Dash for R` test fixtures have a prefix `dashr_`.
```
alexcjohnson: `dashr` does not have a prefix `dashr_`... anyway you mention the `dashr` prefix above now.

tutorial/testing.py
```diff
@@ -171,6 +173,21 @@ def test_bsly001_falsy_child(dash_duo):
...
+
+    - dashr_server
+
+    Start your Dash for R App with `Rscript` in a python `subprocess`. Note that
+    unlike python server fixtures, the start server arguments can only be
+    configured inside the Dash for R App.
+
+    - dashr
```
alexcjohnson: nonblocking, but if it were me I'd move `dashr` before `dashr_server`, similar to how we put `dash_duo` before the two python server fixtures since it's the one we want people to learn first.

tutorial/testing.py
```diff
@@ -145,19 +147,19 @@ def test_bsly001_falsy_child(dash_duo):
...
 
-    To avoid accidental name collision with other pytest plugins,
-    all Dash test fixtures start with the prefix `dash_`.
-
-    - dash_duo
-
-    The default fixture for Dash integration tests, it contains a
-    `thread_server` and a WebDriver wrapped with high-level Dash testing APIs.
+    To avoid accidental name collision with other pytest plugins, all Dash test
+    fixtures start with the prefix `dash` or `dashr`.
```
rpkyle: The vast majority of R users won't know anything about `pytest` or test fixtures, so I might suggest adding a single line to explain what this jargon means, e.g. _in this context, a test fixture is a minimal implementation of a Dash for R app which provides a reliable baseline to configure, initialize and execute the app and test one or more of its features. For more details, please visit https://docs.pytest.org/en/latest/fixture.html_

tutorial/testing.py
```diff
@@ -260,6 +277,43 @@ def test_bsly001_falsy_child(dash_duo):
...
     | `server_url` | set the server_url as setter so the selenium is aware of the local server port, it also implicitly calls `wait_for_page`. return the server_url as property |
 
+    ## Dash for R testing
+
+    We released [Dash for R](https://medium.com/@plotlygraphs/announcing-dash-for-r-82dce99bae13)
+    in July 2019.  To facilitate testing it, we extended the *Python* package
+    `dash.testing` to support Dash for R Apps. The new `dashr` fixture gives
+    us a test development experience in Dash for R that's nearly identical to
+    the `dash_duo` fixture in Dash for Python. Here is a simple example
+    runnable with `pytest`. The R App is written as a string inside the Python
```
rpkyle: Could change "The R app is written..." to "Your Dash app is written ..." or "The Dash for R app is written ..."

tutorial/testing.py
```diff
@@ -260,6 +277,43 @@ def test_bsly001_falsy_child(dash_duo):
     | `session_id` | shortcut to `driver.session_id` |
     | `server_url` | set the server_url as setter so the selenium is aware of the local server port, it also implicitly calls `wait_for_page`. return the server_url as property |
 
+    ## Dash for R testing
+
+    We released [Dash for R](https://medium.com/@plotlygraphs/announcing-dash-for-r-82dce99bae13)
+    in July 2019.  To facilitate testing it, we extended the *Python* package
+    `dash.testing` to support Dash for R Apps. The new `dashr` fixture gives
```
rpkyle: for consistency, would lowercase "Apps" to "apps"

tutorial/testing.py
```diff
@@ -171,6 +173,21 @@ def test_bsly001_falsy_child(dash_duo):
...
+    - dashr_server
+
+    Start your Dash for R App with `Rscript` in a python `subprocess`. Note that
+    unlike python server fixtures, the start server arguments can only be
+    configured inside the Dash for R App.
+
+    - dashr
+
+    The default fixture for Dash for R integration tests. As `dash_duo` in Python,
+    it contains a `dashr_server` and a selenium webdriver with the same Dash
```
rpkyle: I'm not certain, but I believe this is usually capitalized as "Selenium WebDriver".

tutorial/testing.py
```diff
@@ -260,6 +277,43 @@ def test_bsly001_falsy_child(dash_duo):
...
 
+    ## Dash for R testing
+
+    We released [Dash for R](https://medium.com/@plotlygraphs/announcing-dash-for-r-82dce99bae13)
+    in July 2019.  To facilitate testing it, we extended the *Python* package
+    `dash.testing` to support Dash for R Apps. The new `dashr` fixture gives
+    us a test development experience in Dash for R that's nearly identical to
+    the `dash_duo` fixture in Dash for Python. Here is a simple example
+    runnable with `pytest`. The R App is written as a string inside the Python
+    code (the `app` argument can also be a valid path to a `.R` file), and we
```
rpkyle: We should clarify here that the "`app` _argument can also be a call to_ `Rscript -e` _followed by a valid path to an_ `.R` _script_." I believe just passing the path to an R script will fail.

alexcjohnson: Looking at the code:

[`if not (os.path.isfile(app) and os.path.exists(app))`](https://github.com/plotly/dash/blob/fe34ed7d3d904e7c8eb737899a0d586066c13a2c/dash/testing/application_runners.py#L240)
...
[`app = path`](https://github.com/plotly/dash/blob/fe34ed7d3d904e7c8eb737899a0d586066c13a2c/dash/testing/application_runners.py#L256)

seem pretty clear that it's either an app string or a path. `Rscript` then gets called with one of those. I don't see a `-e` anywhere... but yes, @rpkyle if you think it would be useful to users to describe exactly how this gets invoked we can certainly do that.

byronz: @rpkyle the command call is wrapped in applicaiton_runner.py,  and I did tried with Rscript *.R, it was working

PR #614, opened 2019-08-02
Dash Bio dash-1.0.0 compatibility 

shammamah:
Quick fix of https://github.com/plotly/dash-bio/issues/399.

Post-merge checklist:

- [x] I understand


requirements.txt
```diff
@@ -5,7 +5,7 @@ click==6.7
 configparser==3.5.0
 dash==1.0.2
 dash-auth==1.2.0
-dash-bio==0.1.3rc4
+dash-bio==0.1.4rc1
```
chriddyp: It feels odd to still be pushing out rc versions to the community - generally `rc` versions are for internal use, regular versions for stuff we feel like is ready for community use.

shammamah: I see what you mean. This version is identical to the previous working version (0.1.3), except it just has some upgrades to React 16. Do you think it's worth it to just release this version? The reason we've been holding off is because we don't have tests that work with Dash 1.0 yet, but that shouldn't be a blocking issue here.

chriddyp: Yeah - Whatever we officially recommend should be an official version. And it's totally OK to have many small patch releases with little fixes - new versions don't need to contain big features or bug fixes!

Then, reserve rc releases for prototypes or bug fixes that we want community feedback on before we officially recommend it.

mkcor: Sure, going from 0.1.3 to 0.1.4 because we've added "some upgrades to React 16" would make sense. I agree with the versioning convention described by @chriddyp :+1:

mkcor: I mean "would make sense" *with or without tests*

mkcor:
@shammamah I'm confused: I want to approve but I can't figure out what dash-bio 0.1.4 contains, since it's not in `master`... Does https://pypi.org/project/dash-bio/0.1.4/ correspond to https://github.com/plotly/dash-bio/pull/412/files? Why was it closed then?

requirements.txt
```diff
@@ -5,7 +5,7 @@ click==6.7
 configparser==3.5.0
 dash==1.0.2
 dash-auth==1.2.0
-dash-bio==0.1.3rc4
+dash-bio==0.1.4
```
mkcor:
```suggestion
dash-bio==0.1.3
```

shammamah: 0.1.3 does not work with dash 1.0 and above. The correct version is 0.1.4.

shammamah:
@mkcor In the PR for the `0.1.4` version, you can see the changes that I made to the package only amount to the autogenerated docstrings (introduced to Dash in https://github.com/plotly/dash/pull/782), as well as regeneration of auto-generated files. I also changed around one keyword argument in one of the apps, as that has also been changed with `dash-core-components`.  

I didn't want to merge something into `master` if it was failing tests, even if the tests themselves are outdated. This was a quick release to fix an immediate problem (https://github.com/plotly/dash-bio/pull/412#issuecomment-517771664). As I mentioned in the PR, I've rebased the tests-refactoring branch on the branch in https://github.com/plotly/dash-bio/pull/412. This one should be good to merge.

requirements.txt
mkcor: Sorry, I am more confused after your [explanations](https://github.com/plotly/dash-docs/pull/614#issuecomment-518241526) than before!! :( If I understand correctly, there is no functional difference between `dash-bio 0.1.3` and `dash-bio 0.1.4` (only updated docstrings).

Plus, the [updates](https://github.com/plotly/dash-docs/pull/614/files) you're introducing in this PR correspond precisely to what shipped with `dash-bio 0.1.3`: https://github.com/plotly/dash-bio/blob/master/CHANGELOG.md#013---2019-07-31

So, logically, I would expect an update to `dash-bio 0.1.4` as a later step...

mkcor:
> @mkcor In the PR for the `0.1.4` version, you can see the changes that I made to the package only amount to the autogenerated docstrings (introduced to Dash in [plotly/dash#782](https://github.com/plotly/dash/pull/782)), as well as regeneration of auto-generated files.

PR https://github.com/plotly/dash/pull/782 was merged ages ago (June 22)... So, if the changes it introduced weren't reflected in the `dash-bio` package yet, we should definitely address this, but maybe separately? Why not update the `dash-bio` dependency of `dash-docs` first to `0.1.3`? And then to `0.1.4` (as a subsequent PR)?

> I didn't want to merge something into `master` if it was failing tests, even if the tests themselves are outdated. This was a quick release to fix an immediate problem ([plotly/dash-bio#412 (comment)](https://github.com/plotly/dash-bio/pull/412#issuecomment-517771664)). 

Sure, you can't merge a PR if it's not approved or if the tests are failing. I'm thinking we could have taken out the tests for that very release (`0.1.4`) because it seems odd to me to have https://pypi.org/project/dash-bio/0.1.4/ up with no corresponding version over at https://github.com/plotly/dash-bio (in current `master`)...

> As I mentioned in the PR, I've rebased the tests-refactoring branch on the branch in [plotly/dash-bio#412](https://github.com/plotly/dash-bio/pull/412). This one should be good to merge.

Ok. I can see that #411 is a *draft* PR for now; I'll wait for your review request.

shammamah:
>Why not update the dash-bio dependency of dash-docs first to 0.1.3? 
`0.1.3` is not compatible with `dash-1.0.0` and above. The main motivation behind creating/releasing `0.1.4` is exactly that, and the point of this PR is to have compatibility -- not just to use an official release (as opposed to a release candidate).

>I'm thinking we could have taken out the tests for that very release (0.1.4)
I had considered this too, but I personally am not a fan of it. 

>Ok. I can see that #411 is a draft PR for now; I'll wait for your review request.
What I meant is that *this* PR is OK to merge without that PR being merged; `0.1.4` has already been released.

mkcor:
> `0.1.3` is not compatible with `dash-1.0.0` and above.

Can you please tell me why? That's the part I've missed. Basically, besides the updated docstrings (which shouldn't change anything functionality-wise, please correct me if I'm wrong), what are the changes brought by `0.1.4` (wrt `0.1.3`)?

> (...) the point of this PR is to have compatibility -- not just to use an official release (as opposed to a release candidate).

This part I got, I wasn't focusing on the release candidate vs. official release thing.

> > I'm thinking we could have taken out the tests for that very release (0.1.4)
> 
> I had considered this too, but I personally am not a fan of it.

Ok, good to know; the conversation (resp. decision) could have taken place (resp. been announced) over at https://github.com/plotly/dash-bio/pull/412 but you didn't tag me or anything, so I hadn't followed.

> > Ok. I can see that #411 is a draft PR for now; I'll wait for your review request.
> 
> What I meant is that _this_ PR is OK to merge without that PR being merged;

Oh, ok!

> `0.1.4` has already been released.

I know, that's exactly what I'm pointing out as... *wrong*(?) insofar as we (let me copy-paste myself) "have https://pypi.org/project/dash-bio/0.1.4/ up with no corresponding version over at https://github.com/plotly/dash-bio (in current master)..." I have a problem with that, but I'm more than happy to be taught a thing or two. If @chriddyp or @alexcjohnson say it is *right*, I'll follow along.

shammamah:
>Can you please tell me why? That's the part I've missed. 

Sure! If you read the changes to `package.json`, you'll see that I updated the versions of `react` so that the JS distribution of the `dash-bio` package is on React 16, which is required for compatibility with Dash 1.0.0 and above (IIRC). I had to do the same thing for `dash-daq` here: https://github.com/plotly/dash-daq/pull/58

That's the main reason for the version update. Hope that clears things up.

mkcor:
> Hope that clears things up.

It does, thank you so much!

Please sync up with `master` before merging :)

PR #623, opened 2019-08-08
explain that runtime.txt is an optional file.

michaelbabyn:
Part of https://github.com/plotly/streambed/issues/13345

Post-merge checklist:

- [ ] I understand

### DIFF
--- a/tutorial/dash_deployment_server_examples.py
+++ b/tutorial/dash_deployment_server_examples.py
@@ -744,7 +744,7 @@ Requirements = html.Div(children=[
 
     `runtime.txt`
 
-    This file specifies python runtime. For example, its contents would be
+    This optional file specifies python runtime. For example, its contents would be
     `python-2.7.15` or `python-3.6.6`.
 
     ***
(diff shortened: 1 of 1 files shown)


tutorial/dash_deployment_server_examples.py
```diff
@@ -744,7 +744,7 @@ def display_instructions_deploy(method):
 
     `runtime.txt`
 
-    This file specifies python runtime. For example, its contents would be
+    This optional file specifies python runtime. For example, its contents would be
```
alexcjohnson: What happens if you omit it? Which Python version do you get?

michaelbabyn: You get 3.6.7 at the moment. Whatever is defined as the default [in the buildpack](https://github.com/plotly/heroku-buildpack-python/blob/e4307fc2c0f3e531aac2cdafc916db9905f69480/bin/compile#L52).

alexcjohnson: Good. Please include exactly that in the text here.

PR #625, opened 2019-08-12
495 prevent update doc

### LINKED ISSUE #495: Add preventUpdate to the docs
It took me a while to dig through the community forum to find it, i think it might be useful to people if we add a section about how to abort / skip callback to the FAQ section: https://github.com/plotly/dash/pull/190

alinastarkov:
Closes #495 
Remove `s == textwrap.dedent` from dcc.Markdown

Post-merge checklist:

- [x] I understand

### DIFF
--- a/tutorial/chapter_index.py
+++ b/tutorial/chapter_index.py
@@ -102,12 +102,13 @@ chapters = {
     'state': {
         'url': '/state',
         'content': state.layout,
-        'name': 'Part 4. Callbacks With State',
+        'name': 'Part 4. More About Callbacks',
         'description': 'Basic callbacks are fired whenever the values ' \
                        'change. Use Dash `State` with Dash `Inputs` to ' \
                        'pass in extra values whenever the `Inputs` ' \
                        'change. `State` is useful for UIs that contain ' \
-                       'forms or buttons.'
+                       'forms or buttons. Use `PreventUpdate` Exception ' \
+                       'to leave the callback output unchanged' \
     },
 
     'graphing': {
--- a/tutorial/chapter_index.py
+++ b/tutorial/chapter_index.py
@@ -107,7 +107,7 @@ chapters = {
                        'change. Use Dash `State` with Dash `Inputs` to ' \
                        'pass in extra values whenever the `Inputs` ' \
                        'change. `State` is useful for UIs that contain ' \
-                       'forms or buttons. Use `PreventUpdate` Exception ' \
+                       'forms or buttons. Use the `PreventUpdate` exception ' \
                        'to leave the callback output unchanged' \
     },
 
--- a/tutorial/examples/prevent_update.py
+++ b/tutorial/examples/prevent_update.py
@@ -2,27 +2,45 @@ import dash
 import dash_core_components as dcc
 import dash_html_components as html
 from dash.dependencies import Input, Output
-from dash.exceptions import PreventUpdate
 
 external_stylesheets = ['https://codepen.io/chriddyp/pen/bWLwgP.css']
 
 app = dash.Dash(__name__, external_stylesheets=external_stylesheets)
 
 app.layout = html.Div([
-    html.Button('Click here to see the content', id='button'),
-    html.Div(id='body-div')
+    html.P('Enter a composite number to see its prime factors'),
+    dcc.Input(id='num', type='number', debounce=True, min=1, step=1),
+    html.P(id='err', style={'color': 'red'}),
+    html.P(id='out')
 ])
 
-
 @app.callback(
-    Output(component_id='body-div', component_property='children'),
-    [Input(component_id='button', component_property='n_clicks')]
+    [Output('out', 'children'), Output('err', 'children')],
+    [Input('num', 'value')]
 )
-def update_output(n_clicks):
-    if (n_clicks == None):
-        raise PreventUpdate;
-    else: 
-        return "Elephants are the only animal that can't jump"
+def show_factors(num):
+    if num is None:
+        # PreventUpdate prevents ALL outputs updating
+        raise dash.exceptions.PreventUpdate
+
+    factors = prime_factors(num)
+    if len(factors) == 1:
+        # dash.no_update prevents any single output updating
+        # (note: it's OK to use for a single-output callback too)
+        return dash.no_update, '{} is prime!'.format(num)
+
+    return '{} is {}'.format(num, ' * '.join(str(n) for n in factors)), ''
+
+def prime_factors(num):
+    n, i, out = num, 2, []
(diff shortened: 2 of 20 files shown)


tutorial/examples/prevent_update.py
```diff
@@ -0,0 +1,28 @@
+import dash
+import dash_core_components as dcc
+import dash_html_components as html
+from dash.dependencies import Input, Output
+from dash.exceptions import PreventUpdate
+
```
alinastarkov: I created a simple example but I make a more complex one if needed

tutorial/chapter_index.py
```diff
@@ -102,12 +102,13 @@
...
         'url': '/state',
         'content': state.layout,
-        'name': 'Part 4. Callbacks With State',
+        'name': 'Part 4. More About Callbacks',
         'description': 'Basic callbacks are fired whenever the values ' \
                        'change. Use Dash `State` with Dash `Inputs` to ' \
                        'pass in extra values whenever the `Inputs` ' \
                        'change. `State` is useful for UIs that contain ' \
-                       'forms or buttons.'
+                       'forms or buttons. Use `PreventUpdate` Exception ' \
```
alexcjohnson:
```suggestion
                       'forms or buttons. Use the `PreventUpdate` exception ' \
```

tutorial/examples/prevent_update.py
```diff
@@ -0,0 +1,28 @@
...
+    html.Div(id='body-div')
+])
+
+
+@app.callback(
+    Output(component_id='body-div', component_property='children'),
+    [Input(component_id='button', component_property='n_clicks')]
+)
+def update_output(n_clicks):
+    if (n_clicks == None):
```
alexcjohnson:
```suggestion
    if n_clicks is None:
```

tutorial/getting_started_part_2.py
```diff
@@ -271,12 +271,12 @@
 
 
     dcc.Markdown(s('''
-        The next part of the Dash tutorial covers an additional concept of
-        Dash callbacks: `State`
+        The next part of the Dash tutorial covers additional concept of
```
alexcjohnson:
```suggestion
        The next part of the Dash tutorial covers additional concepts of
```

tutorial/getting_started_part_2.py
```diff
@@ -271,12 +271,12 @@
...
+        Dash callbacks: `State` and Exception
```
alexcjohnson:
```suggestion
        Dash callbacks: `State` and `PreventUpdate`
```

tutorial/state.py
```diff
@@ -57,14 +59,14 @@
...
 
-    dcc.Markdown('''
-    ***
-
-    The next chapter of the user guide explains how to use callback
-    principles with the `dash_core_components.Graph` component
-    to make applications that
-    respond to interactions with graphs on the page.
-    '''.replace('    ', '')),
+    dcc.Markdown(s('''
```
alexcjohnson: These `s == textwrap.dedent` wrappers are unnecessary in dash 1.0 https://github.com/plotly/dash-core-components/pull/569 - please remove them when working in a file that still has them. Also I notice some trailing whitespace 🔪

tutorial/state.py
```diff
@@ -7,13 +7,15 @@
...
 examples = {
     'basic-input': tools.load_example('tutorial/examples/basic-input.py'),
-    'basic-state': tools.load_example('tutorial/examples/basic-state.py')
+    'basic-state': tools.load_example('tutorial/examples/basic-state.py'),
+    'prevent-update': tools.load_example('tutorial/examples/prevent_update.py'),
 }
 
 layout = html.Div([
-    html.H1('Dash State'),
+    html.H1('Dash State and Exceptions'),
```
alexcjohnson: Again, this isn't about all exceptions, it's specifically about `PreventUpdate`.

tutorial/state.py
alexcjohnson: Also, minor point - you can merge this and the previous `dcc.Markdown`

tutorial/examples/prevent_update.py
alexcjohnson: This example is good (and fun! 🐘) but since we discussed multiple outputs already in the previous section, we should also add an example of `dash.no_update` for partial updates to the same page where this example is shown. Perhaps the simplest example is keeping a previous valid output while reporting an error. Here's one I like:
```py
app.layout = html.Div([
    html.P('Enter a composite number to see its prime factors'),
    dcc.Input(id='num', type='number', debounce=True, min=1, step=1),
    html.P(id='err', style={'color': 'red'}),
    html.P(id='out')
])

@app.callback(
    [Output('out', 'children'), Output('err', 'children')],
    [Input('num', 'value')]
)
def show_factors(num):
    if num is None:
        # PreventUpdate prevents ALL outputs updating
        raise dash.exceptions.PreventUpdate

    factors = prime_factors(num)
    if len(factors) == 1:
        # dash.no_update prevents any single output updating
        # (note: it's OK to use for a single-output callback too)
        return dash.no_update, '{} is prime!'.format(num)

    return '{} is {}'.format(num, ' * '.join(str(n) for n in factors)), ''

def prime_factors(num):
    n, i, out = num, 2, []
    while i * i <= n:
        if n % i == 0:
            n = int(n / i)
            out.append(i)
        else:
            i += 1 if i == 2 else 2
    out.append(n)
    return out
```

alexcjohnson: What I meant was we should keep both examples - yours to show basic usage because that will help people understand the concept, plus mine to show how to use it for partial update with multiple outputs.

alexcjohnson:
Looking great. Thanks for 🔪 all those extra `textwrap.dedent` calls - and for fixing the instructions encouraging their use! Just the one comment about including both examples https://github.com/plotly/dash-docs/pull/625#discussion_r313518889 and I think this will be good to go.

tutorial/examples/prevent_update_button.py
```diff
@@ -0,0 +1,27 @@
...
+    html.Button('Click here to see the content', id='button'),
+    html.Div(id='body-div')
+])
+
+@app.callback(
+    Output(component_id='body-div', component_property='children'),
+    [Input(component_id='button', component_property='n_clicks')]
+)
+def update_output(n_clicks):
+    if n_clicks == None:
```
alexcjohnson: Per my earlier [comment](https://github.com/plotly/dash-docs/pull/625#discussion_r313131138) and to keep all our examples consistent -> https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_none.html

PR #639, opened 2019-08-30
WIP - [merge after 1.3.0 release]  document about remote driver

byronz:
Post-merge checklist:

- [x] I understand


tutorial/testing.py
```diff
@@ -43,6 +44,39 @@
     can check the details from both [Firefox](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Headless_mode)
     and [Chrome](https://developers.google.com/web/updates/2017/04/headless-chrome).
 
+    Remote WebDriver is added in Dash *1.3.0*. You may have two use cases:
```
```suggestion
    Remote WebDriver support is added in Dash *1.3.0*. There are two ways to use it:
```

tutorial/testing.py
```diff
@@ -43,6 +44,39 @@
...
+
+    1. Run `pytest --remote --log-cli-level DEBUG -k bsly001` to grab a Chrome
+    WebDriver from a local hosted grid at `http://localhost:4444/wd/hub`
+    2. Run `pytest --webdriver Firefox --log-cli-level DEBUG --remote-url https://grid_provioder_endpoints`
```
```suggestion
    2. Run `pytest --webdriver Firefox --log-cli-level DEBUG --remote-url https://grid_provider_endpoint`
```

Is `--log-cli-level` necessary here? Or is it just important for debugging connection to the remote? If the latter, perhaps move that to the Caveats section?

Similarly, I guess `--webdriver Firefox` is required if your grid is running FF? Perhaps just mention this then **"to connect with a remote grid in the cloud running Firefox"**

tutorial/testing.py
```diff
@@ -43,6 +44,39 @@
...
+    to connect with a remote grid in the cloud. Note that you don't need to use
+    `--remote` as soon as the `--remote-url` value is set and different than
+    the default one.
+
+    ### Caveats
+
+    It's important to note that we cannot fully test and guarantee that the
+    above cases will work with any given selenium grid you have. The limitation
+    might come from how the network is set up, the limitation of different
+    hosting OS and how the docker-compose was configured.
```
```suggestion
    hosting OS or how docker-compose was configured.
```

tutorial/testing.py
```diff
@@ -43,6 +44,39 @@
...
+
+    You might need to do some extra WebDriver Options Tuning to have the tests
+    running in a particular Selenium-Grid. And there is a back door open for
+    this purpose with `pytest_setup_options` hook defined in `plugin.py`. The
```
```
    You might need some extra WebDriver Options tuning to run the tests
    in a particular Selenium-Grid. There is a back door open for
    this purpose with the `pytest_setup_options` hook defined in `plugin.py`. The
```
(still waiting for multiline suggestions support...)

tutorial/testing.py
byronz: yes, I think users would need the DEBUG level to see the communications on both sides. --webdriver here is to emphasize that you still need to set the browser type here, should also add a note that chrome is the default one

PR #646, opened 2019-09-11
More snapshots 638

byronz:
Post-merge checklist:

- [x] I understand


.circleci/config.yml
```diff
@@ -32,8 +39,9 @@ jobs:
           name: Run tests
           command: |
               . venv/bin/activate
-              python --version
-              python -m unittest tests.test_integration.Tests
+              git clone --depth 1 https://github.com/plotly/dash.git dash-main
```
byronz: this can be removed once the change on dash side get released

alexcjohnson: why not do this in the `requirements.txt`? See example #6 https://pip.readthedocs.io/en/stable/reference/pip_install/#examples
`git+https://git.repo/some_pkg.git#egg=SomePackage[PDF]` (is `-e` important to us?)

alexcjohnson: oh I see, you don't want to include `dash[testing]` as part of requirements.txt for regular use, just on CI. So yes, cleaner with pip than straight `git clone` but we can leave it outside of `requirements.txt`. I still might have moved it up to the "Install dependencies" step rather than "Run tests" but that's not a big deal.

PR #652, opened 2019-09-17
add note re: append_script wrt serve_locally=True to 1.0 migration guide

### LINKED ISSUE #651: mention that `append_script` won't load external scripts when `serve_locally=True`
We should add a note that `app.scripts.append_script` will cause an error when trying to load an external js script if `serve_locally=True`, and that `external_scripts` should be used instead. I think this can go in the 1.0 migration guide since because upgrading to 1.0 would surface this error.
```
A local version of https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.4/MathJax.js?config=TeX-MML-AM_CHTML is not available
```

michaelbabyn:
resolves #651 

Post-merge checklist:

- [ ] I understand


michaelbabyn:
In the case I saw it was an old app that a user had just updated to 1.0 and `app.scripts.append_script` was in there for a while without causing any issues but I just did a search and it looks like we do have some old example apps that use it too.

cldougl:
@michaelbabyn do you think you could make an issue to update the old examples you mentioned (w/ a list of the old examples/where you're seeing it)
It would be good if we could get all all of our examples updated past the dash 1.x.x update

PR #655, opened 2019-09-25
Issue 606 - Document clientside callbacks

### LINKED ISSUE #606: document clientside callbacks
Not something we necessarily *want* everyone to start using, but we have it so there should be docs.
https://github.com/plotly/dash-renderer/pull/143
https://github.com/plotly/dash/pull/672

Marc-Andre-Rivet:
Fixes #606 

Post-merge checklist:

- [x] I understand

### DIFF
--- a/tutorial/examples/table/editing_loading_state.py
+++ b/tutorial/examples/table/editing_loading_state.py
@@ -1,4 +1,6 @@
 from time import sleep
+
+import random
 from random import randint
 
 import dash
@@ -7,6 +9,9 @@ from dash.dependencies import Input, Output
 import dash_table
 import dash_core_components as dcc
 import dash_html_components as html
+
+# For the documentation to always render the same values
+random.seed(0)
 
 app = dash.Dash(__name__)
 
--- a/tutorial/examples/table/editing_loading_state.py
+++ b/tutorial/examples/table/editing_loading_state.py
@@ -1,7 +1,6 @@
 from time import sleep
 
-import random
-from random import randint
+from random import randint, seed
 
 import dash
 from dash.exceptions import PreventUpdate
@@ -11,7 +10,7 @@ import dash_core_components as dcc
 import dash_html_components as html
 
 # For the documentation to always render the same values
-random.seed(0)
+seed(0)
 
 app = dash.Dash(__name__)
 
--- a/tutorial/performance.py
+++ b/tutorial/performance.py
@@ -109,10 +109,6 @@ considerably by introducing `restyle` calls into this logic. If you or
 your company would like to sponsor this work,
 [get in touch](https://plot.ly/products/consulting-and-oem/).
 
-'''),
-
-    dcc.Markdown('''
-
 ***
 
 ### Clientside Callbacks
@@ -190,10 +186,6 @@ execute on the browser's main thread and that they will block rendering
 and events processing while the being processed. Dash does not currently
 support asynchronous clientside callbacks and will fail if a `Promise`
 is returned.
-
-'''),
-
-    dcc.Markdown('''
 
 ***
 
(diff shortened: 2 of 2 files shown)


tutorial/examples/table/editing_loading_state.py
```diff
@@ -8,6 +10,9 @@
 import dash_core_components as dcc
 import dash_html_components as html
 
+# For the documentation to always render the same values
+random.seed(0)
```
Marc-Andre-Rivet: @shammamah This random has some interesting effects when running percy diffs against this page 😁 

Seed the random so as to always output the same values. This prevents percy from identifying diffs on every run.

alexcjohnson: cc @wbrgss ^^

tutorial/performance.py
```diff
@@ -109,6 +109,92 @@ def slow_function(input):
 your company would like to sponsor this work,
 [get in touch](https://plot.ly/products/consulting-and-oem/).
 
+'''),
+
+    dcc.Markdown('''
```
alexcjohnson: minor: could be combined with the previous `dcc.Markdown`?

tutorial/performance.py
```diff
@@ -109,6 +109,92 @@ def slow_function(input):
...
+
+**Note**: It is important to keep in mind that clientside callbacks will
+execute on the browser's main thread and that they will block rendering
+and events processing while the being processed. Dash does not currently
+support asynchronous clientside callbacks and will fail if a `Promise`
+is returned.
+
+'''),
+
+    dcc.Markdown('''
```
alexcjohnson: same

tutorial/performance.py
```diff
@@ -109,6 +109,92 @@ def slow_function(input):
...
+
+***
+
+### Clientside Callbacks
+
+Sometimes callbacks can incur a significant overhead, especially when they :
+- receive and/or return very large quantities of data (transfer time)
+- are called very often (queuing and handshake)
```
alexcjohnson: I'd specifically call out network latency here - this is what I've seen as the most common reason for using clientside, especially if the app is deployed in some restricted environment where there may be a lot of extra overhead due to VPNs, corporate monitoring, or whatever.

In fact I'd include some sort of a note about clientside callbacks not being possible when you need to refer to global variables on the server, or of course DB calls... which is very often the case when you have large quantities of data involved.

tutorial/performance.py
Marc-Andre-Rivet: Sure - this was actually done voluntarily to make it clear they were different blocks but I don't mind.

Marc-Andre-Rivet: https://github.com/plotly/dash-docs/pull/655/commits/e3bdabf1ecddad38d2416694b51100f27b3d38ec

tutorial/examples/table/editing_loading_state.py
```diff
@@ -1,4 +1,6 @@
 from time import sleep
+
+import random
 from random import randint
```
alexcjohnson: some linters will complain about `random` being imported 2 ways - so either just do `import random` or `from random import randint, seed`

tutorial/performance.py
alexcjohnson: Yeah, understandable. But I think we want to err on the side of improved performance here; and markdown is pretty self-documenting about where the blocks break. Thanks for the change.

tutorial/performance.py
Marc-Andre-Rivet: https://github.com/plotly/dash-docs/pull/655/commits/323c116438c130baa574bdd13b655cba8fa3099a

tutorial/examples/table/editing_loading_state.py
Marc-Andre-Rivet: https://github.com/plotly/dash-docs/pull/655/commits/323c116438c130baa574bdd13b655cba8fa3099a

PR #657, opened 2019-10-01
update changelog links and text

alexcjohnson:
Companion to https://github.com/plotly/dash/pull/947 - to be merged about the same time.
Also removes an obsolete file `changelog.py`


alexcjohnson:
I merged https://github.com/plotly/dash/pull/947 - but actually I realize we DON'T want to merge this PR just yet, since these links point to `master`, which will only get updated when we issue a new release.

So I'll leave this to be merged upon the next Dash release - unless someone would like to argue for pointing to the `dev` branch changelogs, which show the `UNRELEASED` section.

Marc-Andre-Rivet:
Changed destination to `update-dash-140`, merging! 🚀

PR #662, opened 2019-10-03
exclude datepickers from base css td/th overrides

alexcjohnson:
The docs portion of https://github.com/plotly/dash-core-components/issues/658


alexcjohnson:
Before:
Screen Shot 2019-10-03 at 11 32 09 AM
After:
Screen Shot 2019-10-03 at 11 31 25 AM

PR #669, opened 2019-10-15
remove google analytics tags as we have the google tag manager

VeraZab:
relates to: https://github.com/plotly/marketing-team/issues/118

having both google analytics and [google tag manager](https://github.com/plotly/dash-docs/blob/b10701c8514b87d86645f6edeb808ccdcfa4143d/run.py#L283) double counts pageviews


tutorial/examples/external_css_and_js/external-resources-init.py
```diff
@@ -4,7 +4,6 @@
 
 # external JavaScript files
 external_scripts = [
-    'https://www.google-analytics.com/analytics.js',
```
alexcjohnson: This one we should keep, it never gets executed, it's just displayed as an example of the simplest way to load an external script on your page.

Good call on removing `plotly_ga.js` though!

nicolaskruchten:
💃

PR #683, opened 2019-10-31
update dash.testing doc to 1.5.1

byronz:
Post-merge checklist:

- [ ] I understand


.circleci/config.yml
```diff
@@ -38,8 +38,8 @@ jobs:
           name: Run tests
           command: |
               . venv/bin/activate
-              pip install -e git+https://github.com/plotly/dash.git#egg=dash[testing]
-              pytest --junitxml=test-reports/snapshots.xml tests/snapshots
+              pip install -e git+https://github.com/plotly/dash.git@dbg-dt-callbacks#egg=dash[testing]
```
byronz: with 1.5.x, I found some tricky edge cases for `visit_and_snapshot` API to handle properly,  this branch contains the  corresponding tuning work

tutorial/testing.py
```diff
@@ -182,6 +182,66 @@ def test_bsly001_falsy_child(dash_duo):
     variable, so the regression result is only available in Plotly's CircleCI
     setup.
 
+    ## Dash for R testing
```
byronz: @rpkyle  this part basically reflects to the changes we did for hot-reloading work.

requirements.txt
```diff
@@ -13,7 +13,6 @@ dash-canvas>=0.0.10
 decorator==4.3.0
 enum34==1.1.6
 flake8==3.5.0
-Flask==1.0
```
rpkyle: @byronz Did the Flask requirement get rolled into another list of packages somewhere? Surprised to see it deleted here.

tutorial/testing.py
```diff
@@ -182,6 +182,66 @@ def test_bsly001_falsy_child(dash_duo):
...
+        # folder You can change this with `start_server(app, cwd='/my/cwd/')`
+        dashr.start_server(app)
+
+        assert dashr.find_element("#container").text == "Hello Dash for R Testing"
+
+
+    def test_rstr002_r_with_file_path(dashr):
+
+        # Alternatively, the app can be a filepath defining the Dash for R
+        # the `os.path.dirname(__file__)` is an useful Python trick to get the path
```
rpkyle: @byronz Hmm, do we want to advertise this approach? It seems that @alexcjohnson's `cwd`-setting feature should help most users avoid the need to ☝️ retrieve the file path using this method.

tutorial/testing.py
```diff
@@ -310,64 +370,38 @@ def test_bsly001_falsy_child(dash_duo):
     | `toggle_window()` | switch between the current working window and the newly opened one. |
     | `switch_window(idx)` | switch to window by window index. shortcut to `driver.switch_to.window`. raise `BrowserError` if no second window present in browser |
     | `open_new_tab(url=None)` | open a new tab in browser with window name `new window`. `url` if not set, equals to `server_url` |
-    | `percy_snapshot(name)` | visual test API shortcut to `percy_runner.snapshot` it also combines the snapshot `name` with python versions |
-    | `visit_and_snapshot(resource_path, hook_id, assert_check=True)` | common task used in dash-docs testing: it visits a URL path by `resource_path`, makes sure the page is fully loaded by the `hook_id` at the bottom, takes a snapshot and returns to the main page. `assert_check` is a switch to enable/disable an assertion that there is no devtools error alert icon |
+    | `percy_snapshot(name, wait_for_callbacks=False)` | visual test API shortcut to `percy_runner.snapshot` it also combines the snapshot `name` with the actual python versions. `wait_for_callbacks` controls if the snapshot is taken until all dash callbacks are fired, default False. |
```
rpkyle: @byronz I would reword "`wait_for_callbacks` controls if the snapshot is taken until all dash callbacks are fired, default False." to

"The `wait_for_callbacks` parameter controls whether the snapshot is taken only after all callbacks have fired; the default is `False`."

tutorial/testing.py
```diff
@@ -310,64 +370,38 @@ def test_bsly001_falsy_child(dash_duo):
...
+    | `visit_and_snapshot(resource_path, hook_id, wait_for_callbacks=True, assert_check=True)` | common task used in dash-docs testing: it visits a URL path by `resource_path`, makes sure the page is fully loaded by the `hook_id` at the bottom, takes a snapshot and returns to the main page.  `wait_for_callbacks` controls if the snapshot is taken until all dash callbacks are fired, default True. `assert_check` is a switch to enable/disable an assertion that there is no devtools error alert icon |
```
rpkyle: For "common task used in dash-docs testing: it visits a URL path by `resource_path`, makes sure the page is fully loaded by the `hook_id` at the bottom, takes a snapshot and returns to the main page" I might suggest rewording as 

"This method automates a common task during dash-docs testing: the URL described by `resource_path` is visited, and completion of page loading is assured by waiting until the element described by `hook_id` is fetched. Once `hook_id` is available, `visit_and_snapshot` acquires a snapshot of the page."

tutorial/testing.py
```diff
@@ -182,6 +182,66 @@ def test_bsly001_falsy_child(dash_duo):
...
+    ## Dash for R testing
+
+    We released [Dash for R](https://medium.com/@plotlygraphs/announcing-dash-for-r-82dce99bae13)
+    in July 2019.  To facilitate testing it, we extended the *Python* package
+    `dash.testing` to support Dash for R apps.
+
+    The new `dashr` fixture gives us a test development experience in
+    Dash for R that's nearly identical to the `dash_duo` fixture in
+    Dash for Python. In this context, a `dashr` fixture is an instance of
+    Python class which provides a reliable baseline to execute the app and test
```
```suggestion
    Python class which provides a reliable framework to execute the app and test
```
"Baseline" to me means some complete state that you add a particular perturbation to, but `dashr` isn't complete on its own, it's more of a structure into which you insert your test app.

tutorial/testing.py
```diff
@@ -310,64 +370,38 @@ def test_bsly001_falsy_child(dash_duo):
...
+    | `visit_and_snapshot(resource_path, hook_id, wait_for_callbacks=True, assert_check=True)` | common task used in dash-docs testing: it visits a URL path by `resource_path`, makes sure the page is fully loaded by the `hook_id` at the bottom, takes a snapshot and returns to the main page.  `wait_for_callbacks` controls if the snapshot is taken until all dash callbacks are fired, default True. `assert_check` is a switch to enable/disable an assertion that there is no devtools error alert icon |
     | `take_snapshot(name)` | hook method to take a snapshot while selenium test fails. the snapshot is placed under `/tmp/dash_artifacts` in Linux or `%TEMP` in windows with a filename combining test case `name` and the running selenium session id |
+    | `zoom_in_graph_by_ratio(elem_or_selector, start_fraction=0.5, zoom_box_fraction=0.2, compare=True)` | zoom out a graph (provided with either a Selenium WebElement or CSS selector) with a zoom box fraction of component dimension, default start at middle with a rectangle of 1/5 of the dimension use `compare` to control if we check the SVG get changed |
```
rpkyle: I might rewrite this as

"Given a specified fraction of the zoom box for an individual component, zoom out from a graph (identified either using a Selenium WebElement or CSS selector). The default specifies a rectangular zoom box which is 1/5 the dimension of the graph itself. The `compare` option controls whether the resulting SVG should be compared with the previous capture for changes."

tutorial/testing.py
```diff
@@ -310,64 +370,38 @@ def test_bsly001_falsy_child(dash_duo):
...
+    | `click_at_coord_fractions(elem_or_selector, fx, fy)` |  Use `ActionChains` to click a Selenium WebElement with a coordination fractions |
```
rpkyle: Would rewrite "with a coordination fractions" as "given a coordination fraction".

tutorial/testing.py
```diff
@@ -310,64 +370,38 @@ def test_bsly001_falsy_child(dash_duo):
...
-        # You can change this with eg: dashr.start_server(app, cwd='/my/path/')
-        dashr.start_server(app)
-        assert dashr.find_element("#container").text == "Hello Dash for R Testing"
+    This section enumerates a full list of Dash App related properties and APIs
+    apart from the previous browser ones.
 
+    | API | Description |
+    | --- | ----------- |
+    | `devtools_error_count_locator` | property returns the selector of the error count number in the devtool UI |
+    | `dash_entry_locator` | property returns the selector of react entry point, it can be used to verify if an Dash is loaded |
```
rpkyle: For "if an Dash is loaded", do we mean instead "if a Dash app is loaded"?

tutorial/testing.py
```diff
@@ -310,64 +370,38 @@ def test_bsly001_falsy_child(dash_duo):
...
+    | `dash_outerhtml_dom` | property returns the BeautifulSoup parsed Dash DOM from outerHTML |
+    | `dash_innerhtml_dom` | property returns the BeautifulSoup parsed Dash DOM from innerHTML |
+    | `redux_state_paths` | property returns the `window.store.getState().paths` |
+    | `redux_state_rqs` | property returns `window.store.getState().requestQueue` |
+    | `window_store` | property returns `window.store` |
+    | `get_local_storage(store_id="local")` | get the local storage by the id, default to `local` |
```
rpkyle: Suggest changing "get the local storage by the id" to "get the value of the local storage item given its id, default is `local`."

tutorial/testing.py
```diff
@@ -182,6 +182,66 @@ def test_bsly001_falsy_child(dash_duo):
...
+    app$layout(htmlDiv(list(htmlDiv(id='container',children='Hello Dash for R testing'))))
+    app$run_server()
+    '''
+
+    # a test case is a simple Python function with the same prefix convention
+    # dashr is the default fixture combines the API for serving the app
+    # and selenium tests.
+    def test_rstr001_r_with_string(dashr):
+
+        # Since the app is the string chunck we defined above, the dashr
```
```suggestion
        # Since the app is the string chunk we defined above, the dashr
```

tutorial/testing.py
```diff
@@ -182,6 +182,66 @@ def test_bsly001_falsy_child(dash_duo):
...
+        # start_server by default uses the temporary folder as current working
+        # folder You can change this with `start_server(app, cwd='/my/cwd/')`
+        dashr.start_server(app)
+
+        assert dashr.find_element("#container").text == "Hello Dash for R Testing"
+
+
+    def test_rstr002_r_with_file_path(dashr):
+
+        # Alternatively, the app can be a filepath defining the Dash for R
```
```suggestion
        # Alternatively, the app can be a path to a separate file containing your Dash for R app
```

tutorial/testing.py
```diff
@@ -182,6 +182,66 @@ def test_bsly001_falsy_child(dash_duo):
...
+        # the `os.path.dirname(__file__)` is an useful Python trick to get the path
+        # of a test path, so it will always work no matter where you start the
```
```suggestion
        # of the currently running script, so it will always work no matter where you start the
```

tutorial/testing.py
alexcjohnson: "coordinate," not "coordination." Perhaps for clarity:
```suggestion
    | `click_at_coord_fractions(elem_or_selector, fx, fy)` |  Use `ActionChains` to click a Selenium WebElement at a location a given fraction of the way `fx` between its left (0) and right (1) edges, and `fy` between its top (0) and bottom (1) edges. |
```

requirements.txt
byronz: dash already set a flask version, and I found that it starts to have some version dep conflicts now, https://github.com/plotly/dash/blob/dev/requires-install.txt

tutorial/testing.py
byronz: good point, I don't have a preference here, up to you as R chef

tutorial/testing.py
byronz: you can use ```suggestion ``` next time :)

tutorial/testing.py
rpkyle: My inclination would be to avoid mentioning `os.path.dirname(__file__)` and encourage users to pass a path via `cwd` if they have a reason to override the default. I suspect the vast majority of users will be OK with the default, but I may well be wrong.

byronz: the `__file__` trick here is to solve a slightly different problem than `cwd`, it makes it more flexible to address the full path, especially under the context of pytest. 
cwd=None (by default) is still going to work as expected.

PR #689, opened 2019-11-06
Revert "update dash.testing doc to 1.5.1"

chriddyp:
Reverts plotly/dash-docs#683

dash.plot.ly is down right now, seems to be from this PR


BRONSOLO:
:dancer:

PR #691, opened 2019-11-08
added layout animation in example

emmanuelle:
Post-merge checklist:

- [x] I understand

### DIFF
--- a/tutorial/examples/getting_started_graph.py
+++ b/tutorial/examples/getting_started_graph.py
@@ -55,7 +55,8 @@ def update_figure(selected_year):
             yaxis={'title': 'Life Expectancy', 'range': [20, 90]},
             margin={'l': 40, 'b': 40, 't': 10, 'r': 10},
             legend={'x': 0, 'y': 1},
-            hovermode='closest'
+            hovermode='closest',
+            transition = {'duration': 500},
         )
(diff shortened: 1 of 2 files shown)


emmanuelle:
I don't understand the CI fail for python 2.7. Seems to be related to snapshots. Could the transitions prevent the snapshots from behaving correctly? (are they tested only for python 2??)

alexcjohnson:
I doubt it has anything to do with transitions, probably just a spurious error (perhaps related to https://github.com/plotly/dash/pull/1001) but then looks like you tried to rerun the same test - which percy isn't very good at unless you wait a very long time. Better to push a new dummy commit if you want to rerun the test. Or just wait until https://github.com/plotly/dash/pull/1001 is included in a release (related: https://github.com/plotly/dash-docs/pull/690) and try again at that point.

nicolaskruchten:
Hmm I think you'll need to hardcode the x and y ranges and set the transition order to be 'traces first' or something to avoid the "smooth layout transition/rough data transition" effect that we're seeing here

emmanuelle:
The percy comes from the added code + explanation. I'll let the reviewer accept the percy diff. @nicolaskruchten @alexcjohnson (and thanks for your help).

PR #696, opened 2019-11-14
Update to dash-daq==0.3.1

Marc-Andre-Rivet:
Lots of commits here (will squash) as I was trying to update all the optional component libraries at first.

Limits itself to updating to latest DAQ and fixes doc usage problems encountered along the way.


tutorial/examples/canvas_color.py
```diff
@@ -34,7 +34,7 @@
             daq.ColorPicker(
                 id='color-picker',
                 label='Brush color',
-                value='#119DFF'
+                value=dict(hex='#119DFF')
```
Marc-Andre-Rivet: Old docs used `dash-daq==0.2.1` which accepts a string. The prop in the repo history **never** accepted this type for `value`... which means the docs are running off a version of daq that was locally built and never properly committed to the repo.. <insert expletive of choice>.

tutorial/examples/daq_components/dark_theme_provider.py
```diff
@@ -24,66 +24,66 @@
         className='dark-theme-control'
     ), html.Br(),
     daq.ColorPicker(
-        value=17,
+        value=dict(hex='#171717'),
```
Marc-Andre-Rivet: Another completely broken ColorPicker value usage that works flawlessly in `dash-daq==0.2.1`

tutorial/examples/canvas_color.py
Marc-Andre-Rivet: Other less insane option is that this version packages its own version of `prop-types` which will override what the renderer provides for evaluation, for this package.

Marc-Andre-Rivet:
Haven't gone through all the diffs but so far they seem fine in the app.
https://dash-docs-pr-696.herokuapp.com/dash-daq/darkthemeprovider

Will investigate further.

PR #700, opened 2019-11-18
Fix usage for implicit pixels style strings

Marc-Andre-Rivet:
React accepts `margin: '0'` as `0px` but ignores `margin: '10'`, need to add 'px' for it to be picked up..

Searching through the docs with `['"]\w+['"]: ['"]\d+['"]`, found these examples are that can be updated. The non-zero usages were incorrect / broken.

As discussed with @alexcjohnson 

React allows `'0'` to be treated as `0` but does not allow non-zero unit-less strings to be treated as `px`. It will treat numbers as `px` though.

We will probably never want to validate/transform all props on `style_**` props in the future so that `'10'` becomes `'10px'`.

### DIFF
--- a/tutorial/dashbio.py
+++ b/tutorial/dashbio.py
@@ -98,8 +98,8 @@ DASHBIO_COMPONENTS = {
             'column_labels': 'list(df.columns.values)',
             'row_labels': 'list(df.index)',
             'hidden_labels': '[\'row\']',
-            'height': '800',
-            'width': '600'
+            'height': 800,
+            'width': 600
         },
         'component_wrap': 'dcc.Graph(figure=_[0])',
         'setup_code': '''df = pd.read_csv(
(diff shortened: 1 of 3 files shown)


tutorial/dashbio.py
```diff
@@ -98,8 +98,8 @@
             'column_labels': 'list(df.columns.values)',
             'row_labels': 'list(df.index)',
             'hidden_labels': '[\'row\']',
-            'height': '800',
-            'width': '600'
+            'height': 800,
+            'width': 600
```
Marc-Andre-Rivet: Maybe someone can validate this.. usage I saw for figures is height/width are numbers, not strings.

tutorial/dashbio.py
```diff
@@ -137,7 +137,7 @@
     'Ideogram': {
         'description': '''A visual representation and analysis tool for chromosome bands.''',
         'params': {
-            'chrHeight': '250'
+            'chrHeight': 250
```
Marc-Andre-Rivet: As per Ideogram's proptypes (number only)

tutorial/table/filtering_chapter.py
```diff
@@ -61,7 +61,7 @@
             html.Td([
                 html.H4(
                     html.P([html.Code('='), ' ', html.Code('eq')]),
-                    style={'margin': '0'}),
+                    style={'margin': '0px'}),
```
Marc-Andre-Rivet: Allowed to use `0` but just to be consistent. Using the `\d+px` format because it's more frequently used for styles in the docs.

PR #707, opened 2019-11-21
test for broken links on every page we visit

alexcjohnson:
let's see what we find!

PR #708, opened 2019-11-21
update normalized redirects

chriddyp:
Post-merge checklist:

- [ ] I understand

### DIFF
--- a/dash_docs/server.py
+++ b/dash_docs/server.py
@@ -128,3 +128,7 @@ def redirect_dcc_confirm_provider():
 def redirect_dcc_confirm():
     return redirect('/dash-core-components/confirmdialog', code=301)
 
+@server.route('/dash-core-components/faq')
+def redirect_faq():
+    return redirect('/faqs', code=301)
+
--- a/dash_docs/server.py
+++ b/dash_docs/server.py
@@ -68,63 +68,63 @@ app.config.suppress_callback_exceptions = True
 
 @server.route('/deployment/on-premise')
 def redirect_dds():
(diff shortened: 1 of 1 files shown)


chriddyp:
Based off of https://github.com/plotly/dash-docs/pull/707/files.

I don't think we need to put in a redirect for `/faqs` since that was just a broken link

chriddyp:
(rather than a valid link that changed)

alexcjohnson:
Good call, thanks @chriddyp!

Also `/dash-core-components/logout_button` -> `/dash-core-components/logoutbutton`

I wouldn't mind including `/faq` as a redirect, as that's probably what I'd type if I tried to enter this URL manually... but you're right, just a typo so not a big deal.

alexcjohnson:
BTW is it intentional to use `302` rather than `301` as the redirect code? Permanent `301` seems more semantically correct but I don't know if there are other considerations.

chriddyp:
yeah it should be a permanent one, i think that's just a typo that got repeated

chriddyp:
Huh, getting a percy issue in the tests?
```
        new_retry = self.new(
            total=total,
            connect=connect, read=read, redirect=redirect, status=status_count,
            history=history)
    
        if new_retry.is_exhausted():
>           raise MaxRetryError(_pool, url, error or ResponseError(cause))
E           urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='percy.io', port=443): Max retries exceeded with url: /api/v1/builds/ (Caused by ResponseError('too many 503 error responses'))
```

PR #709, opened 2019-11-22
fixes a couple of bugs in interactive vis

nicolaskruchten:
Post-merge checklist:

- [x] I understand

### DIFF
--- a/dash_docs/tutorial/examples/getting_started_crossfilter.py
+++ b/dash_docs/tutorial/examples/getting_started_crossfilter.py
@@ -7,7 +7,7 @@ external_stylesheets = ['https://codepen.io/chriddyp/pen/bWLwgP.css']
 
 app = dash.Dash(__name__, external_stylesheets=external_stylesheets)
 
-df = pd.read_csv('https://gist.githubusercontent.com/chriddyp/cb5392c35661370d95f300086accea51/raw/8e0768211f6b747c0db42a9ce9a0937dafcbd8b2/indicators.csv')
+df = pd.read_csv('https://plotly.github.io/datasets/country_indicators.csv')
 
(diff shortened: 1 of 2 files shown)


nicolaskruchten:
Right now the country-indicator demos show 1952/1957 in the sliders but show blank plots because the `Value` field in the CSV is null for those years, but a trailing `.dropna()` fixes that.

In https://dash.plot.ly/interactive-graphing the slider accepts years that aren't in the dataset at all right now, resulting in broken-looking interactivity, and the `step=None` fixes that.

alexcjohnson:
Something weird is going on here (which thankfully showed up in the percy diffs) - if I look at https://dash-docs-pr-709.herokuapp.com/getting-started-part-2 - specifically the "multiple inputs" graph 2/3 of the way down, the data look totally different from what they were before - and largely uncorrelated in the default view of life expectancy vs fertility rate, whereas previously there was a strong correlation.
Now:
Screen Shot 2019-11-22 at 10 11 56 AM

Previously:
Screen Shot 2019-11-22 at 10 12 49 AM

Are we doing something with the data that's incompatible with `dropna`?

nicolaskruchten:
hmm nice catch. that is really weird, i'll dig into it.

nicolaskruchten:
Well, this dataset is kind of whack:

image

nicolaskruchten:
Basically there are many combinations of year/indicator1/indicator2 which will result in an empty plot. I think we should just switch to a much smaller, fully-populated dataset.

nicolaskruchten:
OK I've created a new dataset with no indicators or years that have all-NaNs, but I've left all the NaNs in there, so nothing is out of order.

PR #710, opened 2019-11-22
new crossfilter recipe with multiple outputs

nicolaskruchten:
This PR proposes a new, streamlined and easier to understand "generic crossfilter recipe" based off of multiple outputs instead of a harder-to-understand function generator.

I've basically unrolled most of the loops to make it clearer to Python beginners how the three figures inter-relate.

--- 

Post-merge checklist:

- [x] I understand

### DIFF
--- a/dash_docs/tutorial/examples/crossfilter_recipe.py
+++ b/dash_docs/tutorial/examples/crossfilter_recipe.py
@@ -9,146 +9,91 @@ external_stylesheets = ['https://codepen.io/chriddyp/pen/bWLwgP.css']
 
 app = dash.Dash(__name__, external_stylesheets=external_stylesheets)
 
+# make a sample data frame with 6 columns
 np.random.seed(0)
-df = pd.DataFrame({
-    'Column {}'.format(i): np.random.rand(30) + i*10
-    for i in range(6)})
+df = pd.DataFrame({"Col " + str(i+1): np.random.rand(30) for i in range(6)})
 
 app.layout = html.Div([
     html.Div(
-        dcc.Graph(
-            id='g1',
-            config={'displayModeBar': False}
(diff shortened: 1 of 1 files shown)


nicolaskruchten:
Note: the output of this app is basically identical to the output of the previous recipe, so the GIF doesn't need to change :)

nicolaskruchten:
image

dash_docs/tutorial/examples/crossfilter_recipe.py
```diff
@@ -9,146 +9,91 @@
...
-
-
-
-# app.callback is a decorator which means that it takes a function
-# as its argument.
-# highlight is a function "generator": it's a function that returns function
-app.callback(
-    Output('g1', 'figure'),
+                'line': { 'width': 1, 'dash': 'dot', 'color': 'darkgrey' },
+                **selection_bounds
```
alexcjohnson: heh, never knew you could use `**` that way in a dict. I guess neither did Python2 😉

nicolaskruchten: hmmm... the old version did this? or maybe a slightly different flavour

nicolaskruchten:
Huh so in py2 `{"a":2, **b}` doesn't work but `dict({"a":2}, **b)` does.

dash_docs/tutorial/examples/crossfilter_recipe.py
alexcjohnson: looks like `dict(type='rect', line={...}, **selection_bounds)` would work, but not `**` inside `{}`

nicolaskruchten: jinx

PR #711, opened 2019-11-22
Prevent link background from coloring modebar

nicolaskruchten:
Post-merge checklist:

- [x] I understand

### DIFF
--- a/dash_docs/assets/override.css
+++ b/dash_docs/assets/override.css
@@ -150,7 +150,7 @@ h2, h3, h4, h5, h6 {
   border-radius: 1px;
 }
 
-.plotly a {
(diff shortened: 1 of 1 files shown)


dash_docs/assets/override.css
```diff
@@ -150,6 +150,10 @@ h2, h3, h4, h5, h6 {
   border-radius: 1px;
 }
 
+.plotly a {
```
alexcjohnson:
```suggestion
.js-plotly-plot .plotly a {
```

Probably doesn't matter here, but that's how we scope plotlyjs CSS internally for maximum specificity - so let's keep that for the sake of making this the best possible reference for our users.

PR #714, opened 2019-11-28
Update Dash Bio examples page. 

shammamah:
Closes https://github.com/plotly/dash-bio/issues/446. 

Post-merge checklist:

- [x] I understand

### DIFF
--- a/dash_docs/tutorial/dashbio.py
+++ b/dash_docs/tutorial/dashbio.py
@@ -18,8 +18,7 @@ DASHBIO_LIBRARY_HEADING = [
 
     Dash Bio is a suite of bioinformatics components that make it simpler to
     analyze and visualize bioinformatics data and interact with them in a Dash
-    application. You can interact with the components shown below directly, or
-    by clicking the image of the component.
+    application.
 
     The source can be found on GitHub at [plotly/dash-bio](https://github.com/plotly/dash-bio).
 
--- a/dash_docs/tutorial/dashbio.py
+++ b/dash_docs/tutorial/dashbio.py
@@ -46,9 +46,9 @@ DASHBIO_COMPONENTS = {
             'name': 'alignment_viewer_p53.fasta',
             'parameter': 'data'
         },
-        'iframe_info': {
-            'location': 'https://dash-gallery.plotly.host/docs-demos-dashbio/alignment_chart',
-            'height': 950
+        'image_info': {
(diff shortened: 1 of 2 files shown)


dash_docs/tutorial/dashbio.py
```diff
@@ -158,11 +156,11 @@
             ['dash_core_components', 'dcc']
         ],
         'setup_code': '''df = pd.read_csv(
-    'https://raw.githubusercontent.com/plotly/dash-bio' +
-    '/tests/dashbio_demos/sample_data/manhattan_data.csv'
+    'https://raw.githubusercontent.com/plotly/dash-bio/' +
+    '/master/tests/dashbio_demos/sample_data/manhattan_data.csv'
```
alexcjohnson: is that a doubled slash at the line break?

dash_docs/tutorial/dashbio.py
```diff
@@ -18,7 +18,8 @@
 
     Dash Bio is a suite of bioinformatics components that make it simpler to
     analyze and visualize bioinformatics data and interact with them in a Dash
-    application.
+    application. You can interact with the components shown below directly, or
+    by clicking the image of the component.
```
alexcjohnson: This seems a bit confusing actually - the implication of this statement, as I read it anyway, is that ALL of the components can be interacted with directly AS WELL AS by clicking on the image (to go to a separate app). Also just scrolling through it wasn't obvious which were live components vs images. I'd be happier if we could (1) word this so it's clearer that it's one or the other, and (2) give some visual indication on each displayed result as to whether it's an image or an actual live component. Perhaps at the top of each image result it can actually state something like

> Try a live demo at: http://dash-gallery.plotly.host/docs-demos-dashbio/alignment_chart

alexcjohnson:
The images have funny grey bars above and below
Screen Shot 2019-11-28 at 9 46 54 PM

dash_docs/tutorial/dashbio.py
shammamah: Good catch! I was wondering why this didn't raise any errors, but apparently you can put as many slashes as you want in a URL: https://github.com/////plotly/////dash-docs///// and it will still work. But I'll fix it anyway!

dash_docs/tutorial/dashbio.py
shammamah: Changed in 384c972!

shammamah:
>The images have funny grey bars above and below

This is due to the docs stylesheet, which gives all links a background colour. I've removed the link wrapper for the images in 384c972.

PR #726, opened 2019-12-11
Let dash docs run in offline environments

chriddyp:


chriddyp:
In the future, we should make the dash bio examples offline-friendly (cc @shammamah ) but for now I've included a flag to ignore them. Here's an issue to make the dash bio examples offline friendly: https://github.com/plotly/dash-docs/issues/727

dash_docs/chapter_index.py
```diff
@@ -372,17 +373,24 @@ def component_list(package, content_module, base_url, import_alias):
...
+                            True if
+                            os.environ.get('IGNORE_DASH_BIO', False)
+                            else False
+                        )
+                    },
+                    **({} if os.environ.get('IGNORE_DASH_BIO', False)
+                    else {'content': tutorial.dashbio.layout}))
                 ] + component_list(
                     dash_bio,
+                    None if os.environ.get('IGNORE_DASH_BIO', False) else
```
alexcjohnson: Is this a pattern we've used elsewhere? Regardless, seems weird to push the responsibility for aborting into the `component_list` function, seems like a clearer pattern would have been `+ ([] if os.environ.get(...) else component_list(...))`

.github/pull_request_template.md
```diff
@@ -10,3 +10,8 @@ If this PR documents a new feature of Dash:
 
 - [ ] Comment on the original Dash issue with a link to the new docs.
 - [ ] Reply to any community thread(s) asking for this feature.
+
+If this PR includes a new dataset available at a remote URL:
+- [ ] I have added this dataset to the `datasets/` folder
+- [ ] I have added a mapping between the remote URL and the filename in the
+`datasets/` folder
```
alexcjohnson:
```suggestion
`datasets/` folder into the `find_and_replace` dict in `dash_docs/tools.py`
```
Thanks for adding this note! Just want to make sure all the steps are spelled out, as I wasn't aware exactly how this worked previously.

dash_docs/chapter_index.py
chriddyp: Yeah, this is a little subtle but if `component_list(..., content_module=None, ...)` then the that function will autogenerate a list of components and their documentation:
https://github.com/plotly/dash-docs/blob/ec0664a513e4bef1e7f6ee9134b6a4c12af22f7f/dash_docs/chapter_index.py#L27-L34

alexcjohnson: OIC, I thought given the name `IGNORE_DASH_BIO` that it was actually skipping this part, but this makes sense.

PR #729, opened 2019-12-17
Remove old Markdown syntax

### LINKED ISSUE #694: Get rid of dedent and replace('    ', '')
Wait for #641 to merge, but we still have `textwrap.dedent` and `s.replace('<4 spaces>', '')` quite a few places (and in at least one place `s.replace('<3 spaces>', '')`, which, who knows WHAT that's doing...), should all be removed - except perhaps one or two places we're concatenating multiple strings, then it may need to remain so the concatenation is done with consistent indentation.

wbrgss:
Fixes #694 

I want to double-check Percy on this one once the PR is made

Post-merge checklist:

- [x] I understand


wbrgss:
The bulk of this was done with commands used for a similar process in a private repo, to quote:

> Commands used:
> 
> 1. `grep -rlPzo "(Markdown\(dedent\()[\s\S]*?\)" . | xargs perl -i -p0e 's/Markdown\(dedent\((.*?)\)\)/Markdown\(\1\)/sg'`
> 
> _removes `dedent` wrappers within `dcc.Markdown()` calls_
> 
> 2. `grep -rlPzo "(Markdown\(textwrap.dedent\()[\s\S]*?\)" . | xargs perl -i -p0e 's/Markdown\(textwrap.dedent\((.*?)\)\)/Markdown\(\1\)/sg'`
> 
> _removes `textwrap.dedent` wrappers within `dcc.Markdown()` calls_
> 
> 3. `grep -rl 'from textwrap import dedent' . | xargs sed -i '/from textwrap import dedent/d'`
> 
> _removes now-unnecessary `dedent` imports (`textwrap` imports are preserved as it's used elsewhere, including non-Markdown `textwrap.dedent`)_
>

dash_docs/tutorial/integrating_dash.py
```diff
@@ -47,28 +49,30 @@ def Syntax(content):
...
-        server=server,
-        routes_pathname_prefix='/dash/'
-    )
-
-    app.layout = html.Div("My Dash app")
-
-    if __name__ == '__main__':
-        app.run_server(debug=True)
-    '''
+        dedent(
```
wbrgss: @alexcjohnson Any idea why this function needs a dedent? I tried variations of adding it back in in `def Syntax`, no luck. Is there a way around it? Does it matter? 

(and throughout d1ec9ac)

alexcjohnson: Thanks for making this PR @wbrgss 😍 

The problem is the concatenation of multiple lines in the `Syntax` definition, along with the `strip` call. `dedent` looks for a common indentation on *all* lines (other than pure whitespace lines), so if we let `Markdown` do it for us, by that point the leading and trailing lines are included in the chunk that gets dedented, and since these have no indentation there will be no spaces removed. Also `strip` will remove newlines (that's what we really want) and spaces from the first line (boo) so you need to call `dedent` *before* `strip`

(and the spaces before the newlines up there I can't imagine have a purpose...)

There's no reason this file should have its own flavor of `Syntax` - it should use the one from `tutorial.components`, which should in fact be moved into `reusable_components` anyway (along with `Example`). Then this common version I believe can be fixed by changing its first line to:
```py
code = dedent(children).strip()
```

Given that the first line of a real code block will never be indented, that should be robust.

wbrgss: OK, good idea @alexcjohnson, thanks for the explanation. I moved out `Syntax` and `Example`; see https://github.com/plotly/dash-docs/compare/36f0834..1ae9ade

I didn't end up needing to modify `reusable_components.Syntax` in order to use `dedent`, at least in order to keep the visual diffs the same. There are three diffs remaining in Percy, the last two are cytoscape noise and the first one is everything shifting up from removing `from textwrap import dedent as d` as far as I can tell.

alexcjohnson:
@wbrgss this PR looks stale - I'm thinking we should copy your [comment about the commands you used](https://github.com/plotly/dash-docs/pull/729#issuecomment-566316776) into #694 and close this, to come back to it later. Do you agree or would you like to try and revive it?

dash_docs/reusable_components/Chapter.py
```diff
@@ -4,7 +4,7 @@
 from .Markdown import Markdown
 
 def s(string_block):
-    return string_block.replace('    ', '')
```
alexcjohnson: Just 🔪 `s` entirely?

alexcjohnson: same in tutorial/dash_deployment_server.py

wbrgss: :heavy_check_mark:

alexcjohnson:
Thanks for updating! Nearly there, just a few issues still pop up in the Percy diffs
- getting-started (some markdown rendering as code)
- sharing-data-between-callbacks (over-indented code)
- integrating-dash (inconsistent indentation)

PR #735, opened 2019-12-29
[WIP] attempt at putting anchor links

### LINKED ISSUE #718: Anchors for the different sections / examples of doc pages
When linking to examples in the dash-docs, it would be great to have anchors to be able to link directly to the relevant part of the doc page. I don't know what would be the best way to do this, maybe a new option of `dcc.Markdown` to generate anchors for titles?

emmanuelle:
Closes #718 (or should close, since this is only a partial solution).

This PR tries to scratch an itch: always answer "go to the FAQ and CTRL-F 'which input etc.'", in the absence of anchor links. 

So here I tried to add anchor links for the different points of the FAQ, but it does not work well yet so I'd appreciate some advice here. Indeed, if I go to http://127.0.0.1:8060/faqs#jquery, it goes to the top of the page, but if I reload the page then it goes to the right location of the anchor (using Firefox)... not sure I understand what's going on here. 

Alternatively, I was looking for ways to generate automatically anchor links from markdown section headings, but this depends on the markdown to html converter which is used by `dcc`. Where can I find information about which converter is used, and whether it's possible to configure it?

### DIFF
--- a/dash_docs/tutorial/faqs.py
+++ b/dash_docs/tutorial/faqs.py
@@ -30,7 +30,7 @@ layout = html.Div([
 
     ## Frequently Asked Questions
 
-    **Q:** *How can I customize the appearance of my Dash app?*
+    <html><a id="customize-appearance"><b>Q</b></a> <a href=#customize-appearance><em>How can I customize the appearance of my Dash app?</em></a></html>
 
     **A:** Dash apps are rendered in the browser as modern standards-compliant
       web apps. This means that you can use CSS to style your Dash app as you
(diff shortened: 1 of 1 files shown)


alexcjohnson:
@emmanuelle thanks for bringing this up - you're quite right, we really need a solution for this. I think there are a bunch of pieces to this puzzle though: https://github.com/plotly/dash-core/issues/89

chriddyp:
Sidebar anchor links were added in #811 . Some more details needed to make page scroll work on page load though...

alexcjohnson:
@emmanuelle seems like we're shooting for a more automatic solution to this issue, and got part-way there with #811 - so I'm going to close this PR, and we can sort out the final pieces elsewhere.

PR #738, opened 2020-01-06
Add advanced filtering docs for table.

shammamah:
Post-merge checklist:

- [x] I understand

### DIFF
--- a/dash_docs/tutorial/table/filtering_chapter.py
+++ b/dash_docs/tutorial/table/filtering_chapter.py
@@ -9,7 +9,7 @@ from dash_docs import reusable_components
 
 examples = {
     example: tools.load_example('tutorial/examples/table/{}'.format(example))
-    for example in ['filtering_fe.py', 'filtering_be.py']
+    for example in ['filtering_fe.py', 'filtering_be.py', 'filtering_advanced.py']
 }
 
 layout = html.Div(
@@ -138,12 +138,6 @@ layout = html.Div(
         In the future we may accept any filter strings, to allow you to
         write your own expression query language.
 
-        > Note: we're planning on adding a structured query object
-        > to make it easier and more robust to manage back-end filter logic.
-        > Follow
-        > [dash-table#169](https://github.com/plotly/dash-table/issues/169)
-        > for updates.
-
         Example:
         """)),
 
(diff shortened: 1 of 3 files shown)


dash_docs/tutorial/table/filtering_chapter.py
```diff
@@ -156,5 +150,93 @@
...
+
+        reusable_components.Markdown("---"),
+
+        reusable_components.Markdown(dedent("""
+        # Advanced filter usage
+
+        Filter queries can be as simple or as complicated as you want
+        them to be. When using filters in the column filters, they are
+        automatically converted to a filter query on that column
+        only. For instance, in the example below, typing the query `ge
```
chriddyp:
```suggestion
        only. For instance, in the example below, typing the query `>
```

chriddyp: Or, it'd be good to know if `>` gets converted into `ge` or not. Does it? 

It'd also be good to have a full list of the "semantic" operators: `gt`, `ge`, `lt`, `le`, etc. That way, members of the community can write complete parsers.

shammamah: >Or, it'd be good to know if > gets converted into ge or not. Does it?

It does get converted (to `gt`), but I agree that it would be a good idea to cover that (maybe in the "Derived filter query structure" section below). 

>It'd also be good to have a full list of the "semantic" operators: gt, ge, lt, le, etc. That way, members of the community can write complete parsers.

I think that might be already there earlier in the page, unless you meant something else: 
https://github.com/plotly/dash-docs/blob/master/dash_docs/tutorial/table/filtering_chapter.py#L97

Screen Shot 2020-01-06 at 12 05 53 PM

chriddyp: Ah OK, that's good to know. Yeah, so let's just make sure it's clear that `ge` etc are the only operators that will appear in the server side filter, never `>`, `=`, `>=`, `<`, `<=`, and `!=`

alexcjohnson: No that's not right - the operators are sent to the back end as entered, though it does normalize spacing and expand shortcuts (default operators) - for example if I enter `st` in `country` and `>10000` in `gdpPerCap` the back end sees:
```
{gdpPercap} > 10000 && {country} contains st
```
The way we did this in the example may be a bit confusing though - the main callback function only sees the letter versions, we normalized to these in the `split_filter_part` function.

shammamah: @alexcjohnson Yep you're right -- the `filter_query` prop has the same operators as whatever the user types in, but `derived_filter_query_structure` has a `subType` key that is the "symbol" version of the operator:  
Screen Shot 2020-01-06 at 2 09 55 PM

So it's the opposite of what I said (the letter-based operators are converted to the symbols, not the other way around). 

It might be a good idea to include this information in the docs, so people can make better decisions about which prop to use (`filter_query` or `derived_filter_query_structure`) when writing their parsers.

Marc-Andre-Rivet: Right. `derived_filter_query_structure` distinguishes between the operator type / subtype (symbol) and the value used to obtain it (e.g. `>` vs `gt`) while `filter_query` is just the string value of the current filter, as entered by the user or automagically via the column filters.

Marc-Andre-Rivet:
I've updated a demo code from last year that visits the derived structure in order to create a `filter+query` on a panda dataset. Maybe it would be a valuable addition, possibly with some clean up, to the extra documentation for advanced usage.

```
import dash
from dash.dependencies import Input, Output
from dash_table import DataTable
import pandas as pd

url = ("https://github.com/plotly/datasets/raw/master/"
       "26k-consumer-complaints.csv")

types = {
    'id': 'numeric',
    'Complaint_ID': 'numeric',
    'ZIP_code': 'numeric',
    'Date_received': 'datetime',
    'Date_sent_to_company': 'datetime',
}

df = pd.read_csv(url)
df['id'] = df['Unnamed: 0']
df = df.drop(['Unnamed: 0'], axis=1)
df = df.reindex(columns=['id']+df.columns[:-1].tolist())
df.columns = [column.replace(" ", "_") for column in df.columns]
df.columns = [column.replace("-", "_") for column in df.columns]

app = dash.Dash()
app.scripts.config.serve_locally = True

app.layout = DataTable(
    id='demo-table',
    data=df.to_dict('rows'),
    columns=[{ 'id': i, 'name': i, 'type': types.get(i, 'any') } for i in df.columns],
    filter_action='custom',
    page_action='none',
    virtualization=True,
    style_cell={
        'min-width': '100px'
    },
    css=[
        { 'selector': '.row-1', 'rule': 'min-height: 500px;' }
    ]
)

def to_string(filter):
    l_type = filter.get('type')
    l_sub = filter.get('subType')

    if l_type == 'relational-operator':
        if l_sub == '=':
            return '=='
        else:
            return l_sub
    elif l_type == 'logical-operator':
        if l_sub == '&&':
            return '&'
        else:
            return '|'
    elif l_type == 'expression' and l_sub == 'value' and type(filter.get('value')) == str:
        return '"{}"'.format(filter.get('value'))
    else:
        return filter.get('value')


def handle_leaf(filter, df_filter):
    return (to_string(filter), df_filter)

def handle_default(filter, df_filter):
    left = filter.get('left', None)
    right = filter.get('right', None)

    (left_query, left_df) = to_panda_filter(left, df_filter)
    (right_query, right_df) = to_panda_filter(right, left_df)

    return ('{} {} {}'.format(
        left_query,
        to_string(filter) if left_query != '' and right_query != '' else '',
        right_query
    ).strip(), right_df)


def handle_contains(filter, df_filter):
    left = filter.get('left', None)
    right = filter.get('right', None)

    (left_query, left_df) = to_panda_filter(left, df_filter)
    (right_query, right_df) = to_panda_filter(right, left_df)
    right_query = right.get('value')

    return ('', right_df[right_df[left_query].astype(str).str.contains(right_query)])


def handle_datestartswith(filter, df_filter):
    left = filter.get('left', None)
    right = filter.get('right', None)

    (left_query, left_df) = to_panda_filter(left, df_filter)
    (right_query, right_df) = to_panda_filter(right, left_df)
    right_query = right.get('value')

    return ('', right_df[right_df[left_query].astype(str).str.startswith(right_query)])


def to_panda_filter(filter, df_filter):
    if filter is None:
        return ('', df_filter)

    l_type = filter.get('type')
    l_sub = filter.get('subType')

    left = filter.get('left', None)
    right = filter.get('right', None)

    if left is None and right is None:
        return handle_leaf(filter, df_filter)
    elif l_type == 'relational-operator' and l_sub == 'contains':
        return handle_contains(filter, df_filter)
    elif l_type == 'relational-operator' and l_sub == 'datestartswith':
        return handle_datestartswith(filter, df_filter)
    else:
        return handle_default(filter, df_filter)


@app.callback(
    Output("demo-table", "data"),
    [Input("demo-table", "derived_filter_query_structure")]
)
def onFilterUpdate(filter):
    (filter, df_filter) = to_panda_filter(filter, df)
    return df_filter.query(filter).to_dict('rows') if filter != '' else df_filter.to_dict('rows')


if __name__ == "__main__":
    app.run_server(port=8080, debug=True)
```

PR #742, opened 2020-01-07
Dash Docs Tutorial suggestions

markzporter:
Post-merge checklist:

- [ ] I understand

### DIFF
--- a/dash_docs/tutorial/state.py
+++ b/dash_docs/tutorial/state.py
@@ -61,7 +61,7 @@ layout = html.Div([
         ## Using PreventUpdate in Callback
 
         In certain situations, you don't want to update the callback output. You can
-        achieve this by raising `PreventUpdate` Exception in the callback function.
+        achieve this by raising `PreventUpdate` exception in the callback function.
     '''),
     Syntax(examples['prevent-update-button'][0]),
     Example(examples['prevent-update-button'][1]),
--- a/dash_docs/tutorial/state.py
+++ b/dash_docs/tutorial/state.py
@@ -61,7 +61,7 @@ layout = html.Div([
         ## Using PreventUpdate in Callback
 
         In certain situations, you don't want to update the callback output. You can
-        achieve this by raising `PreventUpdate` exception in the callback function.
+        achieve this by raising a `PreventUpdate` exception in the callback function.
     '''),
     Syntax(examples['prevent-update-button'][0]),
     Example(examples['prevent-update-button'][1]),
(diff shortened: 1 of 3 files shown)


dash_docs/tutorial/state.py
```diff
@@ -61,7 +61,7 @@
         ## Using PreventUpdate in Callback
 
         In certain situations, you don't want to update the callback output. You can
-        achieve this by raising `PreventUpdate` Exception in the callback function.
+        achieve this by raising `PreventUpdate` exception in the callback function.
```
wbrgss:
```suggestion
        achieve this by raising a `PreventUpdate` exception in the callback function.
```

wbrgss:
@markzporter Since the Circle build re-ran and it's still failing, I think I was wrong about my estimate about what was going wrong in the first place. I suspect it has to do with a timeout in
`tests/integration/test_main_page.py:9 test_snap001_index_page_links` as per the build trace but I don't know why it works on Python 3.6 and not Python 3.7.

My next step would be to try removing your `dcc.Link` in `sharing_state.py` and see if that 3.7 test passes. If so, there may be a problem with crawling that page. The path is in `URLS`:

https://github.com/plotly/dash-docs/blob/25371c4f9c901d7baaf1a944247387d88949c980/dash_docs/chapter_index.py#L159-L166

but you need to do something like "update the search index":

https://github.com/plotly/dash-docs/blob/25371c4f9c901d7baaf1a944247387d88949c980/dash_docs/chapter_index.py#L14-L16

These are just guesses on my end :smile_cat:

markzporter:
Looks like the problem was a stray comma after the `dcc.Link` declaration. Seems to work fine now!

Not sure why the minor version change would impact this. Only reference in the 3.7 spec I can find to a comma is this: 

Screen Shot 2020-01-09 at 2 24 15 PM

[Source](https://docs.python.org/3/whatsnew/3.7.html)

🤷‍♂

wbrgss:
@markzporter hmm, very interesting! Nice find. I normally don't add trailing commas for the last item in a Python lists but I see them everywhere, especially in `children[]` lists. I always thought they were innocuous and figured that people left them in with the expectation of adding more items in the future. AFAIK they're allowed in Python syntax but maybe in this case it's interfering with the crawler.

wbrgss:
@markzporter We'll still need one of the four "code owners" of Dash Docs to approve this PR before you can merge it merged :smile: — unless you'd like me to merge it myself as an admin.

markzporter:
@wbrgss Thanks for my first approval 🤠 

As for getting it merged there's no rush I guess! Always good to have another set of eyes 👀.

Issue #740, opened 2020-01-07
Longer chapter on interactive graphs


emmanuelle:
We've already mentioned a few times that the part on plotly graphs in Dash applications should be expanded. I'm trying here to summarize what could go into a new version:
- the different ways to set `dcc.Graph.figure`: `dict` or plotly figure, which can be created with `plotly.graph_objects` or `plotly.express` (possibly using `fig.update_traces` and `fig.update_layout`).
- **existing** example capturing user interaction (`clickData` etc.)
- document `extendData`  and `restyleData` (#488)
- **existing** example Update Graphs on Hover
- example using uirevision
- document transitions (see #660 )
- document graph responsiveness (see #720)
- example of editable (`config={'editable":True}) figure triggering a callback when edited
- mention clientside callbacks for graphs

Does that make sense? Anything else missing? It's a lot of work but I think it makes sense to consider this as a whole. A lot of questions on the forum would probably disappear with the above additions.

chriddyp:
Since https://github.com/plotly/dash-docs/pull/641, each component now has a dedicated page. These pages aren't directly linked to in `dash.plot.ly` but they are linked to in other environments like Dash Enterprise that have a sidebar:

![image](https://user-images.githubusercontent.com/1280389/72187287-4ce6f600-33c5-11ea-8442-d093e06ea720.png)

We'll be adding a sidebar in dash.plot.ly next quarter as well.

So, could we do a quick first pass at this by just adding a preamble to this page with:
1. A link to plot.ly/python
2. A couple examples (it could be syntax only) showing what `figure` is. 
3. A link to https://plot.ly/python/creating-and-updating-figures/ for the exhaustive list
4. A link to the interactive-graphing chapter
5. A link to whatever community posts we have uirevision & transitions

5 is optional, but 1-4 should be pretty quick and would add some intermediate value.

emmanuelle:
ok; so I should add this to the docstring of `dcc.Graph`, ie here https://github.com/plotly/dash-core-components/blob/dev/src/components/Graph.react.js#L15. Correct?

chriddyp:
You can actually add it in this file, to a variable named `Graph` https://github.com/plotly/dash-docs/blob/master/dash_docs/tutorial/core_component_examples.py that you "export" in this dictionary: https://github.com/plotly/dash-docs/blob/master/dash_docs/tutorial/core_component_examples.py#L12-L42

[...]

PR #745, opened 2020-01-15
Added plotly paragraph to the documentation page of dcc.Graph

emmanuelle:
See #740. At first I had added a couple of examples, but they will not run since we do not have plotly in the doc dependencies, so we'll have to keep this for another time.

### DIFF
--- a/dash_docs/tutorial/core_component_examples.py
+++ b/dash_docs/tutorial/core_component_examples.py
@@ -215,12 +215,45 @@ Graph = html.Div(children=[
         or inline directly in your environment). To see all of these rendering
         environments, see https://plot.ly/python/renderers/.
     """),
+    html.H2('Examples'),
+    html.P("Simple graph defined from a dictionary with the same structure as a plotly figure"),
+    ComponentBlock('''import dash_core_components as dcc
+dcc.Graph(
+        id='example-graph',
+        figure={
+            'data': [
+                {'x': [1, 2, 3], 'y': [4, 1, 2], 'type': 'bar', 'name': 'SF'},
+                {'x': [1, 2, 3], 'y': [2, 4, 5], 'type': 'bar', 'name': u'Montréal'},
+            ],
+            'layout': {
+                'title': 'Dash Data Visualization'
+            }
+        }
+    )
+''', style=styles.code_container),
+    html.P("Graph defined directly from a plotly figure object"),
+    ComponentBlock('''import dash_core_components as dcc
+import plotly.graph_objs as go
+fig = go.Figure(data=[go.Scatter(x=[1, 2, 3], y=[4, 1, 2])])
+dcc.Graph(
+        id='example-graph-2',
+        figure=fig
+    )
+''', style=styles.code_container),
+
 
     html.H2('Interactive Graphing'),
     reusable_components.Markdown("""
     The [Interactive Visualizations](/interactive-graphing) tutorial explains how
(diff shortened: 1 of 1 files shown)


emmanuelle:
Also, I don't understand where the `Graphs` paragraph which was already present in this file is used...

chriddyp:
> Also, I don't understand where the Graphs paragraph which was already present in this file is used...

I think it was unused. The `component_list` function looks for exact matches of variable names and component names:
https://github.com/plotly/dash-docs/blob/f259003d7429d3e60f4574970eca4b6a2f1513ba/dash_docs/chapter_index.py#L19-L40

in particular, here: 
https://github.com/plotly/dash-docs/blob/f259003d7429d3e60f4574970eca4b6a2f1513ba/dash_docs/chapter_index.py#L24-L27

So, if it was renamed to `Graph` it probably would've been picked up

chriddyp:
> but they will not run since we do not have plotly in the doc dependencies, so we'll have to keep this for another time

That's odd, they _should_ run, `plotly` is a dependency of `dash` so it's installed automatically. And our other doc pages (https://dash.plot.ly/getting-started-part-2) display plotly graphs

chriddyp:
obviously tons more stuff we can add here, but 💃 for now!

if you feel up for it, some other quick wins would be:
- a couple simple examples
- some links to dash-specific dash features, even if it's just community forum links for now: uirevision (https://community.plot.ly/t/preserving-ui-state-like-zoom-in-dcc-graph-with-uirevision/15793), transitions (https://community.plot.ly/t/exploring-a-transitions-api-for-dcc-graph/15468)

emmanuelle:
OK, I made the changes (examples + links to forum posts). For plotly I had misinterpreted the error message; it's used indeed but with version pinned at 3.3 and I was using v4 syntax. So I've kept examples minimal for now (the syntax still runs with v4 but is quite outdated, now we import from `graph_objects` for example). But I think it's enough to explain that you can use plotly figure objects. If you don't mind giving it another look @chriddyp  to be sure that I can press the green button.

chriddyp:
Ah OK, got it. I think that a upgrade to v4 would be fairly straightforward (famous last words!) but 💃  for now. We could try a upgrade to v4 and a simple `px` example in a follow up PR if we have time

dash_docs/tutorial/core_component_examples.py
```diff
@@ -174,6 +174,91 @@
...
+    reusable_components.Markdown("""
+    The [Interactive Visualizations](/interactive-graphing) tutorial explains how
+    to capture user interaction events with a `dcc.Graph`, and how to update the
+    `figure` property in callbacks.
+
+    Some advanced features are documented in [community forum](https://community.plot.ly/) posts:
+
+    - How to preserve the UI state (zoom level etc.) of a Graph when updating the Graph in a callback
+    https://community.plot.ly/t/preserving-ui-state-like-zoom-in-dcc-graph-with-uirevision/15793
+    - Graph transitions for smooth transitions on Graph updates
```
chriddyp:
```suggestion
    - Graph transitions for smooth transitions or animations on Graph updates
```

emmanuelle:
Yes @nicolaskruchten thought that it was better to upgrade the doc to v4 when we are not in a hurry. I can always make another branch to see how it goes but I would be cautious :-).

emmanuelle:
I think I need an approving review here before merging.

PR #767, opened 2020-01-29
Clean up frontend filtering example

Marc-Andre-Rivet:


dash_docs/tutorial/examples/table/filtering_fe.py
```diff
@@ -11,29 +11,13 @@
...
-def update_graph(rows):
-    if rows is None:
-        dff = df
-    else:
-        dff = pd.DataFrame(rows)
-
-    return html.Div()
+app.layout = dash_table.DataTable(
+    columns=[
+        {"name": i, "id": i, "deletable": True} for i in df.columns
```
chriddyp:
```suggestion
        {"name": i, "id": i} for i in df.columns
```

chriddyp: Just to make this example as "bare bones" as possible. Otherwise, folks might think that filtering only works if `deletable` is set to `True`

PR #770, opened 2020-02-03
modified source of image in canvas docs

emmanuelle:
Some examples of the dash-canvas docs are currently broken because wikipedia must have changed something with the way images can be downloaded; I added instead the image to the plotly datasets repo so that the problem does not happen again, and modified accordingly the links in the docs (this PR).


emmanuelle:
The percy snapshots display a "cannot render html5 canvas" error, which kind of browser are they using? It works fine with both Firefox and Chromium on my machine...

PR #783, opened 2020-02-17
Offline assets

wbrgss:
Making the docs compatible offline, ideally with zero (0) network requests.

TODO:

~[ ] find an alternative to the calls to https://ghbtns.com/ for the "Star" buttons. We could just  replace the iframes with a static image and a link to the repo. We could also introduce logic for checking if the machine is offline but...that's going to make a network request and kind of make the point moot~

~[ ] Find an alternative to the embedded Codepen request on the "Multi Page Apps" chapter (`dash_docs/tutorial/urls.py``)~

<hr><hr>

Post-merge checklist:

- [x] I understand

- [ ] I used e.g. `<dccLink href="/getting-started" children="the first chapter"/>` instead of `[the first chapter](/getting-started)`.


chriddyp:
💃 

could you also update the `PR_CHECKLIST` to include a checklist item about "If I added an image, this image was saved and referenced locally rather than via an external link"

wbrgss:
Thanks for taking a look @chriddyp. Good idea re: the PR template; I think this should also be replaced because we now recommend using `reusable_components.Markdown` which uses `tools.relpath` internally (in this PR I broke out `html.Img()`s because I was ignorant of this:

>  I used e.g. <dccLink href="/getting-started" children="the first chapter"/> instead of [the first chapter](/getting-started).

Note that @markzporter will be making a similar PR (we split up this task)

wbrgss:
Also, @chriddyp (or @Marc-Andre-Rivet) do you have an opinion on the external requests above in my `TODO` (edited opening comment)? These external requests will not break the app by any means but still may trigger a network request audit/script for airgapped environments.

chriddyp:
I think it's fine if a request is triggered as long as it doesn't break the app.

Marc-Andre-Rivet:
image
Looks like flakiness - py36 and 37 passed, rerunning.

Marc-Andre-Rivet:
There's many requests for `https://codepen.io/chriddyp/pen/bWLwgP.css` in the docs, do we care about those?

wbrgss:
> There's many requests for https://codepen.io/chriddyp/pen/bWLwgP.css in the docs, do we care about those?

The docs look OK without wifi, presumably without this CSS. So I say it's OK given https://github.com/plotly/dash-docs/pull/783#issuecomment-587122275. No strong opinions though.

PR #786, opened 2020-02-19
prevent double relative paths

Marc-Andre-Rivet:
With some usages the path can be doubly relative. I'm not sure exactly what chain of calls makes the path doubly relative, this PR simply prevents it from happening.

### DIFF
--- a/dash_docs/tools.py
+++ b/dash_docs/tools.py
@@ -8,9 +8,7 @@ else:
 
 
 def relpath(path):
-    if path.startswith('/') and
-        'DASH_DOCS_URL_PREFIX' in os.environ and
-        not path.startswith('/{}/'.format(os.environ['DASH_DOCS_URL_PREFIX'].rstrip('/'))):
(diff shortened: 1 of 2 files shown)


dash_docs/tools.py
```diff
@@ -8,7 +8,7 @@
 
 
 def relpath(path):
-    if path.startswith('/') and 'DASH_DOCS_URL_PREFIX' in os.environ:
+    if path.startswith('/') and 'DASH_DOCS_URL_PREFIX' in os.environ and not path.startswith('/{}/'.format(os.environ['DASH_DOCS_URL_PREFIX'].rstrip('/'))):
```
Marc-Andre-Rivet: prevent double relpath

alexcjohnson: If we need to get something out quickly for the release I’m OK with this. But longer-term I’d like to figure out where we’re doubly-prefixing and prevent it - maybe just turn this test into an error, or at least log the stack trace when it happens. As it stands we're turning an existing error into a potential very confusing error down the line, if we ever have a `/Docs/` path within this app that should turn into `/Docs/Docs/`, or some other similar usage of `DASH_DOCS_URL_PREFIX`.

Marc-Andre-Rivet: Absolutely. Here's the follow up for a more systematic fix https://github.com/plotly/dash-docs/issues/787

Issue #771, opened 2020-02-04
Document inline callbacks


Marc-Andre-Rivet:
Implemented in https://github.com/plotly/dash/pull/967

PR #796, opened 2020-02-26
Add documentation of inline clientside callbacks.

shammamah:
### About
Closes https://github.com/plotly/dash-docs/issues/771

Post-merge checklist:

- [x] I understand

- [ ] I considered whether I could replace the `dcc.Markdown` call with `reusable_components.Markdown`, which will replace relative links with `tools.relpath` internally. Otherwise, I used e.g. `<dccLink href=tools.relpath('/getting-started') children="the first chapter"/>` instead of `[the first chapter](/getting-started)` (importing `tools` from `dash_docs`), and set `dangerously_allow_html=true` in the `dcc.Markdown` call.

### DIFF
--- a/dash_docs/tutorial/performance.py
+++ b/dash_docs/tutorial/performance.py
@@ -179,6 +179,28 @@ app.clientside_callback(
     '''),
 
     reusable_components.Markdown('''
+#### Inline Clientside Callbacks
+
+You can also write the JavaScript code directly in your Dash app by
+supplying a value to the "source" argument. Here's an example using
+the same function as above:
+    '''),
+
+    Syntax('''
+from dash.dependencies import Input, Output
+
+app.clientside_callback(
+    Output('out-component', 'value'),
+    [Input('in-component1', 'value'), Input('in-component2', 'value')],
+    source="""
+    function(largeValue1, largeValue2) {
+        return someTransform(largeValue1, largeValue2);
+    }
+    """
+)
+    '''),
+
+    reusable_components.Markdown('''
 
 ***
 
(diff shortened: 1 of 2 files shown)


dash_docs/tutorial/performance.py
```diff
@@ -178,6 +178,28 @@ def large_params_function(largeValue1, largeValue2):
...
+
+app.clientside_callback(
+    Output('out-component', 'value'),
+    [Input('in-component1', 'value'), Input('in-component2', 'value')],
+    source="""
+    function(largeValue1, largeValue2) {
+        return someTransform(largeValue1, largeValue2);
+    }
+    """
+)
```
alexcjohnson: Doesn't the function need to be first? Per the docstring:

https://github.com/plotly/dash/blob/38994ebca24775d6b588051bef6a54699c1eb1ff/dash/dash.py#L1249-L1261

```suggestion
app.clientside_callback(
    """
    function(largeValue1, largeValue2) {
        return someTransform(largeValue1, largeValue2);
    }
    """,
    Output('out-component', 'value'),
    [Input('in-component1', 'value'), Input('in-component2', 'value')]
)
```

dash_docs/tutorial/performance.py
```diff
@@ -178,6 +178,28 @@ def large_params_function(largeValue1, largeValue2):
 )
     '''),
 
+    reusable_components.Markdown('''
+#### Inline Clientside Callbacks
+
+You can also write the JavaScript code directly in your Dash app by
```
alexcjohnson: Personally I'd vote to describe the inline version first, as the preferred way to create clientside callbacks, and describe the original way with a separate script in `assets` as the alternative. @chriddyp do you agree?

chriddyp: yeah i agree

PR #798, opened 2020-02-29
resolve generate_table_example warning

YaCpotato:
Post-merge checklist:

- [x] I understand

- [ ] I considered whether I could replace the `dcc.Markdown` call with `reusable_components.Markdown`, which will replace relative links with `tools.relpath` internally. Otherwise, I used e.g. `<dccLink href=tools.relpath('/getting-started') children="the first chapter"/>` instead of `[the first chapter](/getting-started)` (importing `tools` from `dash_docs`), and set `dangerously_allow_html=true` in the `dcc.Markdown` call.

### DIFF
--- a/dash_docs/tutorial/examples/getting_started_table.py
+++ b/dash_docs/tutorial/examples/getting_started_table.py
@@ -8,11 +8,15 @@ df = pd.read_csv('https://gist.githubusercontent.com/chriddyp/c78bf172206ce24f77
 
 def generate_table(dataframe, max_rows=10):
     return html.Table([
+
+        # Header
         html.Thead(
             html.Tr([
                 html.Th(col) for col in dataframe.columns
             ])
         ),
+
+        # Body
         html.Tbody([
             html.Tr([
                 html.Td(dataframe.iloc[i][col]) for col in dataframe.columns
--- a/dash_docs/tutorial/examples/getting_started_table.py
+++ b/dash_docs/tutorial/examples/getting_started_table.py
@@ -7,21 +7,18 @@ df = pd.read_csv('https://gist.githubusercontent.com/chriddyp/c78bf172206ce24f77
 
 
 def generate_table(dataframe, max_rows=10):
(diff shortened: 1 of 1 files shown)


dash_docs/tutorial/examples/getting_started_table.py
```diff
@@ -9,12 +9,18 @@
...
+            html.Tr([
+                html.Th(col) for col in dataframe.columns
+            ])
+        ])] +
 
         # Body
-        [html.Tr([
-            html.Td(dataframe.iloc[i][col]) for col in dataframe.columns
-        ]) for i in range(min(len(dataframe), max_rows))]
+        [html.Tbody(
```
alexcjohnson: Thanks for the fix @YaCpotato! That warning is certainly annoying.

Just a small note - the syntax can be simplified: we don't need to add these outer lists anymore since they both just have one element, and the `Tr` in `Thead` needn't be in a list since it's the only child. Also the comments `# Header` and `# Body` are kind of moot now that we have `Thead` and `Tbody` elements 😅 So it can just be: 

```py
html.Table([
    html.Thead(
        html.Tr([
            html.Th(col) for col in dataframe.columns
        ])
    ),
    html.Tbody([
        html.Tr([
            html.Td(dataframe.iloc[i][col]) for col in dataframe.columns
        ]) for i in range(len(dataframe))
    ])
])

YaCpotato: "#Header" and "#Body" are syntax annotation? that's awesome. I didn't notice.  
I understand your notes.

alexcjohnson: No, they’re just comments. But now with your fix, the code is self-commenting!

YaCpotato: I pushed code.↓
 "your small note" and "apply #Header #Body" comments.

YaCpotato:
You’re welcome.

PR #807, opened 2020-03-09
transition to plotly 4.x

emmanuelle:
This PR:
- migrates the version of plotly.py to a version of the 4.x series. This will allow to add more examples with `px` etc. (discussed in #740).
- uses the recommended syntax with `graph_objects` instead of `graph_objs`.

To discuss: which template should be used in dash-docs?


emmanuelle:
The CI failure looks unrelated to this PR but I might me missing something....

chriddyp:
💃 once tests pass

chriddyp:
4.x was updated in #811 . still using legacy `graph_objs` though

PR #811, opened 2020-03-10
Rebrand, Search, Anchor Links, and Python 3

chriddyp:


chriddyp:
This is ready for an initial review @alexcjohnson @Marc-Andre-Rivet @rpkyle . Still lots of improvements to make to the design here, but I'd like to get this PR in and follow up with additional PRs before this one gets any bigger.

Best to review this PR commit-by-commit since there are many new files.

Outstanding issues for this PR:
- [ ] Get Heroku builds to work
- [ ] Test in Safari, IE11, Edge, Chrome
- [ ] Test on Mobile. Hide sidebar on Mobile
- [ ] Click through the pages / screenshots and make sure there aren't any big style bugs
- [ ] Remove search icon from header and the associated test

Immediate follow up PRs:
- [ ] Add more links to the header
- [ ] Finalize the styles: link hover styles, border colors
- [ ] Add a footer?
- [ ] Sync up the styles with Dash for R

chriddyp:
@Marc-Andre-Rivet - for this IE11 error " Getting a "Undefined variable in strict mode", object doesn't support forEach", do I need some IE11 polyfill? I would've expected this to be included in the dash-components-boilerplate, but maybe i'm missing something

dash-user-guide-components/src/lib/components/Sidebar.react.js
```diff
@@ -237,12 +238,16 @@ function link(chapter) {
...
-            {chapter.name}
-        </a>
-    );
+    const linkProps = {
+        href: chapter.url,
+        title: title,
+        className: `${active ? 'active': ''}`,
+        children: chapter.name
+    };
+    if (chapter.url.startsWith('http')) {
```
alexcjohnson: `startsWith` is not supported by IE, and I'm not sure if we polyfill string methods

dash-user-guide-components/dash_user_guide_components/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+        'relative_package_path': 'dash_user_guide_components.min.js',
+        'external_url': 'https://unpkg.com/{0}@{2}/{1}/{1}.min.js'.format(
+            package_name, __name__, __version__),
+        'namespace': package_name
+    },
+    {
+        'relative_package_path': 'dash_user_guide_components.min.js.map',
+        'external_url': 'https://unpkg.com/{0}@{2}/{1}/{1}.min.js.map'.format(
+            package_name, __name__, __version__),
+        'namespace': package_name
```
alexcjohnson:
```
dash_user_guide_components.v0_0_1m1584034579.min.js.map:1
Uncaught SyntaxError: Unexpected token ':'
```
Add `'dynamic': True` (which has recently been fixed in the boilerplate)

dash-user-guide-components/src/lib/components/PageMenu.react.js
```diff
@@ -0,0 +1,86 @@
...
+                parent.className === window.location.pathname
+            ) {
+                window.clearInterval(renderPageMenuLinks);
+                return;
+            }
+            const elements = document.querySelectorAll('h1, h2, h3, h4, h5, h6');
+            const links = [];
+            elements.forEach(el => {
+
+                if (!el.href) {
```
alexcjohnson:
```suggestion
                if (!el.id) {
```
or am I missing something?

dash-user-guide-components/src/lib/components/PageMenu.react.js
```diff
@@ -0,0 +1,86 @@
...
+    componentWillUnmount() {
+        this._clearOnLocationChanged();
+    }
+
+    render() {
+        /*
+         * Display links directly via setInterval because we don't know when the
+         * headers will be rendered in the DOM
+         */
+        function renderPageMenuLinks() {
```
alexcjohnson: this is a strange pattern, and moreover doesn't seem to work correctly ATM. I'm seeing this menu consistently and persistently behind, showing the items on the previous page, as I click around on the main sidebar.

A totally separate approach that may be more idiomatic Dash: give this component a prop like `headerList` and make a clientside callback taking the page layout as an input and reducing it to a list of `innerText` values. Then the `onClick` handler would (1) find the element and add the correct ID to it, (2) scroll to that ID.

If you don't want to make that big a change, can we at least switch to a `state` based system? Have an interval running constantly that directly checks the text of all `document.querySelectorAll('h1, h2, h3, h4, h5, h6')` elements and stores it in `state` whenever this changes, triggering a regular rerender? Shouldn't be that big an overhead to just leave that running constantly, wouldn't you think?

dash-user-guide-components/src/lib/components/PageMenu.react.js
```diff
@@ -0,0 +1,86 @@
...
+            if(parent.innerText !== '' &&
+                // When the single page app location changes, update the page menu
+                parent.className === window.location.pathname
+            ) {
+                window.clearInterval(renderPageMenuLinks);
+                return;
+            }
+            const elements = document.querySelectorAll('h1, h2, h3, h4, h5, h6');
+            const links = [];
+            elements.forEach(el => {
```
alexcjohnson: This may be the IE error FWIW - a `NodeList` or whatever doesn't necessarily support array methods, but it has a `length` so you can do a for loop.

dash-user-guide-components/src/lib/components/Sidebar.react.js
chriddyp: Thanks! Fixed, we're getting closer to IE11 support...

dash-user-guide-components/dash_user_guide_components/__init__.py
chriddyp: fixed. do i need it in `.min.js` too or just `.min.js.map`?

dash-user-guide-components/src/lib/components/PageMenu.react.js
chriddyp: fixed 👍

dash-user-guide-components/src/lib/components/PageMenu.react.js
chriddyp: Yeah this pattern is gross. Took the clientside callback approach to trigger rerendering but am still crawling the DOM. Seems to work well now: https://github.com/plotly/dash-docs/pull/811/commits/6737e715d48108bf55623bc760f3a29a6c7c993d

dash-user-guide-components/src/lib/components/PageMenu.react.js
chriddyp: 👍 great thanks! Still a rogue IE11 error out there, but I fixed this

dash-user-guide-components/dash_user_guide_components/__init__.py
alexcjohnson: Just the `.map` - without that it'll be loaded in a `script` tag, which leads to that error.

rpkyle:
I wonder if we could consider modifying the Tutorial breadcrumbs so that 

*Part 2. Layout* maps to "Table of Contents > Layout" instead of "Table of Contents > Getting Started"
*Part 3. Basic Callbacks* maps to "Table of Contents > Basic Callbacks" instead of "Table of Contents > Getting Started Part 2"

since we'd then have consistent page shortcut naming. Given the new layout with the (helpful) breadcrumbs, I feel like "Getting Started" makes a little less sense now.

rpkyle:
It's pretty minor, but the breadcrumb names for the components have odd capitalization, like this one for `dcc.ConfirmDialogProvider`:

image

Instead of `Confirmdialogprovider`, should we use Pascal case here too?

rpkyle:
In the `dash-core-components` section, the "Interactive Tables" and "Upload Component" examples aren't rendering for me. In addition, when you click on the tables example, it links against https://github.com/plotly/dash-table-experiments instead of https://github.com/plotly/dash-table, which I think we want instead.

chriddyp:
thank you for the thorough review @rpkyle ! you found many bugs that were even unrelated to these changes 💯 

I have addressed all of your comments. Once tests pass, can I get a 💃 ?

chriddyp:
Alright @rpkyle - I've fixed IE11 and I'm very close to fixing the last test failure. I won't merge until Python 2 test is fixed, but can we 💃 ?

rpkyle:
> Alright @rpkyle - I've fixed IE11 and I'm very close to fixing the last test failure. I won't merge until Python 2 test is fixed - Assuming that I'll fix it, can we 💃 ?

I just checked out the deployed draft docs, looks great! 🚀 

 💃:books:🕺

alexcjohnson:
@chriddyp not sure if you want to hold up for this or address it later, but something in this PR broke `daq.Slider` - probably just a CSS collision, but the pieces jumped to weird places on the page (target jumped up to the `BooleanSwitch` section?) and when you drag the handle it moves way too far and fast.
Screen Shot 2020-03-17 at 9 01 04 PM

`dcc.Slider` is fine.

chriddyp:
Good catch! Looks like we were on an old version of dash-daq... I'll try updating that first.

chriddyp:
⛵️ ⛵️ ⛵️

PR #816, opened 2020-03-18
simplify requirements.txt

alexcjohnson:
Looks like a pip freeze made it in there as part of #811

Simplifying so we don't have to deal with too many extra bumps like #814

Post-merge checklist:

- [x] I understand


chriddyp:
💃 once tests pass

Issue #821, opened 2020-03-23
Dash for R documentation requires pagemenu for parity with Python docs


rpkyle:
The recent redesign of the Dash for Python documentation offers major improvements in search and navigation, including a handy pagemenu at the right of each page:

![image](https://user-images.githubusercontent.com/9809798/77323219-48a35700-6ceb-11ea-9290-5e6d2c502ed9.png)

We should provide support for this feature in the online R documentation as well, which will require translating this callback:

https://github.com/plotly/dash-docs/blob/44df6823ab7000baee366d0f72611f0c98844bfa/dash_docs/run.py#L216-L259

PR #823, opened 2020-03-24
Add pagemenu to right side of Dash for R documentation

rpkyle:
This PR proposes to address #821 by
- loading the `dashUserGuideComponents` package for R
- adding the Pagemenu to the right hand side of the R docs pages
- introducing a draft of some of the callback logic to support backlinks and left hand side page navigation (currently disabled and untested); additional elements will be introduced in a later PR

Closes plotly/dash-docs#821.


run.R
```diff
@@ -228,38 +229,75 @@ header <- htmlDiv(
...
+#               htmlDiv(id='chapter', className='content'),
+#               className='content-container'
+#             )
+#           ),
+#           pageMenu(id='pagemenu')
+#         )
+#       )
+#     )
+#   )
+# )
```
HammadTheOne: May want to remove this commented out codeblock before merging.

run.R
```diff
@@ -228,38 +229,75 @@ header <- htmlDiv(
...
+        htmlDiv(id='backlinks-bottom', className='backlinks')
+        ),
+        className='rhs-content container-width'),
+            
+      pageMenu(id='pagemenu')
+      
+      ))
   )
 )
+)
```
HammadTheOne: A little bit of autoformat with RStudio on this code-block to clean it up:

```
app$layout(htmlDiv(
  list(
    # Stores used by examples.
    dccStore(id = 'memory'),
    dccStore(id = 'memory-output'),
    dccStore(id = 'local', storage_type = 'local'),
    dccStore(id = 'session', storage_type = 'session'),
    
    # div used in tests
    htmlDiv(id = 'wait-for-layout'),
    
    dccLocation(id = 'url', refresh = FALSE),
    
    header,
    
    htmlDiv(
      className = 'content-wrapper',
      children = list(
        #Sidebar(urls=SIDEBAR_INDEX),
        
        htmlDiv(list(
          htmlDiv(id = 'backlinks-top', className = 'backlinks'),
          htmlDiv(htmlDiv(id = 'chapter', className = 'content'),
                  className = 'content-container'),
          htmlDiv(id = 'backlinks-bottom', className = 'backlinks')
        ),
        className = 'rhs-content container-width'),
        
        pageMenu(id = 'pagemenu')
        
      )
    )
  )
))
```

run.R
```diff
@@ -587,4 +625,73 @@ app$callback(
   }
 )
 
+app$callback(
```
HammadTheOne: Same as above, may want to add commented codeblocks to the scope of this PR or adjust it before merging.

run.R
rpkyle: Yeah, I'll 🔪 these but deliberately commented them out so we could populate/refactor as needed once the other docs changes are done. I'll keep the branch alive after merging until that happens, then delete it.

run.R
rpkyle: Thanks! 👔

rpkyle: fixed in 063d99c2

run.R
rpkyle: Fixed in 063d99c2

dash_docs/assets/pagemenu.js
```diff
@@ -0,0 +1,7 @@
+if(!window.dash_clientside) {window.dash_clientside = {};}
+window.dash_clientside.clientside = {
+  pagemenu: function (children) {
+    console.warn('updating pagemenu');
+    return '';
```
alexcjohnson: Hmm I see this is just copied from the Python side, but the wildcards branch includes an optimization that if a callback returns the same value a prop already had, the update is skipped. I think that'll cause problems here, so better to do something variable like `return String(Date.now())`

cc @chriddyp - do you see any problems with preventing updates on unchanged props, other than needing to modify occasional hacks like this?

rpkyle: Ah, good to know. I'll make this change. Thanks @alexcjohnson!

PR #826, opened 2020-03-25
Refactor Codebase

chriddyp:
Main objective: get the Python & R files to live within the same folders!

Fixed some other things while I was in the neighborhood:
- Various content & example fixes
- New `tools.load_examples` to automatically load the set of examples in any chapter’s `examples/` folder
- A few CSS fixes

There are still some folder differences between Python and R for less-commonly-updated chapters. I ran out of steam. I figured we can address these as we edit the files.

Notable changes:
- dash-core-components has been broken up into multiple folders and files 😌 https://github.com/plotly/dash-docs/tree/nights-since-september/dash_docs/chapters/dash_core_components. Previously, this was one huge 1.6k file (https://github.com/plotly/dash-docs/blob/master/dash_docs/tutorial/core_component_examples.py)
- I didn't do this with all of the folders - daq and bio are still monolithic files... However the R structure is more-or-less what we want so we can follow up with the Python side whenever we end up wanting to edit those files

See the structure here: https://github.com/plotly/dash-docs/tree/nights-since-september/dash_docs/chapters

Dash for R review app here: https://dash-docs-pr-826.herokuapp.com/


chriddyp:
Alright, this is READY FOR REVIEW! 🎉  

I've updated the top comment with some details: https://github.com/plotly/dash-docs/pull/826#issue-393418770

Percy should tell us everything we need for the Python side. The R side will be more difficult, we'll have to visually inspect from the review app... @rpkyle or @HammadTheOne I could use your help here.

I may revert https://github.com/plotly/dash-docs/pull/729 and apply those changes manually to this branch as there are a ton of conflicts 😬 . @wbrgss - would I get most of the changes if I applied the commands you wrote in https://github.com/plotly/dash-docs/pull/729#issuecomment-566316776? Or were there lots of manual changes?

chriddyp:
Alright, I just visually inspected all of the Percy diffs and fixed most of the issues. I left the percy diffs unapproved just so that the reviewers can see which content changed. Percy diffs are the easiest way to review this PR 😸

chriddyp:
Re merge conflicts - looks like about 104 conflicts:
image


So, I'd like to attempt to just re-run the scripts in #729 rather than fixing all of these by hand.

alexcjohnson:
Looking great! Just a few things I notice from the percy diffs:
- https://dash-docs-pr-826.herokuapp.com/integrating-dash and https://dash-docs-pr-826.herokuapp.com/sharing-data-between-callbacks -> extra indentation, these were problematic at one point in #729, perhaps a merge issue?
- Also the first example on https://dash-docs-pr-826.herokuapp.com/clientside-callbacks is over-indented
- Not new here, but I notice a few places the text says "a Dash Enterprise" - probably a find/replace from "a Dash Deployment Server", but the "a" should go away.
- https://dash-docs-pr-826.herokuapp.com/layout - the output container for examples used to scroll horizontally, now the content overflows. (In percy this also means graphs overflow vertically, though that's not an issue on the real pages.)
Screen Shot 2020-03-30 at 7 51 50 AM

- In many of the Cytoscape pages, there's a space before `cyto.Cytoscape`
Screen Shot 2020-03-30 at 8 01 12 AM

- https://dash-docs-pr-826.herokuapp.com/clientside-callbacks Good catch putting `reusable_components.Markdown` back to `dcc.Markdown` but the example still says `rc.Markdown`
Screen Shot 2020-03-30 at 8 08 05 AM

chriddyp:
Thanks @alexcjohnson ! Those issues and the conflicts have been fixed 👍 

Just looking for your review on the Dash for R stuff now @rpkyle

rpkyle:
> Thanks @alexcjohnson ! Those issues and the conflicts have been fixed 👍
> 
> Just looking for your review on the Dash for R stuff now @rpkyle

🕺Sounds good! I'll take a look and get back to you momentarily.

alexcjohnson:
@chriddyp One issue from my review remains: the extra space ahead of `cyto.cytoscape`. Everything else looks good from my end!

chriddyp:
Found it! https://github.com/plotly/dash-docs/pull/826/commits/184126d9cb66745db0ccabb8eab8fedb4564eada

chriddyp:
Alright, we're very close. However, the R docs are serving an older version `base.css` than the Python docs, perhaps the caching isn't working right?

Here's Python
Screen Shot 2020-04-01 at 6 40 26 PM

Here's R:
Screen Shot 2020-04-01 at 6 40 46 PM

rpkyle:
> Alright, we're very close. However, the R docs are serving an older version `base.css` than the Python docs, perhaps the caching isn't working right?
> 

I'm investigating, going to try to get this one across the finish line 🏁 🏃

chriddyp:
Ah NVM, it's not a CSS issue it's a rendering issue. The R side has `<li><p></p></li>`  and the Python side has `<li></li>`. So it's likely an issue with how the markdown is formatted. We could either fix this:
1. with CSS (a selector for `li p`) or investigate the markdown string in the R side 
2. or by making the markdown string in the R docs better match the Python one
3. Or if the strings are the same, then it might be because one of the packages is outdated and might not include some kind of markdown fix that `dccMarkdown` included. Upgrading the package might help.

chriddyp:
Also it looks like this is not a regression, this issue exists on `master`:
image

So, I'd be OK with addressing this issue in a separate PR.

rpkyle:
> Also it looks like this is not a regression, this issue exists on `master`:
> image
> 
> So, I'd be OK with addressing this issue in a separate PR.

Yeah, let's do that. I think it's time for a 💃 , LGTM!

chriddyp:
⛵️ ⛵️ ⛵️ ⛵️

PR #829, opened 2020-03-25
attempt hiding pagemenu for screenshot tests

chriddyp:
Post-merge checklist:

- [x] I understand

- [ ] I considered whether I could replace the `dcc.Markdown` call with `reusable_components.Markdown`, which will replace relative links with `tools.relpath` internally. Otherwise, I used e.g. `<dccLink href=tools.relpath('/layout') children="the first chapter"/>` instead of `[the first chapter](/layout)` (importing `tools` from `dash_docs`), and set `dangerously_allow_html=true` in the `dcc.Markdown` call.


alexcjohnson:
This will solve the percy issue, without adding extra overhead of waiting for these menus to complete - which is fantastic 🎉

But it means we don't have any tests of the functionality of the page menu. Can we add a test that just loads one page, tests that the page menu becomes correct, then clicks a `dcc.Link` and tests that the correct new entries appear? Just hard-coding the entries for two pages that aren't expected to change often should suffice, I'd say.

chriddyp:
Great idea @alexcjohnson  👍 

Added in the recent commits

chriddyp:
There are still some diffs to be ignored in the bio examples, but I'd like to tackle that later given the number of changes made in https://github.com/plotly/dash-docs/pull/826. Once this PR is in, we'll have much more sane diffs to compare in https://github.com/plotly/dash-docs/pull/826.

chriddyp:
alright, tests are passing @alexcjohnson !

PR #831, opened 2020-03-31
Pattern Matching Callbacks Documentation

chriddyp:
Preliminary documentation on the forthcoming “pattern matching callbacks” (codenamed “wildcards”) feature coming in Dash 1.10.0 or 1.11.0 via https://github.com/plotly/dash/pull/1103/.

New chapter on home page:
image

Preview here: https://dash-docs-pr-831.herokuapp.com/pattern-matching-callbacks
image

TODO:
- [x] Update the `description`
- [x] Update with the proper versions of dash when released (and remove tarballs)


chriddyp:
I'll update the base branch back to `master` once #826 is merged into `master`.

chriddyp:
Alright, I'm pretty happy with this chapter now 🎉 . Ready for a review @alexcjohnson , @rpkyle , @Marc-Andre-Rivet , or @nicolaskruchten .

alexcjohnson:
This looks great! One thing I'm not sure we really talked about but we should document (and discuss if it warrants changes!) is the extra items added to `callback_context` to help with variable IDs. You noted:

> In some cases, it may be important to know which dynamic component changed. As above, you can access this by setting id as State in the callback.

Which is a nice trick! But that info is also made available in `dash.callback_context` as `inputs_list`, `states_list`, and `outputs_list` - though it occurs to me, `outputs_list` isn't always a list... these properties are collections of dictionaries `{'id', 'property'}` and inputs and state also contain a `value` key. This is particularly useful for outputs, as it tells you the shape required for the return value: if `outputs_list` is a single dict, you should return a single value. If it's a list of 5 items, you should return a list of 5 items. If it's a list containing a list of 5 items followed by a single item (ie a multi-output containing an `ALL` then a single item) , that's what you should return.

Marc-Andre-Rivet:
This looks good :) Just need the actual `1.11` to clean up all the tarballs and bump version.

chriddyp:
Good point @alexcjohnson ! I just added a bullet point for each example showing what the various data structures for `dash.callback_context` look like.

chriddyp:
Alright I switched back to the forthcoming 1.11.0 release in 994b455. So, once 1.11.0 is up we should be able to just re-run circle CI and then merge 👍

Marc-Andre-Rivet:
@chriddyp @alexcjohnson As we discussed in DM, merging this in as there's no averse affect on feature/behavior. Will create a follow up issue to re-enable devtools asserts once we figure it out.

PR #840, opened 2020-04-05
Fixes for R clientside callbacks example

rpkyle:
This PR proposes to apply a few minor fixes to the clientside callbacks example in the Dash for R docs:
- wrap `children` of `htmlDetails` (`htmlSummary`, `dccMarkdown`) with a `list` to avoid console error
- use `dash:::to_JSON` before running `jsonlite::prettify` to avoid parse error
- pass `pop` instead of `year` within `dccStore`
- index using `[, "year"]` instead of `year`, since the latter is a function within `data.table` (resolves `invalid subscript type 'closure'` error)
- use `dff[, indicator],` instead of `dff$indicator` to avoid passing `NULL` for `y` vector
- modifies the `namespace` to avoid overriding the `clientside-example.js` with `pagemenu.js`, and corrects the `namespace` in the example formatted with markdown also

🙏 to @Marc-Andre-Rivet for helping me resolve the latter issue

### DIFF
--- a/dash_docs/assets/pagemenu.js
+++ b/dash_docs/assets/pagemenu.js
@@ -1,7 +1,8 @@
-if(!window.dash_clientside) {window.dash_clientside = {};}
-window.dash_clientside.clientside = {
-  pagemenu: function (children) {
-    console.warn('updating pagemenu');
-    return String(Date.now());
-  }
-}
+window.dash_clientside = Object.assign({}, window.dash_clientside, {
+  clientside: {
+    pagemenu: function (children) {
+      console.warn('updating pagemenu');
+      return String(Date.now());
+    }
+  }
+});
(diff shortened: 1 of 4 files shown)


dash_docs/chapters/clientside_callbacks/index.R
```diff
@@ -86,7 +86,7 @@ app$callback(
 
     dccMarkdown("
 ```js
-window.my_clientside_library = {
+window.my_clientside_example = {
```
chriddyp, line 89: I thought this had to equal `window.dash_clientside`?

dash_docs/assets/pagemenu.js
```diff
@@ -1,7 +1,8 @@
-if(!window.dash_clientside) {window.dash_clientside = {};}
-window.dash_clientside.clientside = {
-  pagemenu: function (children) {
-    console.warn('updating pagemenu');
-    return String(Date.now());
-  }
-}
+window.dash_clientside = Object.assign({}, window.dash_clientside, {
```
HammadTheOne, line 1: :+1:

PR #841, opened 2020-04-05
Single input

mbegel:
Update documentation of the chapters about callbacks with the PR [#1180](https://github.com/plotly/dash/pull/1180).
Single `Input` in callbacks don't need to be in a list.


alexcjohnson:
Closing - thanks @mbegel! https://github.com/plotly/dash/pull/1180 ended up much more expansive than when this PR was opened, so I think it warrants a more detailed discussion (and more extensive changes to the examples. New PR to come...

PR #854, opened 2020-04-14
Conditional formatting and filtering improvements

chriddyp:


dash_docs/chapters/dash_datatable/conditional_formatting/index.py
```diff
@@ -0,0 +1,1007 @@
...
+                                'column_id': col
+                            },
+                            'backgroundColor': 'tomato'
+                        } for col in df_with_none.columns
+                    ] +
+                    [
+                        {
+                            'if': {
+                                # TODO - This doesn't work?
+                                'filter_query': '{{{}}} = null'.format(col),
```
chriddyp, line 614: @Marc-Andre-Rivet - Do you have any suggestions for filter queries that can match null values? Here's what this example looks like right now
image

If not, I'll suggest that users replace null values with empty strings `''`

chriddyp:
Alright, this is ready for review!
@rpkyle As before, can you review the content of this PR, looking for typos, suggestions, bugs, grammar improvements, etc?
@Marc-Andre-Rivet - Can you take a look at the DataTable examples and, as before, check if any tests are missing on your end?

I would like to get this chapter in sometime this week ✅

chriddyp:
This is the new chapter: https://dash-docs-pr-854.herokuapp.com/datatable/conditional-formatting

chriddyp:
Ping @Marc-Andre-Rivet @rpkyle ! If either of you don't have time for reviews, feel free to reassign. I would still love to get this in by tomorrow.

dash_docs/chapters/dash_datatable/conditional_formatting/index.py
```diff
@@ -0,0 +1,964 @@
...
+        rc.Markdown("""
+        # Conditional Formatting
+
+        Conditional formatting is provided through the `style_data_conditional`
+        property. The `if` keyword provides a set of conditional formatting
+        statements and the rest of the keywords are camelCased CSS properties.
+
+        The `if` syntax supports three operators: `row_index`, `column_id`,
+        and `filter_query`. `filter_query` is the most flexible option.
+        Here is an example of all three:
```
Marc-Andre-Rivet, line 53: * `column_type:'any'|'numeric'|'text'|'datetime'` is also supported
* `column_editable:true|false` is also supported
* with Dash v1.12, `state:'active'|'selected'` will also be supported

Conditional formatting is also available with:
* `style_cell_conditional`: column_id, column_type (no sure why column_editable is not available here?!)
* `style_filter_conditional`: column_id, column_type, column_editable
* `style_header_conditional`: column_id, column_type, header_index, column_editable

As of Dash v1.12, `column_id`, `row_index` and `header_index` also support arrays of values in addition to discrete values

dash_docs/chapters/dash_datatable/conditional_formatting/index.py
Marc-Andre-Rivet: Yes. You can use the unary operator `is nil` or `is blank`.

`{myfield} is nil` will return `true` for null and undefined
`{my field} is blank` will return `true` for null, undefined, and empty string.

Marc-Andre-Rivet:
@chriddyp Might be interesting to show a case where the filter_query compares two fields from the same row. eg `{myfield1} contains {myfield2}` or some variation, as you are not limited to comparing a field against a static value.

chriddyp:
Thank you @Marc-Andre-Rivet !

 @rpkyle , this ship is sailing! I had to get this in by the end of the week. Feel free to comment on the PR after it is merged and I can follow up with another PR with your suggestions.

PR #858, opened 2020-04-20
Render as 15px to fix wavy text

chriddyp:
Hopefully fixes the issues windows users are seeing with e.g.
image

cc @jdamiba @VeraZab

### DIFF
--- a/dash_docs/assets/override.css
+++ b/dash_docs/assets/override.css
@@ -1,7 +1,7 @@
 /* prefixed by https://autoprefixer.github.io (PostCSS: v7.0.26, autoprefixer: v9.7.3) */
 
 body {
-  font-size: 16px;
+  font-size: 15px;
   line-height: 17px;
   letter-spacing: 0px;
   font-family: "Campaign", serif;
--- a/dash_docs/assets/override.css
+++ b/dash_docs/assets/override.css
@@ -17,7 +17,7 @@ body {
 }
 
 h1, h2, h3, h4, h5, h6 {
-  font-family: "jubilat", sans-serif;
+  font-family: "jubilat", serif;
   margin-top: 0;
   margin-bottom: 0;
   font-weight: 300;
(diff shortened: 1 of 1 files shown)


VeraZab:
A comparison here, if you look at the word 'virtualization' for ex, 15px and 14px do look better imo:

16px:
Screen Shot 2020-04-20 at 11 52 21 AM
15px:
Screen Shot 2020-04-20 at 11 53 16 AM
14px:
Screen Shot 2020-04-20 at 11 53 50 AM
17px:
Screen Shot 2020-04-20 at 11 54 32 AM

alexcjohnson:
Is this just an empirical observation or do we have some understanding of why this would matter? Will these users have problems again if they change page zoom?

I also notice that capitalization is inconsistent, any chance that matters to Windows but not elsewhere, so we're actually falling back on a system file or something rather than the typekit file we intended?

```
~/plotly/dash-docs/dash_docs/assets> ag campaign *.css
override.css
7:  font-family: "Campaign", serif;

typography.css
7: * campaign:
36:font-family:"campaign";
42:font-family:"campaign";
48:font-family:"campaign";
54:font-family:"campaign";
60:font-family:"campaign";
66:.tk-campaign { font-family: "campaign",sans-serif; }
`

VeraZab:
oh good catch for the capitalization we should keep it consistent, and the serif, sans-serif rule too.. 
For the zoom, I've actually observed letters looking better, not worst when zoomed in, commented about it [here](https://github.com/plotly/marketing-website/issues/313#issuecomment-611235331)

chriddyp:
> we should keep it consistent, and the serif, sans-serif rule too..

I don't follow here, what does 'keep it consistent' mean?

VeraZab:
this is the right rule: font-family: "campaign",sans-serif;

chriddyp:
Got it, thanks. I updated the fallback for both fonts 👍

PR #866, opened 2020-04-29
Adds documentation for Workspaces

OwenMatsuda:
This PR adds documentation for Dash Enterprise Workspaces.

Post-merge checklist:

- [x] I understand

### DIFF
--- a/dash_docs/chapters/dash_enterprise/dash_enterprise_chapters.py
+++ b/dash_docs/chapters/dash_enterprise/dash_enterprise_chapters.py
@@ -3943,3 +3943,152 @@ Git = html.Div(children=[
 
     ''')
 ])
+
+WorkspacesUsage = html.Div(children=[
+    html.H1('Using Workspaces'),
+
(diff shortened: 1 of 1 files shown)


chriddyp:
@OwenMatsuda Let's add this to https://github.com/plotly/dash-enterprise-docs/ instead as it will be a top-level item rather than a subchapter of app manager

OwenMatsuda:
I put it in here to be used by https://github.com/plotly/dash-enterprise-docs/pull/283. It seemed like the convention, but I could be wrong. I just guessed off of the code since it wasn't really documented.

gzork:
@chriddyp please correct me if I'm wrong, but @OwenMatsuda my understanding is these docs don't need to be added to the OSS dash docs too, so this PR is not needed I think.

OwenMatsuda:
Ohh I see those were just the docs used both in open source and enterprise. Got it.

chriddyp:
That's correct 👍

PR #871, opened 2020-05-02
clean up dcc.Store handling & limitation docs

alexcjohnson:
Requires https://github.com/plotly/dash-core-components/pull/792 to be released first - this lets us get rid of the special handling of `Store` components, ie loading them all in the initial layout and then preventing the examples from reloading them. I also cleaned up the limitations section for this and other obsolete warnings.

@rpkyle I haven't managed to fix my R env locally... would you mind testing this using https://github.com/plotly/dash-core-components/pull/792? FYI: currently if you open https://dashr.plotly.com/dash-core-components/store and try to click on the `LOCALSTORAGE` or `SESSIONSTORAGE` buttons the page will hang. I think that's because the R docs never implemented the `# no-exec` feature we have on the python side to prevent execution of some lines displayed in an example so it has two copies of each store. That's removed in this PR, but there's one other place we used `# no-exec` on the Python side you may want to take a look at:
```
dash_docs/chapters/dash_cytoscape/applications/examples/usage-phylogeny.py
116:path = os.path.join('data', 'apaf.xml')  # no-exec
```


rpkyle:
@alexcjohnson I'm not able to get the docs in this branch to load, which I imagine is related to the "Sizing" chapter issues I mentioned in #882. But each of the buttons in the example works for me, so I'd say it's OK to 💃 .

I'll apply the workaround we discussed earlier today, then we can redeploy and this fix will be applied also.

PR #875, opened 2020-05-05
Add Converting to HTML functionality for Dash components

eliasdabbas:
This adds initial functionality for converting all Dash components to HTML strings with a few pending points: 

- In some cases the content of the docs might need changing, e.g. three backticks across two children of an `html.Div` etc. 
- Instances of `<dccLink>` need to be handled: 
 a. Unify the usage to one form in the source of the docs e.g. `<dccLink href="/page-url">page name</dccLink>` 
 b. Add conversion functionality to `convert_to_html` converting `<dccLink>` to `<a>`
- Instances where a page's conent is in a list in `URL_TO_CONTENT_MAP` to be wrapped in an `html.Div`.

Post-merge checklist:

- [ ] I understand


chriddyp:
🎉 🎉 🎉 

Great stuff @eliasdabbas , I'll review this today!

chriddyp:
This looks great @eliasdabbas ! I think we are very close to shipping this 🚀  I'll comment more line-by-line on some things I just added.

I added some commits that gets it to work locally and on the review app, you can check out the live version here: https://dash-docs-pr-875.herokuapp.com/

It seems to work pretty well! Frame-by-frame, it looks like:
1. SSR content is loaded
2. Once the initial JS finishes loading & runs, Dash's front-end clears out the SSR content with this empty loading div: https://github.com/plotly/dash-renderer/blob/dd86afc6bac2eb11c27acf972293e6d12ffcd4f7/src/AppContainer.react.js#L31
3. Some more JS ends up getting loaded asynchronously as the app begins to render
4. The full app is finally loaded.

So, when you load the page, you'll Frame 2 for about a second. It's a little jarring, because there is a flash of SSR content, and then a flash of nothing, and then the app is progressively loaded.

On fast machines & networks, Frame 2 happens so quickly that you don't even really see Frame 1. However, for bots and for machines without JS, Frame 1 is very functional. That's good enough for me!

In the future, we might add some ability to customize that front-end loading div or delay it for use cases like this.

throttled-2

Since Frame 1 is so short, I'd say that we don't need to spend too much time optimizing or tweaking the style.

So, for this PR to get merged I believe we'll just need the tests to pass: the remaining 6 pages to load OK:
```
Issue with /datatable/conditional-formatting
object of type 'numpy.float64' has no len()
===============
Issue with /performance
'list' object has no attribute 'to_plotly_json'
===============
Issue with /live-updates
'list' object has no attribute 'to_plotly_json'
===============
Issue with /urls
'list' object has no attribute 'to_plotly_json'
===============
Issue with /deployment
'list' object has no attribute 'to_plotly_json'
===============
Issue with /dash-enterprise
'NoneType' object has no attribute 'to_plotly_json'
```

Ping me if you have any questions with the remaining tests / pages or need any help!

dash_docs/convert_to_html.py
```diff
@@ -0,0 +1,159 @@
...
+    return '<{}{}>{}{}'.format(tag, attrib_str, children, closing_tag)
+
+
+def markdown_to_html(md_text, extensions=md_extensions):
+    """Convert ``md_text`` to HTML string.
+    To enable/disable ``extensions`` simply add/remove from the list.
+    Please see the documentation for details:
+    https://python-markdown.github.io/extensions/
+    """
+    escape_html = html_tag_regex.sub('&lt;\g<1>&gt;', md_text)
```
chriddyp, line 154: We'll likely need to replace embedded `<dccLink` before we escape it with `&lt;dccLink`

dash_docs/chapter_index.py
```diff
@@ -1032,6 +1034,14 @@ def create_url_mapping(url_set):
                 'breadcrumb',
                 section['name']
             )
+            # TODO - Remove try/except before pushing to prod
```
chriddyp, line 1037: Since our tests verify that the `convert_to_html` works on all pages, we can remove this `try` / `except` once the tests pass and before we merge into `master`

tests/unit/test_html_conversion.py
```diff
@@ -0,0 +1,45 @@
...
+)
+def test_html_conversion():
+    exceptions = []
+    success = []
+    dcc_link = []
+    from dash_docs.chapter_index import URL_TO_CONTENT_MAP
+    for url in URL_TO_CONTENT_MAP:
+        try:
+            ssr_content = convert_to_html(URL_TO_CONTENT_MAP[url])
+            if '<dccLink' in ssr_content:
```
chriddyp, line 17: Looks like this isn't catching anything because we're converting `<dccLink` to `&lt;dccLink`

tests/unit/test_html_conversion.py
```diff
@@ -0,0 +1,45 @@
...
+                dcc_link.append(url)
+            else:
```
chriddyp, line 19: We can place other verification checks in here, like verifying that the HTML is valid. Any other checks we should do?

dash_docs/convert_to_html.py
chriddyp: Actually, it looks a little more nuanced than this. Here is the markdown wrapper that we're using:
https://github.com/plotly/dash-docs/blob/51b1b71406d3fd4b59792ae3b342f3bed0fdcf98/dash_docs/reusable_components/Markdown.py#L20-L38

Note the `dangerously_allow_html` flag. So, this logic will be something like:
1. If `component_dict['props'].get('dangerously_allow_html', False)`, then don't escape anything _but_ do replace `<dccLink` with `<a` and `</dccLink>` with `</a>`
2. Otherwise, escape the text like you have it.

dash_docs/convert_to_html.py
```diff
@@ -0,0 +1,159 @@
...
+    link_standalone_urls = re.sub('(\s|\n|^)(https?://.*?)($|\s)',
+                                  '\g<1><a href="\g<2>">\g<2></a>\g<3>',
+                                  escape_html)
```
chriddyp, line 157: Could you explain this what we're replacing here, perhaps with a comment? I'm awful at reading regexes 😸

dash_docs/convert_to_html.py
```diff
@@ -0,0 +1,159 @@
...
+                                                not in ['Graph', 'DataTable'] else '300px'
+        else:
+            div.style = {'height': '50px' if component_dict['type'] not in ['Graph', 'DataTable'] else '300px'}
+    return div
+
+
+def convert_to_html(component):
+    if isinstance(component, str):
+        component = html.Span(component)
+    component_dict = component.to_plotly_json()
```
chriddyp, line 105: From the tests, it looks like there are some issues with `to_plotly_json`. A couple options here:
1. You could go through each of these URLs and replace `layout = [` with `layout = html.Div([`.
2. You could check if this is a list, and if so, replace `component` with `html.Div(component)`
```
Issue with /datatable/conditional-formatting
object of type 'numpy.float64' has no len()
===============
Issue with /performance
'list' object has no attribute 'to_plotly_json'
===============
Issue with /live-updates
'list' object has no attribute 'to_plotly_json'
===============
Issue with /urls
'list' object has no attribute 'to_plotly_json'
===============
Issue with /deployment
'list' object has no attribute 'to_plotly_json'
===============
Issue with /dash-enterprise
'NoneType' object has no attribute 'to_plotly_json'
```

chriddyp:
> In some cases the content of the docs might need changing, e.g. three backticks across two children of an html.Div etc.

Could you share an example of a URL where this comes up? I'll load it with JS disabled to see the raw content

dash_docs/convert_to_html.py
```diff
@@ -0,0 +1,161 @@
...
+        else:
+            div.style = {'height': '50px' if component_dict['type'] not in ['Graph', 'DataTable'] else '300px'}
+    return div
+
+
+def convert_to_html(component):
+    if isinstance(component, str):
+        component = html.Span(component)
+    if component is None:
+        return ''
```
chriddyp, line 106: @eliasdabbas - I added this check in just to get the home page from rendering https://github.com/plotly/dash-docs/pull/875/commits/f8cdd74f314a71ab95aef88b716cc15c37ea379c. I don't really understand it though, so feel free to remove it if there is a different kind of fix.

One thing that's odd, and maybe related to this `is None` check is that the home page content doesn't really render. See screenshot below. There should be a list of links here.

image

dash_docs/convert_to_html.py
```diff
@@ -0,0 +1,161 @@
...
+        if _translate_attrib(bool_attr).lower() in bool_html_attrs:
+            if bool(component_dict['props'][bool_attr]):
+                bool_attrs.append(_translate_attrib(bool_attr).lower())
+        bool_str = ' '.join(bool_attrs)
+    if 'style' in component_dict['props']:
+        style_str = 'style=' + _style_to_attrib(component_dict['props']['style'])
+    attrib_str = ' '.join([attrib_str, style_str, bool_str]).strip()
+    children = component_dict['props'].get('children') or ''
+    initial_indent = 0
+    if isinstance(children, str):
```
chriddyp, line 137: Children can be a few different things beyond strings: numbers, None, booleans are all supported. It might be easier to check what it "isn't" (not list-like `isinstance(children, (tuple, MutableSequence))` and not a component `isinstance(children, Component)` - here's how we do this in the component's own recursive crawler:
https://github.com/plotly/dash/blob/5b329f41f48fdbe723eb8e0f81d4c4731d45b1db/dash/development/base_component.py#L276-L295

dash_docs/convert_to_html.py
```diff
@@ -0,0 +1,161 @@
...
+            if 'height' not in div.style:
+                div.style['height'] = '50px' if component_dict['type'] \
+                                                not in ['Graph', 'DataTable'] else '300px'
+        else:
+            div.style = {'height': '50px' if component_dict['type'] not in ['Graph', 'DataTable'] else '300px'}
+    return div
+
+
+def convert_to_html(component):
+    if isinstance(component, str):
```
chriddyp, line 103: https://github.com/plotly/dash-docs/pull/875/files#r420468120 applies here as well

chriddyp:
Alright, I think that's the last of my comments for now. Ping me when you're ready for another review - we're close! 🎉

dash_docs/convert_to_html.py
eliasdabbas: Sure, in some cases we have URLs in the content, like [here](https://dash.plotly.com/dash-html-components/a). The regex makes it linkable, with the URL being the ancho text. 
Thie `http://example.com` becomes `<a href="http://example.com">http://example.com</a>`

I added a comment to the code. We can also think of a better name than `link_standalone_urls`

dash_docs/convert_to_html.py
eliasdabbas: Done. I changed the source code as you mentioned in point 1. I think it's better, more consistent docs, and no changes in the converter logic. 
I also fixed the float issue.

eliasdabbas:
> So, for this PR to get merged I believe we'll just need the tests to pass: the remaining 6 pages to 

Wow, didn't know it could be that fast! I fixed these issues, and I was able to generate all the pages with no issues locally. I pushed the fixes if you want to launch, and will be working on the other issues. 

In addition to the points you mention, the following are basically empty if you look at the HTML, but not causing issues. I'll work on them. 

`/dash-enterprise`
`/datatable`
`/cytoscape`

dash_docs/convert_to_html.py
eliasdabbas: Interesting. This is similar to the problem encountered with the four other pages that are generating HTML with mainly placeholder content, but not the correct one. I'm adding to that list, and will try to figure it out. 

The `is None` check is because I encountered a case where one of the children of an `html.Div` was `None`. I assumed this might be a case to be handled frequently, and it mainly gets called recursively, in which case an empty string is returned. 

One example is here:
`URL_TO_CONTENT_MAP['/dash-enterprise'].children[1].children[1]`

We might get rid of this check altogether and remove all instances of `None` in the children of all divs, unless there's a use case for it that I can't think of?

dash_docs/convert_to_html.py
chriddyp: Great comment 💯  thanks! 👍

dash_docs/convert_to_html.py
eliasdabbas: @chriddyp 

How about this approach: 

We convert all `<dccLink .* />`s to markdown links. 
For example: 

`<dccLink href="/home-page" children="Home Page"/>` becomes: 
`[Home Page](/home-page)`

This conversion happens the first thing in `markdown_to_html`, and these become regular markdown strings. Relative links that work, with descriptive anchor text, is SEO gold. 

I wrote this function and tried it, and got no issues for all pages: 

```python
import re
from dash_docs.convert_to_html import markdown_to_html

text = """
    The <dccLink href="/interactive-graphing" children="Interactive Visualizations"/> tutorial explains how
    to capture user interaction events with a `dcc.Graph`, and how to update the
    `figure` property in callbacks.
    
    Also checkout <dccLink children="the children" href="/home-page"/> page
    """


def _dccLink_to_a_href(text):
    dcclink_regex = re.compile(r'<dccLink .*?/>')
    href_regex = re.compile(r'<dccLink .*?href="(.*?)".*/>')
    children_regex = re.compile(r'<dccLink .*?children="(.*?)".*/>')
    replacements = []
    hrefs = re.findall(href_regex, text)
    children = re.findall(children_regex, text)
    for href, child in zip(hrefs, children):
        replacements.append(f'[{child}]({href})')
    # split by <dccLink .* /> then stitch with `replacements`
    converted = ''.join([''.join(x)
                         for x in zip(re.split(dcclink_regex, text),
                                      replacements + [''])])
    return converted



print(markdown_to_html(_dccLink_to_a_href(text)))

<p>The <a href="/interactive-graphing">Interactive Visualizations</a> tutorial explains how<br />
to capture user interaction events with a <code>dcc.Graph</code>, and how to update the<br />
<code>figure</code> property in callbacks.</p>
<p>Also checkout <a href="/home-page">the cildren</a> page</p>


```

The updated `markdown_to_html` gets an additional line: 

```python
def markdown_to_html(md_text, extensions=md_extensions):
    """Convert ``md_text`` to HTML string.
    To enable/disable ``extensions`` simply add/remove from the list.
    Please see the documentation for details:
    https://python-markdown.github.io/extensions/
    """
    # markdown isn't supported in Python 2, so protect the import
    # we don'ta actually deploy the docs on Python 2, we're
    # just running some tests, so the SSR functionality doesn't
    # need to work
    if sys.version_info < (3, 5):
        return md_text
    import markdown
    convert_dccLinks = _dccLink_to_a_href(md_text)
    escape_html = html_tag_regex.sub('&lt;\g<1>&gt;', convert_dccLinks)
    # convert http://example.com to <a href="http://example.com">http://example.com</a>
    link_standalone_urls = re.sub('(\s|\n|^)(https?://.*?)($|\s)',
                                  '\g<1><a href="\g<2>">\g<2></a>\g<3>',
                                  escape_html)
    return markdown.markdown(textwrap.dedent(link_standalone_urls),
                             extensions=extensions)
```

What do you think?

chriddyp: @eliasdabbas - I like it!

eliasdabbas: Great! 
I just pushed a few commits for this: 
- A function that converts `<dccLink>`s to markdown syntax
- A similar function that does the same for `<blockquote>`s, converting to `> `
- In some cases the closing tags are written as `</dccLink>`, which I changed to `/>`

This should solve the other issues related to `<dccLink>` tags. 
Pending now is the fixing of the 4-5 pages that are generating valid but wrong HTML code. Will update you.

tests/unit/test_html_conversion.py
eliasdabbas: Resolved by converting to markdown syntax `[Children](/href)`

tests/unit/test_html_conversion.py
eliasdabbas: The problem is that the 4-5 problematic pages have valid HTML, it's just the wrong HTML. For the docs site, this is probably good enough since we will check manually anyways, and we don't have thousands of pages, but it needs proper testing and validation for general use. I'll let you know if I think of something.

chriddyp:
This is starting to look good @eliasdabbas ! One issue on the Python-2.7 side of things from circleci:
```
tests/unit/test_html_conversion.py:3: in <module>
    from dash_docs.convert_to_html import convert_to_html
E     File "dash_docs/convert_to_html.py", line 161
E       replacements.append(f'[{child}]({href})')
E                                              ^
E   SyntaxError: invalid syntax
 
```

Unfortunately we can't use this nice python 3 syntax yet as we run our tests in python 2 (about 10-20% of our users still use python 2, so we run tests to at least make sure that our documentation works in py2)

eliasdabbas:
> Could you share an example of a URL where this comes up? I'll load it with JS disabled to see the raw content

`/installation`:   probably because of having > at the beginning of all lines?
`/loading-states`: it was missing a set of closing backticks, I just fixed it 

`/sharing-data/index.py` there's an odd number of backticks,  it's probably because [here](https://github.com/plotly/dash-docs/blob/9d2c0d2e52a14e52cda7928f25027861fac67595/dash_docs/chapters/sharing_data/index.py#L533) we replace "```python" with the empty character, but should also do the same for the closing backticks? 

The HTML of this pages is showing the backticks: 

Screen Shot 2020-05-09 at 12 19 02 PM


In some cases like `/datatable-reference` and `/dash-core-components/link`, I see "\``` shell", which seems that it should be "\``` python" because it's calling Python's `help` function?

eliasdabbas:
> 1. SSR content is loaded
> 2. Once the initial JS finishes loading & runs, Dash's front-end clears out the SSR content with this empty loading div: https://github.com/plotly/dash-renderer/blob/dd86afc6bac2eb11c27acf972293e6d12ffcd4f7/src/AppContainer.react.js#L31
> 3. Some more JS ends up getting loaded asynchronously as the app begins to render
> 4. The full app is finally loaded.
> 

Thanks for the explanation, I had no clue how it works! :) 


> On fast machines & networks, Frame 2 happens so quickly that you don't even really see Frame 1. However, for bots and for machines without JS, Frame 1 is very functional. That's good enough for me!
> 

Yes, I've been trying, on slower connections, but it's looking OK to me. On the phone there's a very brief blink, but I still can't see the intermediate page.

eliasdabbas:
> Unfortunately we can't use this nice python 3 syntax yet as we run our tests in python 2 (about 10-20% of our users still use python 2, so we run tests to at least make sure that our documentation works in py2)

Yes, sorry, my bad! Fixed.

chriddyp:
> Yes, sorry, my bad! Fixed.

Hey @eliasdabbas - Have you pushed up your latest commits yet? 🙂 I'm still seeing some `f'` string errors on circleci's tests: https://app.circleci.com/pipelines/github/plotly/dash-docs/375/workflows/49eafba4-bf58-4d76-9db7-507f4e944a52/jobs/11649/steps

chriddyp:
> I see "``` shell", which seems that it should be "``` python" because it's calling Python's help function?

Yeah, that sounds right 👍 

> here's an odd number of backticks, it's probably because here we replace "```python" with the empty character, but should also do the same for the closing backticks?

Yeah, that's an odd example, I don't totally understand why that particular example is a "special case" and requires replacing backticks... but yes replacing the closing ones too should be fine for now 👍

eliasdabbas:
@chriddyp 

So, it seems that with the following exceptions we are good to go!

`/`
`/dash-enterprise`
`/datatable`
`/cytoscape`

- I pushed the agreed changes (f-strings (thought I already did), shell replaced with python, minor bug with iframes)
- I narrowed down the issue in the above three examples, though I still don't have a solution!
It seems the issue is with `rc.Section`. While the converter is working properly on a plain `rc.Section` string, it's not working while generating the HTML for all the pages. For example: 

```python
print(convert_to_html(html.Div([dcc.Markdown('test markdown'), rc.Section('quickstart', [rc.Markdown('list item 1'), rc.Markdown('list item 2')])])))
```

<div>
 <div class="ssr-dash-core-components-markdown"><p>test markdown</p></div>
  <div class="toc--section">
 <h2 style="border-bottom:thin lightgrey solid; margin-top:50px">quickstart</h2>
  <span></span>
   <ul class="toc--chapters">
 <div class="ssr-dash-core-components-markdown"><p>list item 1</p></div>
  <div class="ssr-dash-core-components-markdown"><p>list item 2</p></div>
</ul>
</div>
</div>


Not sure what I'm missing here. I'll keep digging, just wanted to update you. Please let me know if you have suggestions. Thanks!

chriddyp:
> Not sure what I'm missing here. I'll keep digging, just wanted to update you. Please let me know if you have suggestions. Thanks!

It looks like there was an issue with converting component trees with children that weren't in a list, so e.g.
```
convert_to_html(html.Div(html.Div('Test')))
```

I added a fix to this behavior in https://github.com/plotly/dash-docs/pull/875/commits/a3445ceb66c81b5992de63731a52648a865fe458, take a look to see if

chriddyp:
Home page, cytoscape, datatable, and dash enterprise pages now render 👍 
Screen Shot 2020-05-13 at 1 04 26 PM
Screen Shot 2020-05-13 at 1 04 17 PM
Screen Shot 2020-05-13 at 1 04 07 PM
Screen Shot 2020-05-13 at 1 03 58 PM

chriddyp:
It looks like there are two remaining issues with `/clientside-callbacks` & `/cytoscape/events`
```
Error creating SSR for /clientside-callbacks
Traceback (most recent call last):
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/chapter_index.py", line 1041, in create_url_mapping
    section['content']
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 141, in convert_to_html
    for i, child in enumerate(comp_children)])
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 141, in <listcomp>
    for i, child in enumerate(comp_children)])
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 143, in convert_to_html
    children = convert_to_html(comp_children)
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 141, in convert_to_html
    for i, child in enumerate(comp_children)])
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 141, in <listcomp>
    for i, child in enumerate(comp_children)])
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 141, in convert_to_html
    for i, child in enumerate(comp_children)])
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 141, in <listcomp>
    for i, child in enumerate(comp_children)])
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 115, in convert_to_html
    component = dcc_to_html(component)
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 92, in dcc_to_html
    div.children = markdown_to_html(component_dict['props']['children'])
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 190, in markdown_to_html
    convert_blockquotes = _convert_blockquotes(md_text)
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 174, in _convert_blockquotes
    '> \g<1>', text)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/re.py", line 191, in sub
    return _compile(pattern, flags).sub(repl, string, count)
TypeError: expected string or bytes-like object

Error creating SSR for /cytoscape/events
Traceback (most recent call last):
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/chapter_index.py", line 1041, in create_url_mapping
    section['content']
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 141, in convert_to_html
    for i, child in enumerate(comp_children)])
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 141, in <listcomp>
    for i, child in enumerate(comp_children)])
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 143, in convert_to_html
    children = convert_to_html(comp_children)
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 141, in convert_to_html
    for i, child in enumerate(comp_children)])
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 141, in <listcomp>
    for i, child in enumerate(comp_children)])
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 115, in convert_to_html
    component = dcc_to_html(component)
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 92, in dcc_to_html
    div.children = markdown_to_html(component_dict['props']['children'])
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 190, in markdown_to_html
    convert_blockquotes = _convert_blockquotes(md_text)
  File "/Users/chriddyp/Repos/dash-stuff/dash-docs/dash_docs/convert_to_html.py", line 174, in _convert_blockquotes
    '> \g<1>', text)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/re.py", line 191, in sub
    return _compile(pattern, flags).sub(repl, string, count)
TypeError: expected string or bytes-like object
```

dash_docs/chapters/dash_core_components/Checklist/index.py
```diff
@@ -14,9 +14,9 @@
     rc.Markdown(
     '''
     `dcc.Checklist` is a component for rendering a set of checkboxes.
-    See also <dccLink href="/dash-core-components/radioitems">RadioItems</dccLink>
+    See also <dccLink href="/dash-core-components/radioitems">RadioItems/>
     for selecting a single option at a time or
-    <dccLink href="/dash-core-components/dropdown">Dropdown</dccLink> for
+    <dccLink href="/dash-core-components/dropdown">Dropdown/> for
```
chriddyp: Looks like a typo here and in `Radioitems/index.py`, fixed in 9b59e2d9

chriddyp:
> It looks like there are two remaining issues with /clientside-callbacks & /cytoscape/events

Should be fixed in 80ebccd

eliasdabbas:
Thanks for the fixes @chriddyp ! 
I can see all tests passing, with a few visual issues raised by Percy. 

I knew about the issue raised when `children` is not a list, and I thought it was handled here, but it seems it didn't handle all cases well: 

```python
if isinstance(comp_children, (int, float)):	
    comp_children = [comp_children]	
if len(comp_children) == 1 and not isinstance(comp_children, list):	
        comp_children = [comp_children]
```
Anyway, I think it's looking great with the fixes you added. 

Thanks again, and please let me know what's next and if anything is needed before we launch!

tests/integration/test_main_page.py
```diff
@@ -1,15 +1,23 @@
...
+
 
 from dash_docs.chapter_index import URL_TO_CONTENT_MAP
 
 logger = logging.getLogger(__name__)
 
 
+@pytest.mark.skipif(
+    sys.version_info < (3, 0),
+    reason="Python 2.7's dev flask server is timing out on many pages now"
```
chriddyp, line 14: fyi @plotly/dash-core - I'm skipping the Python 2.7 snapshot tests on the docs now, as flask's dev server has a broken socket connection when loading many pages. The majority of content in this project is loaded during `import`, so we're still implicitly testing for syntax errors in the examples that would cause runtime exceptions.

chriddyp: OK?

alexcjohnson: OK by me. Really feels like Py2 has hit an inflection point recently where it's getting harder to continue using.

chriddyp:
Alright, I think we're good @eliasdabbas ! NICE WORK 🎉 

OK if I merge & deploy?

eliasdabbas:
> Alright, I think we're good @eliasdabbas ! NICE WORK 🎉
> 
> OK if I merge & deploy?

Thank you! 🎉🎉🎉
Can't wait :)

PR #876, opened 2020-05-05
Update Dash to 1.12

Marc-Andre-Rivet:
- Update DataTable conditional formatting example to include `state` usage
- Update Dash to 1.12.0
- Update DAQ to 0.5.0

### DIFF
--- a/dash_docs/chapters/dash_datatable/conditional_formatting/index.py
+++ b/dash_docs/chapters/dash_datatable/conditional_formatting/index.py
@@ -155,16 +155,6 @@ layout = html.Div(
         del df['id']  # no-display
         """),
 
-        # TODO - Add this block in when 1.13 is out.
-        # {
-        #     'if': {
-        #         'state': 'active'  # 'active' | 'selected'
-        #     },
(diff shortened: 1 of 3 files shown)


chriddyp:
@Marc-Andre-Rivet - Could you also add this block into the first example? https://github.com/plotly/dash-docs/blob/master/dash_docs/chapters/dash_datatable/conditional_formatting/index.py#L149-L157. That is, if I have the syntax right.

dash_docs/chapters/dash_datatable/conditional_formatting/index.py
```diff
@@ -141,21 +141,20 @@
                     'textAlign': 'left'
                 },
 
+                {
+                    'if': {
+                        'state': 'active'  # 'active' | 'selected'
+                    },
+                    'backgroundColor': 'DodgerBlue',
+                    'opacity': 0.5,
+                    'border': '2px solid DodgerBlue'
```
chriddyp, line 150:
```suggestion
 ```suggestion
                   'backgroundColor': 'rgba(0, 116, 217, 0.3)'
```
```

PR #889, opened 2020-05-29
Clarify htmlButton example in Dash for R docs

rpkyle:
@mjkallen helpfully suggested in https://github.com/plotly/dashR/issues/126 that the default value of `n_clicks` is confusing, and offered a modification to the `htmlButton` example that helps clarify this.

I've pushed up a revision in this branch that aims to improve the example.

### DIFF
--- a/dash_docs/chapters/dash_core_components/Button/examples/button.R
+++ b/dash_docs/chapters/dash_core_components/Button/examples/button.R
@@ -1,5 +1,4 @@
 library(dash)
-library(dashCoreComponents)
 library(dashHtmlComponents)
 
 app <- Dash$new()
@@ -20,7 +19,12 @@ app$callback(
   params=list(input(id = "button", property = "n_clicks"),
               input(id = "input-box", property = "value")),
   function(n_clicks, value) {
-    sprintf("The input value was \'%s\' and the button has been clicked \'%s\' times", value, n_clicks)
-  })
+    if (is.list(n_clicks)) {
+      sprintf("'n_clicks' is a list of length %d", length(n_clicks))
+    } else {
+      sprintf("'n_clicks' is a vector and you have clicked the button %d time(s)", n_clicks)
+    }
+  }
+)
 
 app$run_server()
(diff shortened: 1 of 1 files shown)


dash_docs/chapters/dash_core_components/Button/examples/button.R
```diff
@@ -20,7 +19,12 @@ app$callback(
   params=list(input(id = "button", property = "n_clicks"),
               input(id = "input-box", property = "value")),
   function(n_clicks, value) {
-    sprintf("The input value was \'%s\' and the button has been clicked \'%s\' times", value, n_clicks)
-  })
+    if (is.list(n_clicks)) {
+      sprintf("'n_clicks' is a list of length %d", length(n_clicks))
+    } else {
+      sprintf("'n_clicks' is a vector and you have clicked the button %d time(s)", n_clicks)
```
alexcjohnson, line 25: Does R have a concept like Python's `repr`, that would show precisely what data structure you have? That might be more useful than trying to describe it in words.

Also you aren't using `value` at all anymore - should either put it back in to the output string or remove it as an input.

dash_docs/chapters/dash_core_components/Button/examples/button.R
```diff
@@ -20,7 +19,13 @@ app$callback(
   params=list(input(id = "button", property = "n_clicks"),
               input(id = "input-box", property = "value")),
   function(n_clicks, value) {
-    sprintf("The input value was \'%s\' and the button has been clicked \'%s\' times", value, n_clicks)
-  })
+    object_as_string <- capture.output(dput(n_clicks))
+    if (is.list(n_clicks)) {
+      sprintf("'n_clicks' is of length %d, with the value '%s'", length(n_clicks), object_as_string)
+    } else {
+      sprintf("'n_clicks' is of length %d, and you have clicked the button %d time(s), and the value is '%s'", length(n_clicks), n_clicks, value)
```
alexcjohnson, line 26:
```suggestion
      sprintf("'n_clicks' is of length %d, you have clicked the button %d time(s), and the value is '%s'", length(n_clicks), n_clicks, value)
```

mjkallen:
Nice work!

PR #899, opened 2020-06-17
Add Dash Bootstrap Components

xhlulu:
Post-merge checklist:

- [ ] I understand


dash_docs/chapter_index.py
```diff
@@ -633,6 +633,22 @@ def component_list(
...
+                'chapters': [
+                    {
+                        'url': 'https://dash-bootstrap-components.opensource.faculty.ai/',
+                        'description': (
+                            'A library of Bootstrap components '
+                            'created by [faculty.ai](https://faculty.ai/). '
+                            'Dash Boostrap Components makes it easier to '
+                            'build consistently styled '
+                            'apps with complex, responsive layouts.'
+                        )
```
rpkyle, line 647: @xhlulu Should we add this on the R side as well?

xhlulu: I'm unsure whether DBC supports R, they don't have any docs on R in their webapge: https://dash-bootstrap-components.opensource.faculty.ai/

xhlulu:
Any idea why [the tests](https://app.circleci.com/pipelines/github/plotly/dash-docs/456/workflows/19e1fdf3-6750-45c6-8656-5888028cffb0/jobs/11938/tests) are failing? This is the output:
```
tests/integration/conftest.py:12: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
venv/src/dash/dash/testing/application_runners.py:76: in __call__
    return self.start(*args, **kwargs)
venv/src/dash/dash/testing/application_runners.py:209: in start
    wait.until(lambda: self.accessible(self.url), timeout=start_timeout)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

wait_cond = <function ProcessRunner.start.<locals>.<lambda> at 0x7f790b59fdd0>
timeout = 60, poll = 0.1, msg = 'expected condition not met within timeout'

    def until(
        wait_cond, timeout, poll=0.1, msg="expected condition not met within timeout"
    ):  # noqa: C0330
        res = wait_cond()
        logger.debug(
            "start wait.until with method, timeout, poll => %s %s %s",
            wait_cond,
            timeout,
            poll,
        )
        end_time = time.time() + timeout
        while not res:
            if time.time() > end_time:
>               raise TestingTimeoutError(msg)
E               dash.testing.errors.TestingTimeoutError: expected condition not met within timeout

venv/src/dash/dash/testing/wait.py:25: TestingTimeoutError
```

I'm not totally sure how i broke the tests; looking at the heroku deployment, it seems fine.

chriddyp:
@xhlulu - Could you try rerunning the tests? Could be an intermittent failure.

dash_docs/chapter_index.py
rpkyle: @xhlulu Yes, it's supported now 🎉 

https://dash-bootstrap-components.opensource.faculty.ai/docs/dashr/

xhlulu: Awesome! In this case i think it's a good idea if they are added to the R side.

alexcjohnson:
A few questions / issues:
- The sidebar link does nothing but open an empty menu - special handling of this case, to keep the link looking as it does but add a clickable entry inside it?
Screen Shot 2020-06-22 at 12 00 14 PM
- Do we want to do something to mark external links like this? Frustrating that this [isn't in unicode](https://www.quora.com/Is-the-symbol-for-external-link-available-in-Unicode-If-so-how-do-I-get-in-on-my-Mac) but we could add it as SVG.
- @xhlulu are you able to figure out adding the link in R or do you need @rpkyle to help?

rpkyle:
> * @xhlulu are you able to figure out adding the link in R or do you need @rpkyle to help?

@xhlulu Let me know if there's anything you need from me, but it would be awesome to add in the DBC content for R here!

xhlulu:
> > * @xhlulu are you able to figure out adding the link in R or do you need @rpkyle to help?
> 
> @xhlulu Let me know if there's anything you need from me, but it would be awesome to add in the DBC content for R here!

I'm so sorry, for some reason I didn't see this. I haven't touched dash-R docs befor e(or R in general), so I'd be happy if @rpkyle you would like to add it to Dash-R docs.

xhlulu:
@alexcjohnson 

> * Do we want to do something to mark external links like this? Frustrating that this isn't in unicode but we could add it as SVG.
> * The sidebar link does nothing but open an empty menu - special handling of this case, to keep the link looking as it does but add a clickable entry inside it?

What's the best way I can resolve those problems in this PR?

xhlulu:
Actually, looking at this line:

https://github.com/plotly/dash-docs/blob/6715265ed9e57b13f82b9f14c73812d80ece7275/dash_docs/chapter_index.py#L536-L548

Would we be able to simply do `'content': None` here without breaking the app?

alexcjohnson:
> Do we want to do something to mark external links like this? Frustrating that this isn't in unicode but we could add it as SVG.

Don't worry about it in this PR - https://github.com/plotly/dash-docs/issues/911

> The sidebar link does nothing but open an empty menu - special handling of this case, to keep the link looking as it does but add a clickable entry inside it?

> Would we be able to simply do 'content': None here without breaking the app?

I'm not sure - maybe that would work, maybe it would work to flatten it so this one looks more like the entries in for example the callbacks section, or maybe neither of those would work and it just needs special handling in the sidebar creator.

> I'd be happy if @rpkyle you would like to add it to Dash-R docs.

Let's finish up the Python side first, but then Ryan can you add the R equivalent in this PR before we merge? Hopefully soon we'll be at a point where everyone feels comfortable adding Py and R content side-by-side but we may need one or two more examples of that first 🙏

xhlulu:
> > Do we want to do something to mark external links like this? Frustrating that this isn't in unicode but we could add it as SVG.
> 
> Don't worry about it in this PR - #911
> 
> > The sidebar link does nothing but open an empty menu - special handling of this case, to keep the link looking as it does but add a clickable entry inside it?
> 
> > Would we be able to simply do 'content': None here without breaking the app?
> 
> I'm not sure - maybe that would work, maybe it would work to flatten it so this one looks more like the entries in for example the callbacks section, or maybe neither of those would work and it just needs special handling in the sidebar creator.
> 
> > I'd be happy if @rpkyle you would like to add it to Dash-R docs.
> 
> Let's finish up the Python side first, but then Ryan can you add the R equivalent in this PR before we merge? Hopefully soon we'll be at a point where everyone feels comfortable adding Py and R content side-by-side but we may need one or two more examples of that first

Sounds good. I'll try out what i mentioned tomorrow and we can see how it goes.

xhlulu:
I just added a "name" to the dictionary, now it shows up in the sidebar as well as the main page. Does that fix it?

xhlulu:
If the sidebar looks the way we want, @rpkyle would you like to take over this PR to add the R docs?

HammadTheOne:
I've updated the Dash R docs to match the entry for the Python docs, and changed `External Link` to `faculty.ai` so people are aware of where they're headed if they select it from the sidebar. 

Closing this PR.

PR #901, opened 2020-06-17
Update dash-cytoscape requirements to v0.1.1

xhlulu:
Let's see if this might solve some of the problems associated with the snippets.

Post-merge checklist:

- [ ] I understand

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ dash-auth==1.3.2
 dash-bio-utils==0.0.5
 dash-bio==0.4.8
 dash-canvas==0.1.0
-dash-cytoscape==0.0.5
+dash-cytoscape==0.1.1
 dash-dangerously-set-inner-html==0.0.2
 dash-daq==0.5.0
 dask==1.2.0


alexcjohnson:
@xhlulu the test failure looks like a real py2 syntax issue?

```
    from .Tree import Tree
E     File "/home/circleci/project/venv/lib/python2.7/site-packages/dash_cytoscape/utils/Tree.py", line 85
E       **self.edge_data
E        ^
E   SyntaxError: invalid syntax
```

rpkyle:
> @xhlulu the test failure looks like a real py2 syntax issue?
> 
> ```
>     from .Tree import Tree
> E     File "/home/circleci/project/venv/lib/python2.7/site-packages/dash_cytoscape/utils/Tree.py", line 85
> E       **self.edge_data
> E        ^
> E   SyntaxError: invalid syntax
> ```

Appears potentially related to (closed) https://github.com/plotly/dash-cytoscape/issues/60. I recall talking with @Marc-Andre-Rivet about this last fall, at the time I wasn't able to build `dash-cytoscape` with py27 at all. 

Might help to require py3+  explicitly, unless we want to modify the package so that py2.7 is supported too.

xhlulu:
I'm not against explicitly requiring python 3, but i think it's a conversation we should have for a new release. As for this, is there something in the code i could change to make the tests pass?

alexcjohnson:
@xhlulu the PR here is updating the version to v0.1.1, but you've got a PR from over a year ago https://github.com/plotly/dash-cytoscape/pull/53 that fixes this issue, saying:
> v0.1.1 broke compatibility with Python 2. Therefore, modified code to be compatible with Python 2.

Seems like there have been no new releases since then? Looks like only [bugfixes and non-code changes since](https://github.com/plotly/dash-cytoscape/compare/v0.1.1...master) so can we make a v0.1.2? There are a bunch of open PRs that would be nice to get into a v0.2 but they may take some time to finish, in the meantime it would be nice to get all the tweaks in the last year released.

xhlulu:
v0.2.0 is going to be released, I don't think this PR is any useful anymore since it's outdated anyway.

Issue #765, opened 2020-01-28
Describe how input argument & function names don't matter


chriddyp:
Maybe it's Point 7 in https://dash.plot.ly/getting-started-part-2

PR #912, opened 2020-06-30
Better Document Decorators

jdamiba:
The purpose of this PR is to better document working with the `app.callback` decorator on the basic callbacks page. 

This PR also fixes up the sidebar on this page with hyperlinks to each section. 

closes https://github.com/plotly/dash-core/issues/113

closes https://github.com/plotly/dash-docs/issues/765

### DIFF
--- a/dash_docs/chapters/basic_callbacks/index.py
+++ b/dash_docs/chapters/basic_callbacks/index.py
@@ -40,7 +40,7 @@ layout = html.Div([
     '''),
 
     html.H4('''
-    Dash App Layout
+    Interactive Dash App Layout
     ''', id='dash-app-layout'),
 
     rc.Markdown(
@@ -55,7 +55,15 @@ layout = html.Div([
     right away. Let's break down what's happening here:
 
     1. The "inputs" and "outputs" of our application interface are described
-       declaratively through the `app.callback` decorator.
+       declaratively as the arguments of the `@app.callback` decorator function.
+    '''),
+
+    rc.Syntax('''Decorator functions are used in Dash to update user interface elements in response to user input. a powerful feature of the Python language.  that allow users to The concept of decorating functions is built into the Python language. functions that wrap other functions in order to modify their behavior. Typically, they are denoted with the `@` prefix. In the context of Dash, they allow users to set up  ''', summary='''
+    Learn more about using decorators in Dash apps.
+    '''),
+
+
+    rc.Markdown('''
     2. In Dash, the inputs and outputs of our application are simply the
        properties of a particular component. In this example,
        our input is the "`value`" property of the component that has the ID
--- a/dash_docs/chapters/basic_callbacks/index.py
+++ b/dash_docs/chapters/basic_callbacks/index.py
@@ -24,8 +24,7 @@ layout = html.Div([
 
     rc.Markdown('''
 
-        In the <dccLink href="/layout" children="previous chapter on
-        `app.layout`"/> we learned that the `app.layout` describes what the app
+        In the <dccLink href="/layout" children="previous chapter"/> we learned that the `app.layout` describes what the app
         looks like and is a hierarchical tree of components.
         The `dash_html_components` library provides classes for all of the HTML
         tags, and the keyword arguments describe the HTML attributes like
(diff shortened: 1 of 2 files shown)


dash_docs/chapters/basic_callbacks/index.py
```diff
@@ -51,25 +50,40 @@
...
+
+        b. You can use any name for the function that is wrapped by the `@app.callback` decorator and for its arguments. However, you must use the same argument name inside the callback function as you do in its definition, just like in a regular Python function.
+
+        c. You must use the same `id` you gave a Dash component in the `app.layout` when referring to it as either an input or output of the `@app.callback` decorator.
+
+        d. The `@app.callback` decorator needs to be directly above the callback function declaration. If there is a blank line between the decorator and the function definition, the callback registration will not be successful.  If you're curious about what the decorator syntax means under the hood, you can read [this StackOverflow answer](https://stackoverflow.com/questions/739654/how-to-make-a-chain-of-function-decorators/1594484#1594484) and learn more about decorators by reading [PEP 318 -- Decorators for Functions and Methods](https://www.python.org/dev/peps/pep-0318/#current-syntax).
```
alexcjohnson, line 68: > If there is a blank line between the decorator and the function definition, the callback registration will not be successful.

Really? Just tested this in py27 and py38 and a blank line seems OK - the linter complains, but the app runs.

dash_docs/chapters/basic_callbacks/index.py
```diff
@@ -51,25 +50,40 @@
...
+    1. The "inputs" and "outputs" of our application's interface are described
+       declaratively as the arguments of the `@app.callback` decorator.
+    '''),
+
+    html.Details(open=False, children=[
+        html.Summary('Learn more about using the `@app.callback` decorator.'),
+        rc.Markdown('''
+        a. By writing this decorator, we're telling Dash to call this function for us whenever the value of the "input" component (the text box) changes in order to update the children of the "output" component on the page (the HTML div).
+
+        b. You can use any name for the function that is wrapped by the `@app.callback` decorator and for its arguments. However, you must use the same argument name inside the callback function as you do in its definition, just like in a regular Python function.
```
alexcjohnson, line 64: Maybe break this into two parts and go into a little more detail, something like:
- You can use any name for the function that is wrapped by the `@app.callback` decorator. The convention is that the name describes the callback output(s).
- You can use any name for the function arguments, but you must use the same names inside the callback function as you do in its definition, just like in a regular Python function. The arguments are positional: first the `Input` items and then any `State` items are given in the same order as in the decorator.

dash_docs/chapters/basic_callbacks/index.py
```diff
@@ -145,7 +163,13 @@
        of how the dataset evolves with time: transitions allow the chart to
        update from one state to the next smoothly, as if it were animated.
 
-    ## Multiple inputs
+    '''),
+
+    html.H4('''
+    Dash App With Multiple Inputs
+    ''', id='dash-app-multiple-inputs'),
```
alexcjohnson, line 170: Is the change to explicit `H4` instead of markdown just to add `id` so we can support scrolling links? If so that's great, I just wonder if we could find a way to do this directly in markdown... that's for another time though, for now I'm just curious your motivation.

jdamiba: You are correct, I switched to explicitly using H4s to support the scrolling sidebar links, as I noticed that they were not working as intended as I was editing the page.  

Seems like currently headings inside of Markdown blocks are not picked up.

dash_docs/chapters/basic_callbacks/index.py
jdamiba: Hmm, that's interesting. To be honest I just assumed it wouldn't work. 

@chriddyp  In the issue you created at https://github.com/plotly/dash-core/issues/113 you explicitly say `this is a very particular python syntax that only works by declaring things directly above functions. if there is a blank line, it won't work`. Do you remember what caused you to have this impression?

dash_docs/chapters/basic_callbacks/index.py
jdamiba: I have updated this part of the explanation with the suggested change.

dash_docs/chapters/basic_callbacks/index.py
jdamiba: @alexcjohnson Is the inclusion of this line blocking this PR from being merged? 

I think it's reasonable to err on the side of caution and explicitly guide users towards using the recommended syntax for decorators even if it is technically possible to include a space.

Issue #724, opened 2019-12-05
Clarify that examples are intended to run in dash


alexcjohnson:
Some of the examples look like they could plausibly run as individual notebook cells (like plotly.py direct usage). We should clarify early on in the docs that all examples are partial or complete dash apps, and in specific cases where we only show the code for a component (like dash-bio, dash-table) we should reiterate this point at the beginning of the section.
https://github.com/plotly/dash-bio/issues/446#issuecomment-562268337

PR #915, opened 2020-07-02
start being explicit about examples being intended to run on command line

jdamiba:
The purpose of this PR is to propose an approach to resolving the issue presented at https://github.com/plotly/dash-docs/issues/724. 

@alexcjohnson If you like this approach I can extend it to the rest of the tutorial in a follow up PR.

### DIFF
--- a/dash_docs/chapters/getting_started/examples/getting_started_core_components.py
+++ b/dash_docs/chapters/getting_started/examples/getting_started_core_components.py
@@ -1,4 +1,8 @@
 # -*- coding: utf-8 -*-
+
+# Run this app with `python app.py` and
+# visit http://127.0.0.1:8050/ in your web browser.
+
 import dash
 import dash_core_components as dcc
 import dash_html_components as html
--- a/dash_docs/chapters/getting_started/examples/getting_started_layout_1.py
+++ b/dash_docs/chapters/getting_started/examples/getting_started_layout_1.py
@@ -1,4 +1,8 @@
 # -*- coding: utf-8 -*-
+
+# Run this app with `python app.py` and
+# visit http://127.0.0.1:8050/ in your web browser.
+
 import dash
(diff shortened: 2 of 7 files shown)


alexcjohnson:
@jdamiba this looks great for the "getting started" page, but later on in the docs I don't think we need _all_ the sample apps to have this extra text & comments. The ones I'm most concerned about are like on https://dash.plotly.com/dash-core-components where it kind of looks like you can just display a single dash component, as if the code example was a notebook cell and the output is the default display of that component, like a jupyter widget:

Screen Shot 2020-07-02 at 5 47 46 PM

This pattern is used in dcc, the html overview, dash-bio, dash-daq... looks like almost all the component docs except maybe table and cytoscape (though I didn't check exhaustively)

I'm thinking maybe the examples can just prepend `app.layout = `, like `app.layout = dcc.Dropdown(...)` with no definition of `app`, just to make it clear there's more to it and this needs to be in an app.

Separately, not for this PR but if we don't have an issue for this we should make one: we should probably have a whole page devoted to dash in jupyter, jupyter lab, colab, etc. I don't see this anywhere, am I missing it?

jdamiba:
I tried the approach you suggested but I don't think that that will work since the component code blocks are actually evaluated.

This means that prepending `app.layout =`  with no definition of `app`, as follows, throws a `NameError`.

```py
rc.ComponentBlock('''import dash_core_components as dcc

app.layout = dcc.Dropdown(
    options=[
        {'label': 'New York City', 'value': 'NYC'},
        {'label': 'Montréal', 'value': 'MTL'},
        {'label': 'San Francisco', 'value': 'SF'}
    ],
    value='MTL'
)''', style=styles.code_container),
```

I suggest merging these changes to the getting-started page while I figure something out for the component pages.

PR #916, opened 2020-07-02
match IDs in basic callback explanation text with IDs in example code

jdamiba:
This PR ensures that the text in this section refers to the same IDs as the example code.


jdamiba:
@alexcjohnson This is a followup PR to https://github.com/plotly/dash-docs/pull/912 as I noticed a mistake.

PR #917, opened 2020-07-03
Create README.md for pdf docs

jdamiba:


alexcjohnson:
I think the idea was to put this into the main README.md, which has up at the top:

Screen Shot 2020-07-06 at 5 29 04 PM

jdamiba:
Thanks for the clarification @alexcjohnson.

I moved the note to the bottom of the main README, removed the link to the PDF docs that was there and deleted the PDF as well.  Please let me know if this works or if you want to keep the file.

README.md
```diff
@@ -46,3 +46,11 @@ on Windows systems `waitress` can be a replacement for `gunicorn`
...
 
 Changes to master will get deployed automatically.
+
+Note: In the past, some corporate networks have blocked our website at https://dash.plot.ly. We created a PDF for those community members.
+
+We have since resolved these issues by moving to https://dash.plotly.com and we’ve stopped maintaining this PDF documentation.
+
+Is your network still blocking our docs and do you still need a PDF version?
+
+If so, please comment on this issue: https://github.com/plotly/dash-docs/issues/897A.
```
alexcjohnson, line 56:
```suggestion
If so, please comment on this issue: https://github.com/plotly/dash-docs/issues/897. 
```

And let's put this note up above where the pdf link used to be - that's the only section of this readme that's intended for viewers, the rest is for contributors.

alexcjohnson:
> and deleted the PDF as well.

💯 no point keeping a big blank file around!

Let's leave #897 open for a while after we merge this PR. Once we're confident we're not going to bring the PDF back, we can remove the note you're adding here and close that issue.

Issue #874, opened 2020-05-04
add prevent_initial_call


alexcjohnson:
https://github.com/plotly/dash/pull/1228 - to be released in Dash 1.12

`@app.callback(prevent_initial_call=True)`

`app.clientside_callback(prevent_initial_call=True)`

`app = Dash(prevent_initial_callbacks=True)` (and then setting `False` in certain callbacks)

nicolaskruchten:
@jdamiba I recommend addressing this at the same time as #852

Issue #852, opened 2020-04-14
Explicitly describe how the callback chain works


chriddyp:
Closely related to #755 

Originally discussed here: https://github.com/plotly/dash/issues/1167

- When anything changes (initial load or later interactions), all the related callbacks are collected, based on both newly-created outputs and newly-changed inputs.
  - On initial load or callbacks returning `children` with new components, none of the inputs are considered to have "changed" unless (1) they're outputs of another callback, or (2) the callback is updating something *outside* the new `children` (not possible for initial load)
  - But the callback will still be queued as an "initial call," unless ALL of its inputs are outputs of something else - then it's no longer considered an "initial call" but it's still put into the queue on the assumption that those outputs will change
- Callbacks are then executed in the order that their inputs are ready.
  - First dispatched are callbacks that have no inputs that are outputs of another callback.
  - As callbacks return, if their outputs were prevented from updating, these are removed as "changed" in any other callbacks they're inputs for. Either way they're marked as no longer blocking the other callback.
  - If a callback has no changed props and is not an "initial call" it's removed from the queue.
  - Then we dispatch any callbacks that are no longer blocked by any of their inputs
  - Repeat until the queue is empty.

Issue #755, opened 2020-01-23
app lifecycle chapter or mention it in the first chapter of callbacks


chriddyp:
explain the order in which things are called and updated:
1. When `python app.py` or `gunicorn app:server` is run, all of the files are "executed". meaning that `df = pd.read_csv('...')` is run when the program starts, rather than when the page is loaded. (so, if the CSV changes, `df` isn't updated until the program restarts. So, users should provide their data via a periodic celery task or by passing it around in a `app.layout` function)
2. `app.layout` is rendered. either the static value or the return of the function
3. `@app.callback` are fired if they exist in particular order. maybe link to a dedicated page on the callback DAG or begin describing how the DAG works here. `None` is passed if initial values aren't provided. mention that you can skip this update with `prevent_initial_call=True`
3. Some components might fire a callback once they get rendered for the first time. like `dcc.Location`. we call these "derived properties". mention which components do this. `dcc.Location.pathname` and associated properties, maybe `DataTable.derived_*` properties? Should verify
5. app waits for user interaction

nicolaskruchten:
@jdamiba I recommend addressing this at the same time as #852

PR #928, opened 2020-07-14
be more explicit about callback chain

jdamiba:
The purpose of this PR is to add a section to the `Advanced Callbacks` page which describes the callback chain in more detail. 

closes https://github.com/plotly/dash-docs/issues/852
closes https://github.com/plotly/dash-docs/issues/755
closes https://github.com/plotly/dash-docs/issues/874


jdamiba:
@alexcjohnson Can you please review this?

dash_docs/chapters/advanced_callbacks/index.py
```diff
@@ -89,5 +89,21 @@ def slow_function(input):
     The [Performance](/performance) section of the Dash docs delves a
     little deeper into leveraging multiple processes and threads in
     conjunction with memoization to further improve performance.
-    ''')
+    '''),
+
+    rc.Markdown('''
+    ## The Callback Chain
+
+    - Upon initial load of a Dash app or based on later interactions, all of the related callbacks are collected, based on both the existence of newly-created outputs and newly-changed inputs.
```
alexcjohnson, line 97: I think we need some examples or other clarification for a number of these points. Ideally I think it'd be nice to have diagrams showing how callbacks can chain, and at least one of these diagrams should be presented alongside explicit dash code - kind of like the callback graphs in the devtools but probably simplified.

> all of the related callbacks are collected

Make it clear that this is recursive: we collect not only those whose inputs changed directly, but also those whose inputs include outputs of another callback we've already collected. It's important that we collect the entire chain up front, or we wouldn't be able to tell for sure which callbacks are blocking others. For example: dropdown A is an input for one callback (X) that sets the value of dropdown B. A second callback (Y) has both dropdown A and B as inputs and outputs to store C. A third callback (Z) uses the store to create a graph D. All three callbacks are collected, but Y is blocked until X completes, and Z is blocked until Y completes.

dash_docs/chapters/advanced_callbacks/index.py
```diff
@@ -89,5 +89,21 @@ def slow_function(input):
...
+      - Upon initial load or in the case of callbacks returning `children` with new compononets, none of the inputs are conisidered to have "changed" unless (1) they're outputs of another callback, or (2) the callback is updating something *outside* the new `children` (not possible upon initial load).
```
alexcjohnson, line 98: > none of the inputs are conisidered to have "changed" unless

"changed" is not only relevant to the logic here, it also determines what the user sees in `callback_context.triggered` - let's point that out and link to the `callback_context` docs.

> (1) they're outputs of another callback

If we walk through an example like the three-callback chain I describe above, we can show exactly what `callback_context.triggered` looks like inside each callback.

> (2) the callback is updating something *outside* the new `children` (not possible upon initial load)

This is a pretty subtle point that probably needs an example all to itself.

dash_docs/chapters/advanced_callbacks/index.py
```diff
@@ -89,5 +89,21 @@ def slow_function(input):
...
+      - But, the callback will still be queued as an "initial call", unless ALL of its outputs are outputs of something else- then it's no longer considered an "initial call", but it's still put into the callback queue on the assumption that those outputs will change.
```
alexcjohnson, line 99: We should mention `prevent_initial_call` here - #874 (may want to describe that in more than one place, or maybe here is sufficient?)

dash_docs/chapters/advanced_callbacks/index.py
```diff
@@ -89,5 +89,21 @@ def slow_function(input):
...
+    - Callbacks are then executed in the order that their inputs are ready.
```
alexcjohnson, line 100: We should note that many server-side callbacks can be requested simultaneously, if they're all unblocked together - and assuming you're running a multi-process server they can execute in parallel.

dash_docs/chapters/advanced_callbacks/index.py
```diff
@@ -89,5 +89,21 @@ def slow_function(input):
...
+      - First dispatched are callbacks that have no inputs that are the outputs of another callback.
+      - As callbacks return, if their outputs were prevented from updateing, these are removed as "changed" in any other callbacks that they are inputs for. Either way, they're marked as no longer blocking the other callback.
```
alexcjohnson, line 102: > prevented from updateing

using `raise PreventUpdate` or `return dash.no_update`

dash_docs/chapters/advanced_callbacks/index.py
```diff
@@ -89,5 +89,21 @@ def slow_function(input):
...
+      - If a callback has no changed props and is not an "initial call", then it is removed from the callback queue.
```
alexcjohnson, line 103: And all of its outputs are also removed as "changed" in other callbacks, possibly resulting in these callbacks being removed as well.

nicolaskruchten:
Notes:

* I recommend addressing #755 and #874 in this same PR, so you can refer to those new sections from the callback-chain stuff
* I recommend stressing under what circumstances the front-end makes distinct HTTP calls the back-end (i.e. once per individual callback firing) and under what circumstances the front-end will issue simultaneous HTTP calls (i.e. once per callback for which the front-end doesn't already know the result will be invalidate by a call it already knows it's got to do)

jdamiba:
@alexcjohnson Can you take another look at this PR? I've reorganized what I had written at `/advanced-callbacks` with some example apps and added a section about the Dash app lifecycle at `/app-lifecycle`.

dash_docs/chapters/advanced_callbacks/index.R
```diff
@@ -94,10 +94,26 @@ The [Performance](/performance) section of the Dash docs delves a
...
 conjunction with memoization to further improve performance.
 "),
-    
+
+dccMarkdown('''
+    ## The Callback Chain
+
+    - Upon initial load of a Dash app or based on later interactions, all of the related callbacks are collected, based on both the existence of newly-created outputs and newly-changed inputs.
+      - Upon initial load or in the case of callbacks returning `children` with new compononets, none of the inputs are conisidered to have "changed" unless (1) they're outputs of another callback, or (2) the callback is updating something *outside* the new `children` (not possible upon initial load).
+      - But, the callback will still be queued as an "initial call", unless ALL of its outputs are outputs of something else- then it's no longer considered an "initial call", but it's still put into the callback queue on the assumption that those outputs will change.
```
alexcjohnson, line 103:
```suggestion
      - But, the callback will still be queued as an "initial call", unless ALL of its inputs are outputs of something else- then it's no longer considered an "initial call", but it's still put into the callback queue on the assumption that those inputs will change.
```

dash_docs/chapters/advanced_callbacks/index.R
```diff
@@ -94,10 +94,26 @@ The [Performance](/performance) section of the Dash docs delves a
...
+    - Callbacks are then executed in the order that their inputs are ready.
+      - First dispatched are callbacks that have no inputs that are the outputs of another callback.
+      - As callbacks return, if their outputs were prevented from updateing, these are removed as "changed" in any other callbacks that they are inputs for. Either way, they're marked as no longer blocking the other callback.
```
alexcjohnson, line 106:
```suggestion
      - As callbacks return, if their outputs were prevented from updating, these are removed as "changed" in any other callbacks that they are inputs for. Either way, they're marked as no longer blocking the other callback.
```

dash_docs/chapters/advanced_callbacks/index.py
```diff
@@ -89,5 +89,193 @@ def slow_function(input):
...
+
+    ### When A Dash App First Loads
+
+    All of the callbacks in a Dash app are executed with the initial value of their inputs when the app is first loaded. This is known as the "initial call" of the callback. To learn how to suppress this behavior, see the documentation for the [`prevent_initial_call`](#prevent-callbacks-from-being-executed-on-initial-load) attribute of Dash callbacks.
+
+    It is important to note that when a Dash app is initially loaded in a web browser by the `dash-renderer` front-end client, its entire callback chain is introspected recursively.
+
+    This allows the `dash-renderer` to predict the order in which callbacks will need to be executed, as callbacks are blocked when their inputs are outputs of other callbacks which have not yet fired. In order to unblock the execution of these callbacks, first callbacks whose inputs are immediately available must be executed. This process helps the `dash-renderer` to minimize the number of network requests it needs to make by making sure it only requests that a callback is executed when all of the callback's inputs have changed.
```
alexcjohnson, line 105:
```suggestion
    This allows the `dash-renderer` to predict the order in which callbacks will need to be executed, as callbacks are blocked when their inputs are outputs of other callbacks which have not yet fired. In order to unblock the execution of these callbacks, first callbacks whose inputs are immediately available must be executed. This process helps the `dash-renderer` to minimize the time and effort it uses, and avoid unnecessarily redrawing the page, by making sure it only requests that a callback is executed when all of the callback's inputs have reached their final values.
```

dash_docs/chapters/advanced_callbacks/index.py
```diff
@@ -89,5 +89,193 @@ def slow_function(input):
...
+)
+
+
+@app.callback(
+    [Output("first_output", "children"), Output("second_output", "children")],
+    [Input("button", "n_clicks")]
+)
+def change_text(n_clicks):
+    if n_clicks is None:
+        return ["n_clicks is None", "n_clicks is None"]
```
alexcjohnson, line 131:
```suggestion
```

`str(None)` is `"None"` so no special handling is needed

dash_docs/chapters/advanced_callbacks/index.py
```diff
@@ -89,5 +89,193 @@ def slow_function(input):
...
+    All of the callbacks in a Dash app are executed with the initial value of their inputs when the app is first loaded. This is known as the "initial call" of the callback. To learn how to suppress this behavior, see the documentation for the [`prevent_initial_call`](#prevent-callbacks-from-being-executed-on-initial-load) attribute of Dash callbacks.
+
+    It is important to note that when a Dash app is initially loaded in a web browser by the `dash-renderer` front-end client, its entire callback chain is introspected recursively.
+
+    This allows the `dash-renderer` to predict the order in which callbacks will need to be executed, as callbacks are blocked when their inputs are outputs of other callbacks which have not yet fired. In order to unblock the execution of these callbacks, first callbacks whose inputs are immediately available must be executed. This process helps the `dash-renderer` to minimize the number of network requests it needs to make by making sure it only requests that a callback is executed when all of the callback's inputs have changed.
+
+    Examine the following Dash app:
+    '''),
+
+    rc.Syntax('''
```
alexcjohnson, line 110: Can we do `rc.Syntax` / `rc.Example` pairs with these new examples, like we do most other places, so users can see the apps in action alongside the code?

dash_docs/chapters/advanced_callbacks/index.py
```diff
@@ -89,5 +89,193 @@ def slow_function(input):
...
+    if n_clicks is None:
+        return ["n_clicks is None", "n_clicks is None"]
+    return ["n_clicks is " + str(n_clicks), "n_clicks is " + str(n_clicks)]
+
+app.run_server(debug=True)
+
+    '''),
+
+    rc.Markdown('''
+    Notice that when this app is finished being loaded by a web browser and ready for user interaction, the `children` of the `html.Div` components is not the text that is defined when the components are declared in the app's layout but rather is the result of the `change_text()` callback being executed. This is because the "initial call" of the callback occured with `n_clicks` having the value of `None`.
```
alexcjohnson, line 139:
```suggestion
    Notice that when this app is finished being loaded by a web browser and ready for user interaction, the `html.Div` components do not say "callback not executed" as declared in the app's layout, but rather "n_clicks is None," the result of the `change_text()` callback being executed. This is because the "initial call" of the callback occurred with `n_clicks` having the value of `None`.
```

dash_docs/chapters/advanced_callbacks/index.py
```diff
@@ -89,5 +89,193 @@ def slow_function(input):
...
+
+    ### As A Direct Result of User Interaction
+
+    Most frequently, callbacks are executed as a direct result of user interaction, such as clicking a button or selecting an item in a dropdown menu. When such interactions occur, Dash components communicate their new values to the `dash-renderer` front-end client, which then requests that the Dash server execute any callback function that has the newly changed value as input.
+
+    If a Dash app has multiple callbacks, the `dash-renderer` batches requests for callbacks to be executed based on whether or not they can be immediately executed with the newly changed inputs. If several inputs change simultaneously, then requests are made to execute them all.
```
alexcjohnson, line 145:
```suggestion
    If a Dash app has multiple callbacks, the `dash-renderer` requests callbacks to be executed based on whether or not they can be immediately executed with the newly changed inputs. If several inputs change simultaneously, then requests are made to execute them all.
```

I wouldn't want to give the impression that we merge multiple callbacks into a single request, which "batches requests" might imply. Someday we might do this, but not today :)

dash_docs/chapters/advanced_callbacks/index.py
```diff
@@ -89,5 +89,193 @@ def slow_function(input):
...
+
+    In this circumstance, it is possible that multiple requests are made to execute the same callback function. This would occur if the callback in question has already been requested and its output returned before the new components which are also its inputs are added to the layout.
+
+    '''),
+
+
+    rc.Markdown('''
+    ## Prevent Callbacks From Being Executed on Initial Load
+
+    You can use the `prevent_initial_call` attribute of callbacks to prevent callbacks from being fired when the app initially loads, as in the following example app.
```
alexcjohnson, line 227: I don't see `prevent_initial_call` below?

dash_docs/chapters/app_lifecycle/index.py
```diff
@@ -0,0 +1,27 @@
+import dash_core_components as dcc
+import dash_html_components as html
+from dash_docs import tools
+from dash_docs import reusable_components as rc
+
+examples = tools.load_examples(__file__)
+
+layout = html.Div(children=[
+    rc.Markdown('''# Dash App Life Cycle
```
alexcjohnson, line 9:
```suggestion
    rc.Markdown('''
    # Dash App Life Cycle
```
To ensure our automatic dedenting of markdown works correctly

dash_docs/chapters/app_lifecycle/index.py
```diff
@@ -0,0 +1,27 @@
...
+
+    This section describes the lifecyle of a Dash app.
+
+    1. When `python app.py` or `gunicorn app:server` is run, all of the files in a Dash app are executed. This means that if you have a statement such as `df = pd.read_csv('...')`, it is run when the program starts, rather than when the page is loaded. Therefore, if the CSV changes after the program starts, the `df` will not be updated until the program restarts. For this reason, it is recommended that users provide their data via a periodic `celery` task or by using a function to generate the `app.layout`.
```
alexcjohnson, line 13:
```suggestion
    1. When `python app.py` or `gunicorn app:server` is run, all of the files in a Dash app are executed. This means that if you have a statement such as `df = pd.read_csv('...')`, it is run when the program starts, rather than when the page is loaded. Therefore, if the CSV changes after the program starts, `df` will not be updated until the program restarts. In this case, it is recommended to provide data via a periodic `celery` task or by setting `app.layout` as a function to regenerate the layout with each page load.
```

dash_docs/chapters/app_lifecycle/index.py
```diff
@@ -0,0 +1,27 @@
...
+
+    2. The `app.layout` is rendered and the app's callback chain is introspected to determine the relationships between inputs, callbacks, and outputs.
+        - Note that this is a recursive process:  the `dash-renderer` collects not only those callbacks whose inputs can be changed directly through user interaction, but also those callbacks whose inputs are outputs of another callback that has already been collected. It's important that the `dash-renderer` collects the entire callback chain up front, or else it wouldn't be able to determine which callbacks are blocking others.
+
+    3. All callbacks that have inputs in currently the app layout are executed. This is known as the "initial call" of the callbacks, and this behavior can be suppressed by using the `prevent_initial_call` attribute.
+        - If a Dash app has multiple callbacks, the `dash-renderer` batches requests for callbacks to be executed based on whether or not they can be immediately executed with the newly changed inputs.
```
alexcjohnson, line 19:
```suggestion
        - If a Dash app has multiple callbacks, the `dash-renderer` requests callbacks to be executed based on whether or not their inputs might be changed as a result of another callback.
```

jdamiba:
@alexcjohnson I have updated this PR with your suggested commits as well and used the `rc.Syntax/rc.Example` workflow to interactively embed the example Dash apps. 

Please note that while I have updated the text in the R `Advanced Callbacks` chapter to match the Python text, I have not yet created R versions of the example apps or created an R version of the `Dash App Lifecycle` chapter.

alexcjohnson:
> I have updated this PR with your suggested commits as well and used the rc.Syntax/rc.Example workflow

Fantastic, love it!

> Please note that while I have updated the text in the R Advanced Callbacks chapter to match the Python text, I have not yet created R versions of the example apps or created an R version of the Dash App Lifecycle chapter.

OK great, glad that's on your radar. It's looking really good, but I'll hold off a final review until the rest of the R pieces are in place.

Not to be implemented in this PR, but I noticed this and started wondering if we should try to DRY some of this across languages -> #950

jdamiba:
@alexcjohnson This PR is ready for your review- I have made R versions of the example apps and put them in the appropriate place, and have created a "Dash App Lifecycle" chapter for the R version as well. 

Please note that since https://github.com/plotly/dashR/issues/227 is outstanding, I cannot document that feature on the R side.

dash_docs/chapters/advanced_callbacks/index.py
```diff
@@ -89,5 +89,72 @@ def slow_function(input):
     The [Performance](/performance) section of the Dash docs delves a
     little deeper into leveraging multiple processes and threads in
     conjunction with memoization to further improve performance.
-    ''')
+    '''),
+
+    rc.Markdown('''
+    ## When Are Callbacks Executed?
+
+    This section describes the circumstances under which the `dash-renderer` front-end client can make a request to the Dash back-end server to execute a callback function.
```
alexcjohnson, line 97:
```suggestion
    This section describes the circumstances under which the `dash-renderer` front-end client can make a request to the Dash back-end server (or the clientside callback code) to execute a callback function.
```
just to make it clear that all of this applies exactly the same way to clientside callbacks.

dash_docs/chapters/app_lifecycle/index.py
```diff
@@ -0,0 +1,28 @@
...
+
+layout = html.Div(children=[
+    rc.Markdown('''
+    # Dash App Life Cycle
+
+    This section describes the lifecyle of a Dash app.
+
+    1. When `python app.py` or `gunicorn app:server` is run, all of the files in a Dash app are executed. This means that if you have a statement such as `df = pd.read_csv('...')`, it is run when the program starts, rather than when the page is loaded. Therefore, if the CSV changes after the program starts, `df` will not be updated until the program restarts. In this case, it is recommended to provide data via a periodic `celery` task or by setting `app.layout` as a function to regenerate the layout with each page load.
+
+    2. The `app.layout` is rendered and the app's callback chain is introspected to determine the relationships between inputs, callbacks, and outputs.
```
alexcjohnson, line 16: Step 1 is happening entirely on the server, but step 2 is in the browser without describing how we got there. Maybe we want a step between like:

2 - When the page loads and `dash-renderer` starts, it asks the server for both the initial layout (`app.layout`) and a list of all the callbacks registered to the app. This list of callbacks is only provided at page load and can't be modified later, so the app needs to know about all of its callbacks from the beginning even for components that do not exist in the initial layout.

alexcjohnson (approved):
@jdamiba this looks great! I just made 2 small comments (commented on the Python side but they apply to both Py and R) and then I think this is ready! Nice job on the `prevent_initial_call` part, and you handled the Py/R difference there well.

PR #959, opened 2020-09-09
[WIP] New performance section

chriddyp:
http://dash.plotly.com/performance hasn't been updated in a couple years. We've learned a lot since then!

This is currently a work-in-progress. I've committed an outline of how these sections might be organized. More to come soon.


chriddyp:
stale

PR #962, opened 2020-09-12
Added GTM to index string body and header

### LINKED ISSUE #945: add GTM to dash R docs and create GA view for it
discovered as part of: https://github.com/plotly/marketing-team/issues/288

HammadTheOne:
This PR address #945  and adds GTM tags to the Dash R docs. 

Closes #945 

Post-merge checklist:

- [x] I understand

### DIFF
--- a/app.R
+++ b/app.R
@@ -7,4 +7,35 @@ metas <- list(
 
 app <- Dash$new(assets_folder="dash_docs/assets", meta_tags = metas)
 
+index_string <-
+  "<!DOCTYPE html>
+        <html>
+          <head>
+            {%meta_tags%}
+            <title>{%private$name%}</title>
+            {%favicon%}
+            {%css_tags%}
+            <script>
+              (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+              new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+              j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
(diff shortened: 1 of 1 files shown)


app.R
```diff
@@ -7,4 +7,35 @@ metas <- list(
...
 
+index_string <-
+  "<!DOCTYPE html>
+        <html>
+          <head>
+            {%meta_tags%}
+            <title>{%private$name%}</title>
+            {%favicon%}
+            {%css_tags%}
+            <script>
```
rpkyle, line 18: @HammadTheOne Not sure if we need the anchors (probably splitting hairs here 😅 ), but I see them here in @VeraZab's instructions and not in this commit:
Screen Shot 2020-08-21 at 2 45 10 PM

rpkyle (approved):
Looks good to me overall. This reminds me that we have no unit or integration tests for the `dash-docs`, as we do with Python:

https://github.com/plotly/dash-docs/tree/master/tests/unit

Probably something to tackle eventually. Thanks for adding the GTM tags! If @VeraZab thinks this looks good, I say 💃

app.R
VeraZab: was going to give this a tests once the new, dashr only pages, are deployed (as discussed on slack: https://plotly.slack.com/archives/C07MS2763/p1600174658002900) but I think it's probably simplest/fastest to merge this in, and then I'll almost immediately see if anything's off in analytics, and we could correct.

> Not sure if we need the anchors
not quite sure which anchors we're talking about. do you mean those `<!--` ? It's just inline comments that are not really necessary. So I think we're good.

rpkyle: 👍

PR #964, opened 2020-09-15
Add server routes and redirects section to Dash for R docs

rpkyle:
This PR proposes to contribute a server routes and redirects section, following discussion with @alexcjohnson and @Marc-Andre-Rivet about the ideal location to add these details. 

The documentation contributes examples and reflects new functionality merged in plotly/dashR#225, namely regarding the use of the `server_route` and `redirect` methods.


dash_docs/chapters/routes/index.R
```diff
@@ -0,0 +1,134 @@
...
+  dccMarkdown("
+As discussed in the [`Deployment Chapter`](deployment), 
+Dash for R uses the Fiery web framework under the hood.
+Fiery leverages the `routr` package to handle routing,
+but with an interface that may feel unfamiliar to Dash
+developers.
+
+To make it straightforward to specify user-defined routes
+and redirects in your Dash app, we've added some convenience
+methods that are more user-friendly.
```
alexcjohnson, line 19:
```suggestion
Fiery leverages the `routr` package to handle routing.
While it's possible to access this directly via
`app$server$attach` and `server$plugins$request_routr`, 
we've added some convenience methods that are more
user-friendly for redirects and other user-defined routes.
```

dash_docs/chapters/routes/index.R
```diff
@@ -0,0 +1,134 @@
...
+
+The `server_route` method takes three arguments: a `path`
+which references the URL for which routing should be performed,
+a `handler` function that instructs Dash how to handle requests
+for that path, and (optionally) the `methods` supported for
+the request. If you don't specify `methods`, Dash will assume
+that the desired method is `get`.
+
+The URL path is comprised of strings, parameters (strings
```
alexcjohnson, line 28:
```suggestion
The URL path is composed of strings, parameters (strings
```

dash_docs/chapters/routes/index.R
```diff
@@ -0,0 +1,134 @@
...
+
+    app$server_route('/getting-started', function(request, response, keys, ...) {
+      response$status <- 307L
+      response$set_header('Location', '/layout')
+      TRUE
+    })
+```
+
+
+We could also incorporate a wildcard to handle subpaths:
```
alexcjohnson, line 55:
```suggestion
We could also incorporate a wildcard to handle subpaths -
in this case redirecting `/getting-started/` with anything
after it to just `/layout`:
```

dash_docs/chapters/routes/index.R
```diff
@@ -0,0 +1,134 @@
...
+    library(dash)
+    library(dashCoreComponents)
+    library(dashHtmlComponents)
+
+    app <- Dash$new()
+
+    app$server_route('/getting-started', function(request, response, keys, ...) {
+      response$status <- 307L
+      response$set_header('Location', '/layout')
+      TRUE
```
alexcjohnson, line 50: Is it obvious to users what `TRUE` does here? Do you always return `TRUE` or is that the response body, that's just empty in this case because it's a redirect?

dash_docs/chapters/routes/index.R
```diff
@@ -0,0 +1,134 @@
...
+
+---
+
+## Adding redirects to your Dash application
+
+While the `server_route` method is powerful and easy to
+use, if you're just looking to add a few redirects from
+one path to another, it can feel a bit verbose. Dash also
+supports a `redirect` method that makes it even simpler
+to add permanent redirects to your app.
```
alexcjohnson, line 91:
```suggestion
use, if you're just looking to add permanent redirects from
one path to another, Dash also supports a simple `redirect`
method.
```

alexcjohnson:
There are some redirects in the old style at the end of `run.R` - can we replace these with the new syntax?

dash_docs/chapters/routes/index.R
```diff
@@ -0,0 +1,134 @@
...
+
+Like `server_route`, the `redirect` method takes three
+arguments. The `old_path` argument specifies the path to
+redirect, while `new_path` accepts either the path to
+which requests should be rerouted or a function that is
+used to construct it, and `methods` works exactly as
+above (with `get` as its default option).
+
+For a simple path-to-path redirect, the syntax is as
```
HammadTheOne, line 100:
```suggestion
For a simple path-to-path redirect, this syntax is as
```

dash_docs/chapters/routes/index.R
```diff
@@ -0,0 +1,134 @@
...
+concise as it gets. Here we redirect all traffic to
```
HammadTheOne, line 101:
```suggestion
concise as it gets. Here we redirect all traffic from
```

dash_docs/chapters/routes/index.R
```diff
@@ -0,0 +1,134 @@
...
+library(dashCoreComponents)
+library(dashHtmlComponents)
+
+utils <- new.env()
+source('dash_docs/utils.R', local=utils)
+
+layout <- htmlDiv(list(
+  htmlH1('Server Routes and Redirects'),
+  dccMarkdown("
+As discussed in the [`Deployment Chapter`](deployment),
```
HammadTheOne, line 11:
```suggestion
As covered in the [`Deployment Chapter`](deployment), 
```

HammadTheOne:
Looks pretty good to me and I was able to run the examples and use redirects in my own test app with both methods by following the docs directly. 

One issue I ran into, and I'm not certain if it's within the scope of this PR or an issue with Dash R itself, but while the docs run fine with the `master` version of Dash R, with the `dev` version I'm noticing this error message when sourcing the `devtools` and `external_resources` chapters, which I'm guessing is relating to plotly/dashr#225. This prevents the docs from running. I can open a separate issue for that if we need.

image

Other than that, it looks good to me.

dash_docs/chapters/routes/index.R
rpkyle: This is idiosyncratic to `routr`, and has to do with whether the route should be the last in the stack to be called:

> Lastly, each handler must return either `TRUE` indicating that the next route should be called, or `FALSE` indicating no further routes should be called.

It's described in greater detail [here](https://www.data-imaginist.com/2017/introducing-routr/), but I can add a sentence explaining why this is needed if it would be helpful.

rpkyle:
> There are some redirects in the old style at the end of `run.R` - can we replace these with the new syntax?

Yes, of course, but right now there's an issue with the docs that needs to be fixed before we can upgrade `dash` to the new version.

I can edit `run.R` in this PR to use the new syntax, then open a new PR to fix the issues, and merge the latter before the former.

dash_docs/chapters/routes/index.R
alexcjohnson: Given that we're trying to shield users from `routr`, yes we should document this and anything else that requires knowledge of it.

In the future I wonder if we could wrap the handler to add the `TRUE` automatically - if there's never a reason for users to return anything else.

alexcjohnson:
> and merge the latter before the former.

Your call - either fix the updating issues first, or make a new issue or comment on the relevant issue on updating `dash`. Doesn't need to be in this PR but we don't want to forget to update the redirects eventually.

dash_docs/chapters/routes/index.R
alexcjohnson: Or at least wrap it so `TRUE` is the default return, and if you want chaining behavior you have to explicitly return `FALSE`

rpkyle: @alexcjohnson Thinking about this a bit more, it seems tricky to permit the case where a user might want to conditionally return `TRUE` or `FALSE` based on the specifics of the request (e.g. which URL path is being resolved).

If we're not concerned about that, I could check to see if the final line of the function body is `TRUE` or `FALSE`, and append `TRUE` if it's neither of these things, but if there is a legitimate use case where the developer wants to stop iterating over the route stack based on a parameter, it could get messy.

alexcjohnson: Oh I wouldn't want to try and inspect the function body, I was just thinking of a wrapper that would look at the return value, like:
```
function handlerWrapper(request, response, keys, ...) {
  retVal <- handler(request, response, keys, ...)
  if (is.null(retVal)) {
    return(TRUE)
  }
  else {
    return(retVal)
  }
}
```
Wouldn't something like that work? Anyway not something to be done now, and as it's a relatively minor improvement even if it works I wouldn't call this high priority. Just need to make sure we document this, along with a link to where they can find out more info about constructing these handlers - what attributes & methods the `response` object has, what `request` and `keys` look like... in fact, I guess there's something to be said for keeping this to exactly the `routr` handler format (including requiring the boolean return value) so that users can plug in any examples they come across verbatim.

Marc-Andre-Rivet: I think this is an interesting simplification for lots of scenarios but out of scope for documenting the current version of DashR. @rpkyle could you open an issue to document the suggestion. If/when we do this for DashRR in the future, it might be interesting to consider what if any equivalent simplification might be possible in DashPy/Flask too.

Marc-Andre-Rivet:
@rpkyle For https://github.com/plotly/dash-docs/pull/964#issuecomment-694322566, since the DashR version has been updated, could we update the redirect style as part of this PR now?

dash_docs/chapters/routes/index.R
rpkyle: Yes, agree completely. Opened an issue: https://github.com/plotly/dashR/issues/238.

rpkyle:
> There are some redirects in the old style at the end of `run.R` - can we replace these with the new syntax?

I'm going to go ahead and do that now before merging, no reason not to at this point.

Marc-Andre-Rivet (approved):
💃 @rpkyle Provided the above comment is addressed

rpkyle:
> One issue I ran into, and I'm not certain if it's within the scope of this PR or an issue with Dash R itself, but while the docs run fine with the `master` version of Dash R, with the `dev` version I'm noticing this error message when sourcing the `devtools` and `external_resources` chapters, which I'm guessing is relating to [plotly/dashR#225](https://github.com/plotly/dashR/pull/225). This prevents the docs from running. I can open a separate issue for that if we need.

I believe I've fixed this upstream in #1010, so the docs should run now with those fixes.

rpkyle:
> There are some redirects in the old style at the end of `run.R` - can we replace these with the new syntax?

fixed in c378f5e

dash_docs/chapters/routes/index.R
rpkyle: > Given that we're trying to shield users from `routr`, yes we should document this and anything else that requires knowledge of it.
> 
> In the future I wonder if we could wrap the handler to add the `TRUE` automatically - if there's never a reason for users to return anything else.

added in 8a3bef45

Issue #263, opened 2018-11-09
DashR Docs


nicolaskruchten:
Placeholder

rpkyle:
Following discussion with @chriddyp, need to begin assembling Dash Docs written in DashR and served via DDS. Draft plan (slightly more advanced placeholder, subject to change):

- Aim for release by mid-February
- Manually create DashR docs that are very close to the content and level of
detail and quality found in current Dash for Python online documentation
- Begin by creating index page with link to Part 3 chapter on callbacks, then draft that content first.
- In essence, successful documentation will comprise four pieces:

- [ ] Multipage app
- [ ] Handwritten copy
- [ ] Autogenerated examples, which execute R code live
- [ ] Live interactive demos derived from that R code

rpkyle:
Closing as the intitial documentation is available online at https://dashr.plot.ly.

PR #966, opened 2020-09-16
Add demo button in banner 

tobinngo:
Added Dash Enterprise demo button in Dash documentation banner as per dash-core [#263](https://github.com/plotly/dash-core/issues/263).

Also included right-hand sidebar word-wrap and max-width as per dash-enterprise-docs [#540](https://github.com/plotly/dash-enterprise-docs/issues/540).

### Before
image

### After
banner_button_3

### Before 
banner_button_3_resizing

### After
banner_button_4_resizing

### DIFF
--- a/dash_docs/assets/override.css
+++ b/dash_docs/assets/override.css
@@ -101,6 +101,7 @@ h6 { font-size: 20px; line-height: 1.6;  margin-bottom: 0.75rem; margin-top: 0.7
     border-bottom-width: 1px;
     cursor: pointer;
     opacity: 0.5;
+    word-wrap: break-word;
 }
 .page-menu--link:hover {
     border-bottom-style: solid;
--- a/dash_docs/assets/override.css
+++ b/dash_docs/assets/override.css
@@ -479,14 +479,23 @@ a {
     }
 }
 
-@media (max-width: 800px) {
-    .header .links iframe {
+@media (max-width: 1000px) {
+    .header .links a.links--announcements {
+        display: none;
+    }
+    .header .links a.links--gallery {
+        display: none;
+    }
+    .header .links a.links--show-and-tell {
+        display: none;
+    }
+    .header .links a.links--community-forum {
         display: none;
     }
 }
 
-@media (max-width: 700px) {
-    .header .links {
+@media (max-width: 550px) {
+    .header .links Iframe {
         display: none;
     }
 }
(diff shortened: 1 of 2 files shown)


dash_docs/run.py
```diff
@@ -47,7 +47,7 @@ def create_contents(contents):
                 ), href='/', id='logo-home'),
             ], className='logo'),
 
-            # HEADS UP!
+            # HEADS UP! 
             # If you are modifying these header links,
             # make sure to check that the responsive design still works
             # The breakpoints are set in override.css
```
chriddyp, line 53: Could you check the responsive design as per this comment? We might need to adjust the breakpoint width at which point we hide the header links

chriddyp: basically make the screen width smaller and inspect it manually

tobinngo: Complete:

chriddyp: I don’t see any changes in the code or breakpoints. Did the existing breakpoints work OK with this new button in the header?

tobinngo:
The existing breakpoints do work, but they are not very smooth. The button in the header works for me. It is what Jiny asked for.

chriddyp:
On smaller screen sizes, the multi-line headers look broken to me.
image

 I'd recommend increasing the breakpoint width so that the button or the links are hidden at this point. 

That is, change `705px` here to be more like `800px`:
https://github.com/plotly/dash-docs/blob/847f9beb0ef5ce9b49f22bc1589072d17a937662/dash_docs/assets/override.css#L482-L492

tobinngo:
I'll make those changes.

jingningzhang1:
Fantastic. Thanks @tobinngo!

chriddyp:
I'm still seeing the buttons and links move onto multiple lines rather than becoming hidden. We might need to play around a bit more with those media query breakpoints

resizing

dash_docs/assets/override.css
```diff
@@ -443,16 +444,19 @@ a {
   }
 }
 
-@media (max-width: 1150px) {
+/* @media (max-width: 1150px) {
     .page-menu {
         display: none;
     }
-}
+} */
```
chriddyp, line 451: We can just 🔪 out code instead of commenting out. The old code will always be available in git history if we want to refer to it later 🙂

chriddyp:
💃 This looks good to me! I just sent an email invite for you to join Percy, a service we use for visual tests & diffs. Once you've joined, click on the "Details" button in the tests block above, inspect the diffs, and approve them. Once approved, you're good to merge!

chriddyp (approved):
💃

PR #974, opened 2020-09-24
First few chapters of Dash.jl tutorial

jdamiba:
closes https://github.com/plotly/dash-core/issues/216

items outstanding:

- [x] generic crossfilter recipe app 

- [ ] styling through reusable components

- [ ] embedded interactive examples

- [ ] deployment script


dash_docs/chapters/basic_callbacks/examples/multi-inputs.jl
```diff
@@ -0,0 +1,95 @@
+using CSV, DataFrames, Dash, DashHtmlComponents, DashCoreComponents, PlotlyJS
+
+
+url = "https://plotly.github.io/datasets/country_indicators.csv"
+download(url, "country-indicators.csv")
+
+df = DataFrame(CSV.File("country-indicators.csv"))
+
+dropmissing!(df)
```
rpkyle, line 9: Interesting, do we do this on the R and Python side as well? I didn't see similar code there. It looks like the Python filename is `getting_started_multiple_viz.py`, but the R and Julia are ones are both `multi-inputs.*`. 

Might be something worth changing so all three have the same structure, not sure.

dash_docs/chapters/basic_callbacks/examples/multi-inputs.jl
```diff
@@ -0,0 +1,95 @@
...
+    ),
+    dcc_graph(id = "indicator-graphic"),
+    dcc_slider(
+        id = "year-slider",
+        min = minimum(years),
+        max = maximum(years),
+        marks = Dict([Symbol(v) => Symbol(v) for v in years]),
+        value = minimum(years),
+        step = nothing,
+    )
```
rpkyle, line 59: I noticed that the Julia and Python docs have `step = nothing` or `step = None` here, but the R docs don't pass `step = NULL`. I think this is because the latter is true by default in R, but in Julia, I believe there's no default value either:

```julia
function dcc_slider(; kwargs...)
        available_props = Symbol[:id, :marks, :value, :className, :disabled, :dots, :included, :min, :max, :tooltip, :step, :vertical, :verticalHeight, :updatemode, :loading_state, :persistence, :persisted_props, :persistence_type]
        wild_props = Symbol[]
        return Component("dcc_slider", "Slider", "dash_core_components", available_props, wild_props; kwargs...)
end
```

What happens when we omit `step = nothing` from the code above? Does the example fail to render correctly?

dash_docs/chapters/getting_started/examples/getting_started_layout_2.jl
```diff
@@ -0,0 +1,43 @@
+using Dash, DashHtmlComponents, DashCoreComponents
+
+
+app = dash()
+
+app.layout = html_div(style = Dict("backgroundColor" => "#111111")) do
```
rpkyle, line 6: It's optional, but I think we could also do instead

```julia
app.layout = html_div(style = ("backgroundColor" => "#1111111",))
```

dash_docs/chapters/getting_started/index.jl
```diff
@@ -0,0 +1,321 @@
...
+    dcc_markdown("""
+    Dash apps are composed of two parts. The first part is the `layout` of the app and it describes what the application looks like.
+    The second part describes the interactivity of the application and will be covered in the [next chapter](/basic-callbacks).
+
+    Dash provides Julia methods for all of the visual components of the application.
+    We maintain a set of components in the `DashCoreComponents` and the `DashHtmlComponents` packages
+    but you can also build your own with JavaScript and React.js
+
+    Note: Throughout this documentation, Julia code examples are meant to be saved as files and executed using `julia app.jl`.
+    These examples are not intented to run in Jupyter notebooks as-is,
```
rpkyle, line 25:
```suggestion
    These examples are not intended to run in Jupyter notebooks as-is,
```

dash_docs/chapters/getting_started/index.jl
```diff
@@ -0,0 +1,321 @@
...
+    ```
+    \$ julia app.jl
+    ```
+
+    In the `julia` REPL you can run the follwing code, assuming that `app.jl` is in your current working directory.
+    ```
+    julia> include("app.jl")
+    ```
+
+    You can visit the app by visiting the URL http://127.0.0.1:8000.
```
rpkyle, line 69:
```suggestion
    You can visit the app by visiting the URL http://127.0.0.1:8050.
```

dash_docs/chapters/getting_started/index.jl
```diff
@@ -0,0 +1,321 @@
...
+                data = [
+                    (x = ["giraffes", "orangutans", "monkeys"], y = [20, 14, 23], type = "bar", name = "SF"),
+                    (x = ["giraffes", "orangutans", "monkeys"], y = [12, 18, 29], type = "bar", name = "Montréal"),
+                ],
+                layout = (title = "Dash Data Visualization", barmode="group")
+            )
+        )
+    end
+
+    run_server(app, "0.0.0.0", 8000, debug=true)
```
rpkyle, line 53: For parity with the instructions that follow below (using `127.0.0.1` instead of `0.0.0.0`):

```suggestion
    run_server(app, debug=true)
```

dash_docs/chapters/getting_started/index.jl
```diff
@@ -0,0 +1,321 @@
...
+    """),
+
+    dcc_markdown("""
+
+    Run the app with
+    ```
+    \$ julia app.jl
+    ```
+
+    In the `julia` REPL you can run the follwing code, assuming that `app.jl` is in your current working directory.
```
rpkyle, line 64:
```suggestion
    In the `julia` REPL you can run the following code, assuming that `app.jl` is in your current working directory.
```

dash_docs/chapters/getting_started/index.jl
```diff
@@ -0,0 +1,321 @@
...
+
+    Note:
+
+    1. The `layout` is composed of a tree of "components" like `html_div` and `dcc_graph`.
+    2. The `DashHtmlComponents` package has a component for every HTML tag. The `html_h1("Hello Dash")` component
+    generates a `<h1>Hello Dash</h1>` HTML element in your application.
+    3. Not all components are pure HTML. The `DashCoreComponents` package describes higher level components
+    that are interactive and genrated with JavaScript, HTML, and CSS through the React.js library.
+    4. Each component is described entirely through keyword attributes. Dash is *declarative*: you will
+    priimarily describe your application through these attributes.
```
rpkyle, line 79:
```suggestion
    primarily describe your application through these attributes.
```

dash_docs/chapters/getting_started/index.jl
```diff
@@ -0,0 +1,321 @@
...
+    dcc_markdown("""
+
+    In this example, we modified the inline styles of the `html_div` and
+    `html_h1` components with the style property. `html_h1("Hello Dash",
+    style = Dict("color" => "#7FDBFF", "textAlign" => "center"))` is rendered
+    in the Dash application as `<h1 style="text-align: center; color: rgb(127, 219, 255);">Hello Dash</h1>`
+
+    There are a few important differences between the `DashHtmlComponents` and HTML attributes:
+
+    1. The `style` property in HTML is a semicolon separated string. In Dash, you can just suppy a `Dict`.
```
rpkyle, line 153:
```suggestion
    1. The `style` property in HTML is a semicolon separated string. In Dash, you can just supply a `Dict`.
```

dash_docs/chapters/getting_started/index.jl
```diff
@@ -0,0 +1,321 @@
...
+    2. The keys in the `style` `Dict` are [camelCased]("https://en.wikipedia.org/wiki/Camel_case"). So instead of `text-align`, it's `textAlign`.
+    3. The HTML `class` attribute is `className` in Dash.
+    4. The children of the HTML tag is specified through the `children` keywork argument. By convention, this is
```
rpkyle, line 156:
```suggestion
    4. The children of the HTML tag is specified through the `children` keyword argument. By convention, this is
```

dash_docs/chapters/getting_started/index.jl
```diff
@@ -0,0 +1,321 @@
...
+    dcc_markdown("""
+    The `DashCoreComponents` package includes a component called `graph`.
+
+    `graph` renders interactive data visualizations using the open source
+    plotly.js JavaScript graphing library. Plotly.js supports over 35 chart types
+    and renders charts in both vector-quality SVG and high-performance WebGL.
+
+    The `figure` argument in the `dcc_graph` component is the same `figure` argument
+    that is used by `plotly.py`, Plotly's open-source Python graphing library. Check out the
+    plotly.py documentatioin and gallery to learn more.
```
rpkyle, line 210: This section feels a bit weird here, since we're mostly writing about Julia stuff. Not sure if we can either tie this in to PlotlyJS.jl or remove entirely, but should probably not keep the `plotly.py` reference as-is.

```suggestion
    The `figure` argument in the `dcc_graph` component is the same `figure` argument
    that is used by `plotly.py`, Plotly's open-source Python graphing library. Check out the
    plotly.py documentation and gallery to learn more.
```

dash_docs/chapters/getting_started/index.jl
```diff
@@ -0,0 +1,321 @@
...
+
+    run_server(app, "0.0.0.0", 8000, debug=true)
+
+    ```
+    """),
+    html_h1("Core Components"),
+    dcc_markdown("""
+    The `DashCoreComponents` package includes a set of higher level components like
+    dropdowns, graphs, markdown blocks, and more. Like all Dash components, they are described
+    entirely declaratively. Every optiion that is configuragble is available as a keyword
```
rpkyle, line 281:
```suggestion
    entirely declaratively. Every option that is configurable is available as a keyword
```

dash_docs/chapters/installation/index.jl
```diff
@@ -0,0 +1,53 @@
...
+
+app = dash()
+
+app.layout = html_div() do
+    dcc_markdown("""
+    ## Dash Installation
+
+    In order to use Dash, please ensure that you are using a version of Julia >= 1.2.
+
+    To install the most recently version:
```
rpkyle, line 12:
```suggestion
    To install the most recent version:
```

dash_docs/chapters/sharing_data/index.jl
```diff
@@ -0,0 +1,296 @@
...
+    ### Why Share State?
+
+    In some apps, you may have multiple callbacks that depend on expensive data processing tasks like
+    making SQL queries, running simulations, or downloading data.
+
+    Rather than have each callback run the same expensive task, you can have one callback run the task and
+    then share the results with the rest of the callbacks.
+
+    This need has been somewhat ameliorated now that you can have [multiple outputs](/basic-callbacks) for one callback.
+    This way, that expensive task can be done once and immediately outsed in all the outputs. But in some cases this
```
rpkyle, line 30:
```suggestion
    This way, that expensive task can be done once and immediately used in all the outputs. But in some cases this
```

rpkyle (commented):
Just some initial comments -- overall things are looking good, it's really coming along nicely. One thing it looks like we still need to resolve is the mechanism for tying the different chapters together. 

In R we have `app.R` and `run.R`, which bundles everything together, but I think we've yet to add something similar for Dash.jl (and I believe we'll also need this before we can deploy the docs to Heroku).

Marc-Andre-Rivet (commented):
@jdamiba Here are the things I found, either for immediate attention or issue creation + adding to the 4.2 project board

- missing `favicon` (would fix this now)

- missing syntax highlighting for Julia code (would fix this now)
--> requires overriding the default highlights settings for dcc/Markdown. To do that, generate a highlight.js package that includes Julia alongside the standard inclusions for dcc (https://github.com/plotly/dash-core-components/blob/dev/src/third-party/highlight.js, https://github.com/plotly/dash-table/blob/dev/src/third-party/highlight.js) and put it in the assets folder of the docs
--> create an issue in DCC and dash-table to add Julia to the list of defaults we load

- no logo vs. logo-home differentiation for Plotly vs. docs root (DashR docs also doesn't do that at the moment so I think this is fine for now

- misaligned styles (should be trivial, in which case do it now, if complicated for some reason, create an issue)
Julia: 
image

R / Python
image

rpkyle:
> @jdamiba Here are the things I found, either for immediate attention or issue creation + adding to the 4.2 project board
> 
>     * missing `favicon` (would fix this now)

I believe the `favicon.ico` is already in the assets folder (and other assets are loading properly); if true, this may be related to handling of favicons in the backend rather than missing resources.

>     * missing syntax highlighting for Julia code (would fix this now)
>       --> requires overriding the default highlights settings for dcc/Markdown. To do that, generate a highlight.js package that includes Julia alongside the standard inclusions for dcc (https://github.com/plotly/dash-core-components/blob/dev/src/third-party/highlight.js, https://github.com/plotly/dash-table/blob/dev/src/third-party/highlight.js) and put it in the assets folder of the docs
>       --> create an issue in DCC and dash-table to add Julia to the list of defaults we load

Just another data point: I believe `highlight.js` already has support for Julia syntax highlighting, but it's unclear to me why it's not functioning as expected:

https://github.com/highlightjs/highlight.js/blob/master/src/languages/julia.js

rpkyle:
@jdamiba Just cross-referencing here in case it can be resolved in parallel: https://github.com/plotly/Dash.jl/issues/73.

jdamiba:
@rpkyle That's a really interesting issue- I ran into the same problem during the development of that example, and Alexandr made a PR to fix the underlying issue - https://github.com/plotly/Dash.jl/pull/65. So theoretically as of Dash.jl 0.1.2, this problem should not be affecting users.

Marc-Andre-Rivet:
@rpkyle for https://github.com/plotly/dash-docs/pull/974#issuecomment-712235977 - we load language resources on a per-need basis - highlight.js is pretty big if you load all languages, so we've been limiting ourselves to those we support in Dash and a set that made sense, Julia wasn't part of that list, the links above show the setup for DCC and Table.

jdamiba:
@Marc-Andre-Rivet @alexcjohnson  

I've figured out Julia code-highlighting in https://github.com/plotly/dash-docs/commit/386330ecdf18227ef13df47892fa6aa0cdceef23. 

Screen Shot 2020-10-20 at 8 19 09 AM

If you think this looks good, I can deploy these same changes to Heroku today.

Marc-Andre-Rivet:
@jdamiba https://github.com/plotly/dash-docs/pull/974#issuecomment-712810614 👍 Once this is released in Dash (dcc+table: https://github.com/plotly/dash-core-components/issues/870 and https://github.com/plotly/dash-table/pull/841) we'll want to rollback the docs patch.

Marc-Andre-Rivet (approved):
:shipit: 🚢

PR #980, opened 2020-10-01
Refactoring of LoadExampleCode

waralex:
`LoadExampleCode` now parses example into the`Expr` (AST tree) before executing it. In the resulting tree, all elements are evaled, except for the following:

- `app = ...` and `run_server(...)` ignored
- `app.layout = ...` executed separately and returned as `layout` field in the result of the `LoadExampleCode`
- `callback!(...)`  added to the `callback!(app)` function of the `LoadExampleCode` result. Calling this function with a specific application will accept all callbacks that are encountered in the example  to that application

All the checks above only work at the top level of the code with the example. I.e. there is no recursive traversal of submodules and functions.

### DIFF
--- a/dash_docs/chapters/basic_callbacks/index.jl
+++ b/dash_docs/chapters/basic_callbacks/index.jl
@@ -2,9 +2,8 @@ using Dash
 using DashHtmlComponents
 using DashCoreComponents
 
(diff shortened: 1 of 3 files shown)


rpkyle:
🎉 @waralex Thanks Alexandr! I think @jdamiba has already merged in your edits and mine, so I'll close both our PRs for housekeeping reasons.

Issue #965, opened 2020-09-15
An Ecosystem Section in the docs


chriddyp:
How Dash can be used with other libraries in the Python ecosystem. Including pandas, vaex, arrow, scikit-image, holoviews, scikit-learn, etc

nicolaskruchten:
geopandas

PR #981, opened 2020-10-01
Add HoloViews section under new Ecosystem Integration chapter

jonmmease:
# Overview
This PR adds a new top-level "Ecosystem Integration" chapter, intended to satisfy https://github.com/plotly/dash-docs/issues/965.  Under this chapter, a new "HoloViews" section is added.

This section documents an in-development capability to display interactive HoloViews objects in Dash.  The particular focus is on how HoloViews makes it possible to implement linked selection and datashader workflows without manually defining callbacks.  This is implemented in the open HoloViews PR at https://github.com/holoviz/holoviews/pull/4605.  This documentation PR should be considered WIP until that PR is merged and the next version of HoloViews is released. 

cc @jackparmer @nicolaskruchten @chriddyp 

@jbednar, @philippjfr, @jlstevens: Feel free to look over the HoloViews background info here if you're so inclined. I'm trying to walk the line of being concise and accurate and providing enough links into the HoloViews documentation for interested folks to dig in and learn more. Thanks for your help over on https://github.com/holoviz/holoviews/pull/4605!

# Screenshots / GIFs
Because the plots in the PR will be a bit difficult to review before this HoloViews version is released, here are Screenshots/GIFs that correspond to each section:

## Display Simple HoloViews Elements with Dash
simple

## Styling Figures Produced by HoloViews
styled

## Linked Selections with HoloViews
linked_selections

## Visualizing Large Datasets with HoloViews and Datashader
datashader

## Combining Datashader and Linked Selections
linked_selections_datashader

## Advanced HoloViews
(These are direct ports of existing holoviews examples, each example includes link to corresponding section in the HoloViews documentation)
#### Bounds & selection stream example
bounds_and_selections_stream

### BoundsX stream example
boundsx_stream

### BoundsY stream example
boundsy_stream

### RangeXY stream example
rangexy_stream

### Multiple selection streams example
multiple_selection_streams

### Point Selection1D stream example
selection1d_stream

### DynamicMap Container
dynamicmap_kdims

---
## Post-merge checklist:

- [x] I understand

- [x] This image was saved and referenced locally rather than via an external link


jlstevens:
Very excited to see this!

I'll be taking a closer look later but everything you've said about HoloViews that I've read so far looks accurate.

dash_docs/chapters/holoviews/examples/bounds-stream.py
```diff
@@ -0,0 +1,55 @@
...
+import holoviews as hv
+from holoviews import streams
+from holoviews.plotting.plotly.dash import to_dash
+
+# Declare distribution of Points
+points = hv.Points(
+    np.random.multivariate_normal((0, 0), [[1, 0.1], [0.1, 1]], (1000,))
+)
+
+# Declare points selection selection
```
jbednar, line 15:
```suggestion
# Declare points selection
```
?

dash_docs/chapters/holoviews/examples/datashader-link-selections.py
```diff
@@ -0,0 +1,58 @@
...
+).opts(title="Datashader with %d points" % len(dataset))
+
+hist = selection_linker(
+    hv.operation.histogram(dataset, dimension="petal_width", normed=False)
+)
+
+# Use plot hook to set the default drag mode to box selection
+def set_dragmode(plot, element):
+    fig = plot.state
+    fig['layout']['dragmode'] = "select"
```
jbednar, line 38: This looks like a good example of what we're trying to make easier in https://github.com/holoviz/holoviews/pull/4463 .

dash_docs/chapters/holoviews/examples/dynamic-map.py
```diff
@@ -0,0 +1,33 @@
...
+app = dash.Dash(__name__)
+
+# Dash display
+components = to_dash(app, [dmap])
+
+app.layout = html.Div(
+    [components.graphs[0],
+     components.kdims["phase"],  # Phase slider
+     components.kdims["frequency"],  # Frequency Slider
+     components.store]
```
jbednar, line 29: Is there no way to tell Dash to show all the sliders, whatever they are, so that this part of the code isn't dependent on the details of the dmap?

jbednar: In all of these files I just keep wishing that `to_dash` would return something that `html.Div` could consume directly, so that users could simply do:

```
app = dash.Dash(__name__)
app.layout = html.Div(to_dash(app, dmap))
```

There's probably something about how it works that stops this, but it just seems like there's a simpler API in there somewhere, at least for typical cases.

dash_docs/chapters/holoviews/index.py
```diff
@@ -0,0 +1,443 @@
...
+        both the figure viewports and other interactive states like the active
+        selection produced by the `link_selection` examples below. If 
+        `reset_button=False`, the default, then this list will be empty.
+    - `store`: A Dash [`Store`](/dash-core-components/store) component that is used
+        internally to maintain the joint interactive state of all of the converted 
+        Dash components.  
+    - `kdims`: For Dimensioned HoloViews containers, the `kdims`
+        property holds a dictionary from key-dimension names to Dash components
+        that represent sliders for each key-dimension. Dimensioned Containers are
+        not discussed further here, see the
```
jbednar, line 123:
```suggestion
        that represent sliders for each key dimension. Dimensioned Containers are
        not discussed further here; see the
```

dash_docs/chapters/holoviews/index.py
```diff
@@ -0,0 +1,443 @@
...
+    
+    See the
+    [Applying Customizations](http://holoviews.org/user_guide/Applying_Customizations.html)
+    section of the HoloViews documentation for more information on styling figures
+    usng the options system.
+    
+    ### Plot Hooks
+    HoloViews aims to expose the most common plot options through the `opts` syntax
+    above, but the coverage of possible plotly configuration options is not exhaustive.
+    HoloViews provides a system call plot hooks to make it possible to apply arbitrary
```
jbednar, line 199:
```suggestion
    HoloViews provides a system called "plot hooks" to make it possible to apply arbitrary 
```

dash_docs/chapters/holoviews/index.py
```diff
@@ -0,0 +1,443 @@
...
+    be used to create a Dash button component. When this button is clicked, the plot
+    viewport and selection states are reset to their original values. 
+     
+    For more background on linked selections in Holoviews, see the
+    [Linked Brushing](http://holoviews.org/user_guide/Linked_Brushing.html)
+    documentation section.
+    
+    Try using the box-selection tool to select regions of space in each figure and
+    notice how the selection of the corresponding data is displayed in both figures.
+    Note that only box selections is supported right now. Lasso selection support
```
jbednar, line 242:
```suggestion
    Note that only box selection is supported right now. Lasso selection support
```

dash_docs/chapters/holoviews/index.py
```diff
@@ -0,0 +1,443 @@
...
+        examples['link-selections.py'][0],
+        style=styles.code_container
+    ),
+    html.Div(examples['link-selections.py'][1], className="example-container"),
+    rc.Markdown('''
+    ## Visualizing Large Datasets with HoloViews and Datashader
+    Another HoloViews feature that is particularly convenient for Dash users is the
+    integration with [Datashader](https://datashader.org/).
+    
+    Datashader is a Python library for quickly creating a variety principled
```
jbednar, line 255:
```suggestion
    Datashader is a Python library for quickly creating a variety of principled
```

dash_docs/chapters/holoviews/index.py
```diff
@@ -0,0 +1,443 @@
...
+    Datashader can handles tens to hundreds of millions. The difference is that rather
+    than passing the entire dataset from the Python server to the browser for rendering,
+    Datashader rasterizes the dataset to a heatmap or image, and only this heatmap or
+    image is transferred to the browser for rendering.
+    
+    To effectively use Datashader in an interactive context, it's necessary to rerender
+    the dataset each time the figure viewport changes. This can be accomplished in 
+    Dash by installing a callback function that listens for changes to the
+    `relayoutData` prop.  With HoloViews, this can be accomplished without manually
+    defining any callbacks.
```
jbednar, line 268:
```suggestion
    `relayoutData` prop.  Because of how HoloViews packages data lazily (without rendering it immediately), replaying this pipeline of transformations can be accomplished without manually
    defining any callbacks, making Datashader much easier to use than if invoked without HoloViews.
```

jbednar (commented):
Very nice!

dash_docs/chapters/holoviews/examples/dynamic-map.py
jbednar: Or maybe I wish that `html.Div` would do something useful with the output of `to_dash`, automatically showing _some_ layout of what's returned (as HoloViews does with Bokeh), and only requiring users to mess with the individual bits if they want to rearrange them or suppress some of them.

nicolaskruchten: This is a good point! We definitely do want to have access to the components like we do now, but it would also be nice to have some sort of auto-layout. 

Maybe it makes sense to have two functions instead of just `to_dash()`? Something like `to_dash_components()` would do the current thing, and `to_dash_layout()` would do something more automagical, returning some assembly of `html.Div`s and the components? So you could do just `app.layout = to_dash_layout(...)` if you like or `app.layout = html.Div(...something..., to_dash_layout(), ...something...)` or if you want more fine-grained control, you extract the individual components and lay them out yourself, as with the current approach.

jbednar: I'm clearly spoiled by working with HoloViews and Panel :-), where there's no tension of that sort -- a layout or displayable plot is already always also a collection of indexable components; the two are quite deliberately one and the same in the HoloViz world so that one never ends up having to make an irrevocable choice like that.  

Here we are bridging between HoloViz and Dash, and it seems like there are two ways of such bridging: providing two pathways as you suggest (_either_ components _or_ already laid out), or else making Dash able to handle the combination of the two (a laid-out collection of components that can be pulled apart and put back together in any combination if desired, but which is already also ready to display as-is otherwise). 

Providing the two paths has the advantage of requiring no changes to Dash, which could be  a big practical consideration, because it means that people don't have to worry about the Dash version, can use an already installed Dash server, etc.  Providing the one path is more in keeping with the HoloViz design philosophy; apart from not having to learn two different functions it just makes Dash a bit more attractive and more natural as a server for HoloViews objects, but it does require updating Dash, and these are maybe not important considerations for Dash users. 

You folks are the ones best equipped to make a call like that, but I do vote for doing one of these two approaches!

nicolaskruchten: The `children` argument of `html.Div()` expects a list of Dash components, and that contract would be very hard to change at this point. 

> a layout or displayable plot is already always also a collection of indexable components

Dash components already mostly behave this way: one can access the children of a component, so if `to_dash()` returned a single instance of `html.Div` one could crawl through it and grab all the graphs and such, but that's not really very ergonomic. I'd probably favour having two functions, but if that's not desirable, we could have `to_dash()` return an instance of a subclass of `html.Div()` (or a monkey-patched instance) that provides the `.graphs[]` and `.kdims[]`-style accessors to the laid-out components?

jonmmease: > we could have to_dash() return an instance of a subclass of html.Div() (or a monkey-patched instance) that provides the .graphs[] and .kdims[]-style accessors to the laid-out components?

This is an interesting idea!

Another alternative would be to add a `children` property to the result of `to_dash` that is a just the list of all of the components. This way you could do

```python
app.layout = html.Div(style={...}, children=to_dash(...).children)
```

An advantage here is that it would be pretty natural to customize other `Div` attributes like `style`/`className` declaratively:

```
`app.layout = html.Div(style={...}, children=to_dash(...).children)`
```

nicolaskruchten: Ah yes, just `.children` would work too if it was a list of components, but then the styling might leave something to be desired, it'd be a bit raw. Maybe something like `html.Div(children=to_dash(...).laid_out)` which would return something a bit more complex than a straight list, like a tree of divs in a grid layout or something? HoloViews has a concept of layout, right? Maybe some of that could be returned here also? like "all in a line" or "a 2x2 grid"?

jonmmease: > HoloViews has a concept of layout, right?

On its own, all of the layout that HoloViews does is reflected in the subplot structure of individual figures. Once you're talking about displaying multiple separate HoloViews objects, you're in the realm of Panel (and now Dash). 

In principle, we could take this up a level and have `to_dash` work on Panel layouts / components. There is a lot of overlap in what both Dash and Panel call components (https://panel.holoviz.org/user_guide/Components.html), but it would take a bit of care to figure out what could be faithfully converted.

> Ah yes, just .children would work too if it was a list of components, but then the styling might leave something to be desired, it'd be a bit raw. Maybe something like html.Div(children=to_dash(...).laid_out) which would return something a bit more complex than a straight list, like a tree of divs in a grid layout or something?

Yeah, maybe just `.div`/`.laid_out` prop that holds some opaque `Div` component with a substructure that the user shouldn't worry about.  But this is one of those long-tail things were I don't see a clear place to draw the line of how much configuration to support with new args in the `to_dash` method. e.g. it would be nice to have the option to return figures in DBC or DDK cards if you're using those libraries, but I'm not sure I really want to have that kind of thing flowing through `to_dash`.

💡  In the past, we've talked about having layout helpers that would, for example, line up widgets in a pane on the left and stack / tile figures on the right.  Maybe this kind of layout helper inputs a plain list of components and looks at their types to decide where to put them.  This would decouple layout from `to_dash`, and make it useful in other contexts.

nicolaskruchten: Hehe yes your 💡 is basically the `dash_express` prototype I'm cooking up.

OK, in this case I vote for the simple `to_dash(...).children` which is almost as terse as @jbednar's initial suggest above :)

jbednar: Sounds like you guys have it in hand! Thanks for thinking about ways to make this easier, as I'm really excited about having Dash be a convenient way to serve HoloViews objects!

jonmmease: Ok, I added the `children` property over in https://github.com/holoviz/holoviews/pull/4605 and updated the examples here to use it in 2aa6d81.  Thanks for the discussion!

nicolaskruchten: Awesome! So I haven't been able to run this stuff but the sliders: do they have textual labels in this mode, or are they just sliders?

nicolaskruchten: Ah the "sliders" are divs that wrap sliders and have labels, ok nice.

jackparmer:
Should there also be a simple example like this one, but for a map?

jonmmease:
> Should there also be a simple example like this one, but for a map?

That would be great. Unfortunately, the plotly HoloViews backend doesn't support the HoloViews `Tiles` map element yet (https://holoviews.org/reference/elements/bokeh/Tiles.html).  It definitely could in the future, it would just need to be wired up to the plotly mapbox machinery.

jackparmer:
> It definitely could in the future, it would just need to be wired up to the plotly mapbox machinery.

Awesome. As discussed in Slack, let's spend another week or 2 to get this in. The Plotly + Datashader + Holoviews use cases are so much richer with maps! 🗺️

jonmmease:
Now that plotly map support has landed in HoloViews master. I added three new examples to the docs

## Map Overlay
map_overlay

## Visualizing a Large Geographic Datasets with Datashader
mapbox_datashader

## Mapbox datashader and linked selections
datashader_linked_mapbox

cc @jackparmer

dash_docs/chapters/holoviews/index.py
```diff
@@ -255,24 +255,26 @@
     ),
     html.Div(examples['link-selections.py'][1], className="example-container"),
     rc.Markdown('''
-    ## Visualizing Large Datasets with HoloViews and Datashader
+    ## Visualizing a Large Datasets with Datashader
```
jbednar, line 258:
```suggestion
    ## Visualizing Large Datasets with Datashader
```

dash_docs/chapters/holoviews/index.py
```diff
@@ -316,8 +318,72 @@
     ),
     html.Div(examples['datashader-link-selections.py'][1], className="example-container"),
     rc.Markdown('''
+    ## Map Overlay
+    Most 2-dimensional HoloViews elements can be displayed on top of a map by
+    overlaying them on top of a `Tiles` element.  There are three approaches to 
+    configuring the map that is displayed by a `Tiles` element:
+    
+      1. Construct a `holoviews.Tiles` element with a WMTS tile server url.
```
jbednar, line 326:
```suggestion
      1. Construct a `holoviews.Tiles` element with a templated WMTS tile server url.
```

dash_docs/chapters/holoviews/index.py
```diff
@@ -316,8 +318,72 @@
...
+         E.g. `Tiles("https://maps.wikimedia.org/osm-intl/{Z}/{X}/{Y}@2x.png")`
+      2. Construct a `Tiles` element with a predefined tile server url using a
+         function from the `holoviews.element.tiles.tile_sources` module. E.g. `CartoDark()`
+      3. Construct a `holoviews.Tiles` element with an empty string, then use `.opts`
+         to supply a mapbox token and style.
+         E.g. `Tiles("").opts(mapboxstyle="light", accesstoken="pk...")`
```
jbednar, line 332: The empty string seems a bit awkward here; it certainly confused me the first time I saw it, as I hadn't noticed that aspect of it in the hv PR. I don't know if there is still time to change it in hv (and if so, don't worry about it), but it seems like something like `Tiles("mapbox:light")` would be more obvious and more readable to users.

dash_docs/chapters/holoviews/index.py
```diff
@@ -316,8 +318,72 @@
...
+    of the predefined `StamenTerrain` WMTS map.
+    '''),
+    rc.Markdown(
+        examples['mapbox-points.py'][0],
+        style=styles.code_container
+    ),
+    html.Div(examples['mapbox-points.py'][1],
+             className="example-container"),
+    rc.Markdown('''
+    ## Visualizing a Large Geographic Datasets with Datashader
```
jbednar, line 352:
```suggestion
    ## Visualizing a Large Geographic Dataset with Datashader
```

I'm getting deja vu here! It's hard to see what this entire document is going to be like when rendered, when only reading through the git diffs, but this section seems like it's copying the title from the beginning of this document (line 258). Presumably one of them should be renamed?

jbednar (commented):
Nice!

dash_docs/chapters/holoviews/index.py
jonmmease: Yeah, I'm open to either. I guess my rationale was the backend specific features generally go in `.opts`.

@jlstevens @philippjfr do you have any opinion on this?

jbednar: Yes, that's tricky, but semantic information (such as selecting a dataset) normally goes in the element itself, and in this case the selection of tile source seems like data, not options (it's _what_ is being plotted, not _how_ it is being plotted).  One could imagine other backends also supporting whatever mapbox options Plotly supports, even if they don't initially and even if they never to...

jlstevens: I agree this is tricky and I don't like the empty string much either. That said, I suspect this use of `.opts` is the correct thing to do here as I don't really think `Tiles` with an empty constructor really makes much sense.

jlstevens: Discussing with @jonmmease and @philippjfr, we decided `.opts` is correct for now but `Tiles` should support `None` (i.e `Tiles(None)`).

jonmmease: And the tiles data argument will default to `None`, so `Tiles().opts(...)` will work. PR https://github.com/holoviz/holoviews/pull/4713.

jonmmease:
All tests (except Persy) are passing and this is ready for final review @chriddyp @alexcjohnson

jingningzhang1:
@chriddyp or @alexcjohnson would you be able to approve? When the docs are live, we will be able to link it in Jon's article :)

chriddyp (approved):
💃

jonmmease:
Thanks @chriddyp !  Will this get deployed automatically?

chriddyp:
@jonmmease - Yup 🙂

PR #985, opened 2020-10-06
Add tools.jl [WIP]

rpkyle:
This PR proposes to contribute the Julia version of the functions contained within `tools.py` to the Dash documentation.


alexcjohnson:
@rpkyle can we close this?

rpkyle:
@alexcjohnson I believe @jdamiba just grabbed the functions and committed them to his own feature branch, so that seems reasonable. If we need to add more functions, we'll do it in a new PR.

PR #988, opened 2020-10-08
Add `enable_dev_tools`

chriddyp:


dash_docs/chapters/devtools/index.py
```diff
@@ -276,6 +276,33 @@ def update_graph(value):
...
+```python
+app.enable_dev_tools(
+    dev_tools_ui=True,
+    dev_tools_serve_dev_bundles=True,
+)
+```
+
+We do not recommend using these settings in your deployed applications in general usage because:
+- Displaying serverside tracebacks is a security vulnerability
+- Dev Tools features incur a performance penalty: component validation and loading source maps has a cost.
```
Marc-Andre-Rivet, line 302: Source maps are only loaded when the browser's devtools pane is opened and such don't contribute to the slowdown in most scenarios. The performance impact comes from using unminified code which is larger, the additional JS content like in the renderer's case with the callback graph UI, and using the dev version of React/ReactDOM (potentially the same applies to other libraries but these would be the most significant ones in most cases)

Marc-Andre-Rivet:
@chriddyp Tests seem to fail for iffy reasons
```
AttributeError: module 'enum' has no attribute 'IntFlag'
```

Wonder if we have another one of these deps that are not version locked and that updated in a breaking way..

Marc-Andre-Rivet (approved):

PR #990, opened 2020-10-12
initial attempts at using modules

rpkyle:
This PR merges in the changes to support using chapters via modules in the Dash.jl docs. `DashUserGuideComponents` are included, but not yet used in the docs themselves.


jdamiba:
This is a really big step Ryan! I'm learning a lot about metaprogramming and Julia modules by going over this code.

These thoughts aren't fully fleshed out, but I think I am beginning to have an idea as to why we are only rendering the layout of the example Dash apps that have callbacks when we execute `run.jl`. 

Let's take the `Basic Callbacks` chapter as an example. 

The `layout` of the Dash app for this chapter is defined in `/dash-docs/dash_docs/chapters/basic_callbacks/index.jl`.  

This `layout` contains three types of components:

1. markdown components for the explanatory text
2. components to display the source code of the example Dash apps
3. components to display the `layout` of the example Dash apps

Crucially, the callbacks associated with the example Dash apps are registered on the `app` of this chapter- not it's `app.layout`! For example:

```
basic_input.callback!(app)
basic_state.callback!(app)
callback_chain.callback!(app)
hello_slider.callback!(app)
multi_inputs.callback!(app)
multi_outputs.callback!(app)
simple_callback.callback!(app)
```

Therefore, when we set the `children` of the `chapter` div in the layout of the Dash app defined in `dash-docs/run.jl`, we do not provide any information about the callbacks that are registered on the `app` which has this `layout`. 

So it makes sense that we only see the layouts of the example apps and that the callbacks are not executed- the chapter Dash apps don't have a `run_server()` statement, they're just modules that define the layout of a Dash app and register callbacks on it.

Somehow, in `run.jl` we need to *execute* the callbacks that are registered on the app whose `layout` we are displaying. 

The problem is, I am not sure how to introspect a Dash `app` object to determine which callbacks are registered on it. 

I think I need to look into what actually happens when we call the `callback!` method exported by Dash.jl.

jdamiba:
@rpkyle I think I've figured out how to register the example callbacks on the outermost container Dash app. 

You can run `run.jl` in this branch to see it in action:

https://github.com/plotly/dash-docs/tree/julia-tutorial-rpk-interactive-callbacks

PR #991, opened 2020-10-13
Keep sidebar ad in the lower left corner as the sidebar scrolls over it

alexcjohnson:
That way there's only one thing to scroll on the page, so it's easier to find things in the sidebar. As opposed to right now, you have to scroll to the top in order to _see_ the sidebar menu, then if it's long because you opened some sections you have to scroll around in it.

Also this keeps the ad permanently in the corner, though that wasn't the primary purpose from my standpoint.

- [x] tested responsive behavior is still 👌 

@chriddyp what do you think? cc @jingningzhang1


jingningzhang1:
This is fantastic. Thank you @alexcjohnson!

chriddyp:
💃

PR #994, opened 2020-10-16
Add Dash R Pattern Matching Callbacks Chapter

HammadTheOne:
This PR adds the pattern matching callback chapter to the Dash for R docs following the template of the Python documentation. It includes 4 example apps and fixes for the Dash R docs in light of updates to the `DashR` package that allow the Docs app to be compatible with those changes. It also includes fixes for the `dash_user_guide_components` R package.

To Do: 
- [x] Investigate and resolve `error: All unnamed arguments must be length 1 from NULL` error with `dev` version of Dash R. 
https://github.com/plotly/dashR/pull/233

The error has been traced to [this commit](https://github.com/plotly/dashR/commit/5bd1bbb9be2ee2a49ad90ae676c98c9d36d42120) and involves the `glue::glue` function call. I'll update this PR once that's been fixed so any reviewer can also confirm that PMC's are working on their end. 

Post-merge checklist:

- [ ] I understand

- [x] Comment on the original Dash issue with a link to the new docs.


run.R
```diff
@@ -475,6 +478,12 @@ app$callback(
                   want to avoid the extra time that it takes to make a roundtrip to the server. Clientside
                   callbacks allow you to write your callbacks in JavaScript that runs in the browser."
                 ),
+                components$Chapter(
+                  'Pattern-Matching Callbacks',
+                  href='/pattern-matching-callbacks',
+                  caption="The pattern-matching callback selectors `Match`, `ALL`, & `ALLSMALLER` allow you to
```
rpkyle, line 484:
```suggestion
                  caption="The pattern-matching callback selectors `MATCH`, `ALL`, & `ALLSMALLER` allow you to
```

dash_docs/chapters/external_resources/examples/addingExternalCSS.R
```diff
@@ -2,14 +2,14 @@ library(dash)
 library(dashHtmlComponents)
 
 external_scripts <- list(
-  list('https://www.google-analytics.com/analytics.js'),
+  list(src = 'https://www.google-analytics.com/analytics.js'),
```
alexcjohnson, line 5: Is it necessary in dash for R to wrap these in `list(src = ...)`? On the Python side, if all you have is a `src` (or an `href` for `external_stylesheets`), you can just pass in a string for each script. The only reason you need the list is if you have other attributes (`integrity`, `crossorigin` etc).

HammadTheOne: In Dash for R it is necessary for them to be named elements. The relevant bit of code I found was:

https://github.com/plotly/dashR/blob/add8dd3d93b23c0feb966e39f711ed58be256373/R/utils.R#L679

alexcjohnson: A little farther down though it looks like a plain string is accepted
https://github.com/plotly/dashR/blob/add8dd3d93b23c0feb966e39f711ed58be256373/R/utils.R#L690
and in constructing the html it seems like there's always an `if(is.list(...))`
https://github.com/plotly/dashR/blob/add8dd3d93b23c0feb966e39f711ed58be256373/R/utils.R#L621-L623
Not a big deal, mostly I just want to know whether this behavior matches Python, but if we do have the simpler version it would be nice to use it when it fits.

alexcjohnson: To be clear: I'm not saying the original form is correct, I'm suggesting it looks like a bare string would work
```suggestion
  'https://www.google-analytics.com/analytics.js',
```

alexcjohnson: And as a separate note: did you make this change because the unnamed list threw an error, or because the desired script just didn't show up (or didn't load correctly) on the page? If the latter, perhaps `assertValidExternals` should be improved to catch this case.

HammadTheOne: In this case, I made the change because of the `"A valid URL must be included with every entry in external_scripts. Please sure no 'src' entries are missing or malformed."` error. And I ran into a similar one for the `href` elements. I'll take another look and see what happens in `assertValidExternals`, because it does seem like it _should_ work.

HammadTheOne: Bare strings do work :+1: 

Added/fixed in 3454976.

rpkyle (approved):
Nice work, just made a couple minor edits, but overall looks :sparkles: 

:woman_dancing: LGTM!

PR #1003, opened 2020-10-27
add dash-renderer version requirement to pattern matching callbacks docs

Keou0007:
Post-merge checklist:

- [X] I understand

### DIFF
--- a/dash_docs/chapters/pattern_matching_callbacks/index.py
+++ b/dash_docs/chapters/pattern_matching_callbacks/index.py
@@ -14,6 +14,7 @@ layout = html.Div([
     rc.Markdown(
     '''
     New in Dash 1.11.0!
+    (requires dash-renderer 1.4.0 or greater)
 
     The pattern-matching callback selectors `MATCH`, `ALL`, & `ALLSMALLER`
     allow you to write callbacks that respond to or update an


jwhendy:
I checked that article yesterday as well after the version finding and saw a requirement was listed for `dash`. I'm not sure how that's managed here, but I would suspect that a `dash` release implies versions for the associated libraries as well. It's cleaner to just specify a `dash` version (and more past/future proof, as one would need to update all instances of this to specify other associated package versions), but I would say:

- can you reproduce from a completely fresh environment how to get `dash=1.16.3` but `dash-renderer < 1.8.2`? That was the identified cause. To me, `pip freeze` showing `dash` in a local file still looks suspicious (like it was updated in an atypical manner), so it's not clear what someone would experience if they did `pip install dash` or `pip install --upgrade dash`.
- it would be nice if it was intuitive which versions of each lib are associated with each `dash` [release](https://github.com/plotly/dash/releases). I also checked there and was not sure how @alexcjohnson knew that `dash-renderer` was the key library that needed to be updated. Maybe listing each `dash-*` version associated with each release could be of benefit to users?

Either way, great initiative and contribution! Just not sure what the best solution will be to this challenge. From my original issue, it seems that the vague message is still a contributing factor. One doesn't know if it's their code or something system related, which is certainly frustrating.

Keou0007:
> can you reproduce from a completely fresh environment how to get dash=1.16.3 but dash-renderer < 1.8.2?

`conda install dash` will get it done.

```
  dash               pkgs/main/noarch::dash-1.16.3-py_0
  dash-core-compone~ pkgs/main/noarch::dash-core-components-1.3.1-py_0
  dash-html-compone~ pkgs/main/noarch::dash-html-components-1.0.1-py_0
  dash-renderer      pkgs/main/noarch::dash-renderer-1.1.2-py_0
  dash-table         pkgs/main/noarch::dash-table-4.4.1-py_0
```

As I said in the other thread, these are the newest versions of these packages in the anaconda main repository.
I don't understand how things transition from conda-forge into the main repo, I suspect someone in the anaconda organisation makes decisions on which packages to include in anaconda releases.

My solution in this PR is not really a solution, but it is a something rather than a nothing that would have at least pointed me in the direction of a solution to the problem I was facing. Better error messages also seems necessary, keeping in mind that any changes will obviously only propagate to newer versions, so unless someone wants to get in touch with the anaconda people and ask them to expedite moving some newer version into the main repo, then something like this is probably as good as is possible for now.

I agree that dash releases should come with attached dependency releases. This likely needs to be fixed in the dependency requirements, which would be a different issue and a different PR and would need to be actioned in the conda-forge feedstock. Unfortunately learning how conda packaging works is not something I've gotten around to yet, so that PR is beyond me.

alexcjohnson (approved):
Thanks @Keou0007 - I think your note here is a good idea for now. I'll comment further in https://github.com/plotly/dash/issues/125
💃

PR #1004, opened 2020-10-27
Tool to build dash docs chapter from notebook + WIP chapter on image annotation

emmanuelle:
This PR is made of two contributions:
- I modified and extended `tools.py` so that it can parse a jupyter notebook file (to be included as a markdown file for easier version control, but it could also be an `ipynb` file), where every cell is a `JupyterDash` app. Indeed, a notebook is an easy way to develop small apps for documentation (which can also be hosted on binder for example). 
- This tool was motivated by the notebook which I have written for new documentation on image annotations with shape drawing. This tutorial is still work in progress, I'll polish the content and add a few examples today.

I'm interested in an early high-level review of the "notebook to dash docs chapter" process. I noticed a few caveats, for example:
- one must avoid name collisions between different cells of the notebook, because even if every cell of the notebook is a different `JupyterDash` app, all their layouts are merged in a single Dash app. A future improvement could check for collisions and change component names.
- at the moment it is not possible to have cells not defining a Dash app. Since every cell must be transformed into layout elements (and possibly callbacks to be attached), one could also permit to have notebook cells displaying plotly figures, and include the figure in a `dcc.Graph` as layout element.  

Post-merge checklist:

- [x] I understand

- [x] I considered whether I could replace the `dcc.Markdown` call with `rc.Markdown`, which will replace relative links with `tools.relpath` internally. Otherwise, I used e.g. `<dccLink href=tools.relpath('/layout') children="the first chapter"/>` instead of `[the first chapter](/layout)` (importing `tools` from `dash_docs`), and set `dangerously_allow_html=true` in the `dcc.Markdown` call.

### DIFF
--- a/dash_docs/chapter_index.py
+++ b/dash_docs/chapter_index.py
@@ -245,6 +245,7 @@ URLS = [
             '/dash-bio/'
             '/dash-daq/',
             '/canvas/',
+            '/annotations/',
             '/cytoscape/'
         ],
         'chapters': [
@@ -545,6 +546,21 @@ URLS = [
                     }
                 ]
             },
+            {
+                'name': 'Dash Image annotations',
+                'chapters': [
+                    {
+                        'url': '/annotations',
+                        'name': 'Overview & Reference',
+                        'content': chapters.dash_annotations.index.layout,
+                        'description': (
+                            'Image rendering, drawing, annotations '
+                            'for image processing applications.'
+                        )
+                    }
+                ]
+            },
+
 
             {
                 'name': 'Dash Cytoscape',
--- a/dash_docs/chapters/__init__.py
+++ b/dash_docs/chapters/__init__.py
@@ -1,6 +1,7 @@
 import os
 from .import basic_callbacks
 from .import auth
+from .import dash_annotations
 from .import dash_core_components
 from .import dash_cytoscape
 from .import d3_react_components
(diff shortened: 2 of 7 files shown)


emmanuelle:
The CI failure seems unrelated to this PR, but the `dash-docs` build does not seem to complete ever. 

I've updated the tutorial on annotations, so that the PR is now ready for review I think.

alexcjohnson:
Neat idea @emmanuelle! We may want to have a discussion on Friday about how this would fit with some of the ideas @Marc-Andre-Rivet, @rpkyle, and I were discussing earlier today about how we can make multi-language (ie R, Julia) support more manageable, but anything that makes it easier for more folks to contribute to the docs deserves serious consideration 🥇 

The test failures just come from the new page not being in the sitemap I think - I believe you just need to run `python generate_sitemap.py` - then hopefully the percy job will complete as well.

emmanuelle:
@jonmmease is there any plan to build the equivalent of JupyterDash for Julia and/or R? Indeed, the workflow proposed in this PR (writing a notebook with several cells, each of them corresponding to a JupyterDash app) works well for Python, with less boilerplate code to write the `index.py`. However, the workflow does not translate directly to Julia and R.

jonmmease:
> @jonmmease is there any plan to build the equivalent of JupyterDash for Julia and/or R?

For R, I've talked about this a little bit with @rpkyle, but there were some technical challenges with the R webserver library that made it less than straightforward.  I've not discussed it with anyone from the Julia side.

PR #1011, opened 2020-10-31
Updated Dockerfile for Dash R docs

HammadTheOne:
This update bumps the base R image for the Dash R docs deployment to `3.6.3`, and rebuilds the image with an updated GPG key for accessing CRAN packages. The image is hosted in the `heroku-docker-r` repo on Dockerhub with the tag `3.6.3_heroku18`. 

[x] Verified that updated Dockerfile can be used to successfully deploy docs on Heroku test instance.


rpkyle (approved):
Thanks again @HammadTheOne! :woman_dancing:

PR #1012, opened 2020-11-02
add newline to end of example apps so they format correctly

jdamiba:
Currently, example apps that do not have a trailing newline get an extra set of backticks appended to them. This means that when copying/pasting into a dev environment, users get an error. 

This fix ensures that apps get a trailing newline, preventing this error from occurring. 

Before:
Screen Shot 2020-11-02 at 5 17 20 PM

After:
Screen Shot 2020-11-02 at 6 45 46 PM

### DIFF
--- a/dash_docs/tools.py
+++ b/dash_docs/tools.py
@@ -180,7 +180,7 @@ def load_example(path, relative_path=False):
             raise e
 
     return (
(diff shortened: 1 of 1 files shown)


alexcjohnson (approved):
Excellent! 💃

PR #1013, opened 2020-11-03
run black linter on example apps

jdamiba:
follow up to https://github.com/plotly/dash-docs/pull/1012

With the `pwd` set to `dash-docs/dash_docs/chapters`, I ran `black */examples`.


VeraZab:
yapf is an alternative formatter with lots of options (https://github.com/google/yapf) (suggestion from @jonmmease )
also from dash core weekly: should check if the same issue exists in DED repo

Issue #973, opened 2020-09-18
Update `[Input(...), ...]` to `Input(...), ...`


Marc-Andre-Rivet:
Followup to https://github.com/plotly/dash/pull/1180

- Replace all usages in dash-docs to use the new syntax instead
- Document the old syntax in the callbacks section

PR #1020, opened 2020-11-12
Update single input /state callback definition syntax to not use lists 

jdamiba:
closes https://github.com/plotly/dash-docs/issues/973

The purpose of this PR is to update examples presented to the user such that single Input and single State callback definitions do not use lists, which is possible as of Dash 1.15.0.


jdamiba:
@harryturr Can you take a look at this? It only targets public-facing code that is the `dash_docs/chapters/` directory. Also, it is my understanding that if a callback has mulitple Inputs or State, then those still need to wrapped as a list.

harryturr:
> @harryturr Can you take a look at this? It only targets public-facing code that is the `dash_docs/chapters/` directory. Also, it is my understanding that if a callback has mulitple Inputs or State, then those still need to wrapped as a list.

@jdamiba Based on the [ v1.15.0 changelog](https://github.com/plotly/dash/blob/dev/CHANGELOG.md#changed-4) and [tests in the repo ](https://github.com/plotly/dash/blob/5e717f1bd793ecaffc1200d97749b141feff98d2/tests/assets/todo_app.py#L43), it seems multiple `Input` or `State` does not require lists.

jdamiba:
@harryturr I've updated this PR so that all Outputs, Inputs, and State assignments to callbacks no longer use lists.

Marc-Andre-Rivet (approved):

PR #1024, opened 2020-11-18
dash enterprise sidebar link update

tobinngo:
Updated sidebar link page from `/dash` to `/dash-enterprise`

image

image


alexcjohnson (approved):
Nice catch! 💃

PR #1030, opened 2020-11-24
Adding database connections chapter

HammadTheOne:
This PR complements plotly/dash-enterprise-docs#632 by adding a chapter on Database Connections with Dash apps. The page highlights some of the "best practices" used within the SQL template apps, and some more relevant documentation. 

Once this is merged, a git tag will be created, and the version of `dash-docs` in `dash-enterprise-docs` will be updated as well. 

Relevant Pages to Visit:
http://127.0.0.1:8060/dash-enterprise/database-connections

Post-merge checklist:

- [ ] I understand


dash_docs/chapter_index.py
```diff
@@ -362,6 +362,13 @@ def component_list(
                     'description': 'Use a staged Dash App to test changes before updating '
                     'your production Dash App.'
                 },
+                {
+                    'url': '/dash-enterprise/database-connections',
+                    'content': chapters.dash_enterprise.dash_enterprise_chapters.DataConnections,
+                    'name': 'Dash App SQL Database Connections',
```
rpkyle, line 368: Given that most of the other entries in this section appear to include action verbs to start, what if we did something similar here?

```suggestion
                    'name': 'Connect a Dash App to an SQL Database',
```

dash_docs/chapters/dash_enterprise/dash_enterprise_chapters.py
```diff
@@ -3000,6 +3000,123 @@ def snapshot_page(n_clicks, url, wait_selector):
...
+# # # # # # #
+DataConnections = html.Div(children=[
+    html.H1('Dash App SQL Database Connections'),
+
+    rc.Markdown(
+    '''
+
+    ## Managing Connection Drivers and Libraries
+
+    Dash Apps can use open source libraries in the Python ecosystem to query external databases and datastores in callbacks or job queues.
```
rpkyle, line 3014: If the content here is (or will eventually be) useful to R and Julia users, it might be good to omit the "in the Python ecosystem" bit.

```suggestion
    Dash Apps can use open source libraries to query external databases and datastores in callbacks or job queues.
```

dash_docs/chapters/dash_enterprise/dash_enterprise_chapters.py
```diff
@@ -3000,6 +3000,123 @@ def snapshot_page(n_clicks, url, wait_selector):
...
+
+    The Python DB-API standardizes the interface for most Python database access modules. This allows us to write Dash
+    apps with code that can be modified with little effort to connect to different types of databases with different dialects
+    by following the standardized API. For more information on the Python DB-API, see [PEP-0249](https://www.python.org/dev/peps/pep-0249/).
+
+    To query databases, you'll need to install the Python database packages and in many cases system level dependencies.
+    For complete example, see the "Database Connections" section in Sample Apps & Templates:
```
rpkyle, line 3021:
```suggestion
    For complete examples, see the relevant "Database Connections" section in Sample Apps & Templates:
```

dash_docs/chapters/dash_enterprise/dash_enterprise_chapters.py
```diff
@@ -3000,6 +3000,123 @@ def snapshot_page(n_clicks, url, wait_selector):
...
+    1. Install the database driver & system level dependencies. This will vary by database type and we recommend viewing
+    the examples that we've provided first. Many of the database drivers more easily installed on Linux environments,
+    and so we recommend developing these apps in a [Dash Enterprise Workspace](/Docs/workspaces).
+
+    2. Create a connection. We recommend using [SQLAlchemy](https://www.sqlalchemy.org) if available for your database.
+
+    3. Store your connection's database password as an environment variable in your App Settings in the Dash Enterprise
+    App Manager instead of storing it within code.
+
+    4. Poll the database initially to check that it is ready to begin accepting connections. We can create a simple
```
rpkyle, line 3043: Instead of using "poll the database" here, I wonder if we could simplify the language a bit to avoid jargon. Also, I think we're actually testing multiple things in this step:

```suggestion
    4. Test that the database server has started successfully, verify that it is accepting connections, and validate the credentials and connection string; we can create a simple
```

dash_docs/chapters/dash_enterprise/dash_enterprise_chapters.py
```diff
@@ -3000,6 +3000,123 @@ def snapshot_page(n_clicks, url, wait_selector):
...
+    `try_connection` function which sends a basic query to the database and checks to see if it is successful. If it fails,
+    the connection is retried after an exponentially increasing delay. This allows us to easily diagnose any issues related
+    to the database itself and differentiate them from the Dash app callbacks. Eg:
```
rpkyle, line 3046: Is it possible to elaborate on what "differentiate them from the Dash app callbacks" means here? 

If I understood correctly, I think the gist is that we want to be able to to isolate issues arising from errors in the callback logic vs. those caused by database configuration or connectivity problems.

dash_docs/chapters/dash_enterprise/dash_enterprise_chapters.py
```diff
@@ -3000,6 +3000,123 @@ def snapshot_page(n_clicks, url, wait_selector):
...
+    It is possible to use --preload and celery with a node pool but it requires extra caution. For more details, see:
+    > - [SQLAlchemy: Using Connection Pools with Multiprocessing or os.fork()](https://docs.sqlalchemy.org/en/14/core/pooling.html#using-connection-pools-with-multiprocessing-or-os-fork)
+    > - [Not The Same Pre-fork Worker Model](https://www.yangster.ca/post/not-the-same-pre-fork-worker-model/)
+    > - [SQLAlchemy connection pool within multiple threads and processes](https://davidcaron.dev/sqlalchemy-multiple-threads-and-processes/)
+
+    To disable connection pooling, you will use the `NullPool`:
+    ```
+    from sqlalchemy.pool import NullPool
+    engine = create_engine(
+        'postgresql+psycopg2://scott:tiger@localhost/test',
```
rpkyle, line 3074: I've never used SQLAlchemy, but what does `scott:tiger@localhost/test` reference here? Is this related to the default user and database name?

rpkyle: Ah, now I think I understand -- do we want to change this to something like `postgresql+psycopg2://username:password@localhost/database`?

rpkyle (changes requested):
Overall this looks helpful, and a nice addition to the docs! Just a couple questions about trying to clarify the language a bit more, and once resolved I'd be comfortable merging this one in.

dash_docs/chapters/dash_enterprise/dash_enterprise_chapters.py
HammadTheOne: Fixed in b13cda4

dash_docs/chapters/dash_enterprise/dash_enterprise_chapters.py
HammadTheOne: Updated in b13cda4

HammadTheOne:
After checking out these chapters within the DE docs, it seems we need to make some changes to accommodate the following commit: https://github.com/plotly/dash-docs/commit/dd816aa60d38bad75ee2bde35ed0ac9c64f5e11a.

The `dash-enterprise` chapter no longer has the tiled format, and instead has a list format:
image

I'm going to investigate a little more and get a fix in for this because it seems unintentional.

rpkyle (approved):
I can't speak to the list vs. tile format issue, but the rest of the changes here look good to me. :woman_dancing:

PR #1034, opened 2020-12-02
Revert "Revert "Add HoloViews section under new Ecosystem Integration chapter""

chriddyp:
Reverts plotly/dash-docs#1033

We can use branch to fix the issues that caused the original HoloViews PR to break.

To start debugging, I'd recommend checking out the Heroku review app that will get printed out below. My guess is that it won't work and we can check the heroku logs for the review app.


jonmmease:
Thanks @chriddyp, I'll be watching for the review app here

jonmmease:
@chriddyp I replaced the live Datashader examples with GIFs and everything seems to be working now at https://dash-docs-pr-1034.herokuapp.com/holoviews.

Does this look alright to you?

chriddyp (commented):
💃 works for me!

PR #1039, opened 2020-12-08
Add new chapter for Dash Slicer

almarklein:
*DRAFT: this needs some (yet uncommitted) changes to dash-slicer plus a new release*

Post-merge checklist:

- [x] I understand

Other todo's:

* [x] Add basic guide with examples.
* [x] Add reference docs.
* [x] Need to commit corresponding changes to `dash_slicer`.
* [ ] Need a new release of `dash_slicer`.
* [ ] Bump the version in requirements.txt
* [x] If I run the server locally, it works fine, except that if I reload the page or navigate to `/slicer` directly, I get a weird `TypeError: unhashable type: 'dict'`.
* [ ] As it now, dash-slicer is Python 3.6+. CI needs to know this.
 
cc @emmanuelle @nicolaskruchten 

For the reference docs, I opted to have the markdown generated by `dash_slicer` (with a function call). I figured that it would be easier to keep the ref docs in sync with the implementation this way, plus we can add the (small) ref docs to the GH readme. But if others prefer to have this logic in dash-docs, we could change this.

### DIFF
--- a/dash_docs/assets/sitemap.xml
+++ b/dash_docs/assets/sitemap.xml
@@ -21,6 +21,9 @@
     <loc>https://dash.plotly.com/canvas</loc>
 </url>
 <url>
+    <loc>https://dash.plotly.com/slicer</loc>
+</url>
+<url>
     <loc>https://dash.plotly.com/clientside-callbacks</loc>
 </url>
 <url>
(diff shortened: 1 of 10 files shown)


almarklein:
A screenshot of the error I get when I refresh or navigate to the page by typing in the address bar instead of clicking the link in the chapter index:
afbeelding

almarklein:
Superseded by #1043

PR #1042, opened 2020-12-09
Fix that dict-ids in examples caused server failure

almarklein:
In #1039 I encountered an odd bug: if I navigated to the new chapter it worked fine, but if I refreshed the page, or navigate to the page by entering the url, then I got a descriptive error message, see the image below.

I added some code to the point of failure to see what the value of `attrib` was at that point, and found it was a dictionary-id of a component in one of the examples (well, in all examples, really). 

Fortunately, the fix is easy, I can confirm that the docs in #1039 don't have the bug with the change proposed here. 

afbeelding

### DIFF
--- a/dash_docs/convert_to_html.py
+++ b/dash_docs/convert_to_html.py
@@ -60,7 +60,7 @@ def _translate_attrib(attrib):
         False: 'false',
         'False': 'false',
     }
-    if attrib in attrib_map:
+    if isinstance(attrib, (str, int)) and attrib in attrib_map:
         return attrib_map[attrib]
     return attrib
 
--- a/dash_docs/convert_to_html.py
+++ b/dash_docs/convert_to_html.py
@@ -60,7 +60,8 @@ def _translate_attrib(attrib):
         False: 'false',
         'False': 'false',
     }
-    if isinstance(attrib, (str, bool, int)) and attrib in attrib_map:
+    # Note that bool is a subclass of int
+    if isinstance(attrib, (str, int)) and attrib in attrib_map:
         return attrib_map[attrib]
     return attrib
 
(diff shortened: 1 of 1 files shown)


almarklein:
Quick note, in case people wonder: `bool` is a subclass of `int`.

dash_docs/convert_to_html.py
```diff
@@ -60,7 +60,7 @@ def _translate_attrib(attrib):
         False: 'false',
         'False': 'false',
     }
-    if attrib in attrib_map:
+    if isinstance(attrib, (str, int)) and attrib in attrib_map:
```
chriddyp, line 63: I'm not sure I follow - What about when the attrib is `True` or `False` as included above?

chriddyp: Ah I see your comment above - `bool` is a subclass of `int`. TIL!

chriddyp: Could you just add a comment above this line saying that `bool is a subclass of int`?

chriddyp (approved):
💃🏼 once my comment is considered

dash_docs/convert_to_html.py
almarklein: That works, because `bool` is a subclass of `int` in Python.

almarklein: I'll just add `bool` to the list to avoid confusion ...

almarklein: GH comments are not fast enough to function as a chat, haha. I can undo that last change and add a comment, if you prefer.

almarklein: Updated to use a comment instead.

PR #1043, opened 2020-12-09
Add new chapter for Dash Slicer - v2

almarklein:
This supersedes #1038 (which was on a branch of my fork of dash docs, preventing deploying to the playground).

- [x] I understand

Other todo's:
* [x] The new chapter fails to load if you reload the page. Will be fixed by #1042.
* [x] Dash-slicer is Python 3.6+, can this work?
 
cc @emmanuelle @nicolaskruchten


dash_docs/chapters/dash_slicer/index.py
```diff
@@ -0,0 +1,127 @@
...
+    of the other slicers. This linking is based on what we call the
+    scene_id. This is a property that can be provided when you
+    instantiate a `VolumeSlicer`. By default, the scene_id is derived
+    from the volume. That's why the linking in this example works: each
+    slicer is given the same numpy array object. By explicitly setting
+    the scene_id, multiple views on different data can be linked as
+    well.
+
+    In addition to using the sliders, you can click in one of the
+    views to make the other views go to the clicked location. Try it!
```
emmanuelle, line 81:
```suggestion
    views to make the other views go to the clicked location. Try it!
    Thanks to this navigation mode, you can optionally omit sliders in the layout when two or more views are present.
```

emmanuelle (approved):
Awesome, thanks a lot! I just left a couple of comments which should be straightforward to accept or implement, so :dancer: once they have been addressed (even I believe either @chriddyp, @nicolaskruchten or @alexcjohnson should also approve before merging). 

In a next PR it'd be great to add an example with overlay but the chapter is already super helpful as it it so let's ship it ASAP!

dash_docs/chapters/dash_slicer/__init__.py
```diff
@@ -0,0 +1,13 @@
+import sys
+import dash_html_components as html
+
+
+if sys.version_info > (3, 10 ):
```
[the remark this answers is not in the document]
almarklein: Woops, I turned `sys.version_info > (3, 0)` temporarily into sys.version_info > (3, 10)` to test this page, but forgot to revert!

dash_docs/chapters/dash_slicer/index.py
```diff
@@ -0,0 +1,127 @@
...
+
```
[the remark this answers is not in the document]
almarklein: Done!

almarklein:
> In a next PR it'd be great to add an example with overlay

Good point! We can then also add the new features to come: the overlaying of scatter traces and contrast limits.

emmanuelle:
Thank for addressing my comments! For me the PR is good to go :woman_dancing: , but we'll need an owner to approve the PR. @alexcjohnson @chriddyp @rpkyle @Marc-Andre-Rivet

dash_docs/chapters/dash_slicer/index.py
```diff
@@ -0,0 +1,135 @@
...
+    ```pip install -U dash-slicer```
+
+    The source is on GitHub at [plotly/dash-slicer](https://github.com/plotly/dash-slicer).
+
+    ## Guide
+
+    ### A first example
+
+    Let's get started with a simple example.
+    (The examples on this page are build with dash-slicer version {dash_slicer.__version__})
```
alexcjohnson, line 39:
```suggestion
    (The examples on this page are built with dash-slicer version {dash_slicer.__version__})
```

dash_docs/chapters/dash_slicer/index.py
```diff
@@ -0,0 +1,135 @@
...
+        examples['slicer_example1.py'][0],
+        style=styles.code_container
+    ),
+    html.Div(examples['slicer_example1.py'][1], className='example-container'),
+
+    rc.Markdown('''
+    In the code above, we first load a 3D numpy array (a volumetric image). Then
+    we instantiate a `VolumeSlicer` object with that data. This object is not
+    a Dash component, but has attributes that are. Its `graph` and `slider`
+    are placed in the layout, as well as a handful of `Store` objects that the
```
alexcjohnson, line 52:
```suggestion
    a Dash component, but has attributes that are. We place its `graph` and
    `slider` in the layout, as well as a handful of `Store` components that the
```
Just to make it clear it's up to you as the user to put them into the layout - otherwise this could be interpreted to mean that by instantiating the `VolumeSlicer` these components are automatically added to the layout.

dash_docs/chapters/dash_slicer/index.py
```diff
@@ -0,0 +1,135 @@
...
+    slicer needs to function.
+
+    If the server is run in debug mode, consider setting `dev_tools_props_check`
+    to False, because it has a big impact on the performance.
```
alexcjohnson, line 56: Thanks for mentioning this in the docs! Has this been discussed somewhere? I'm curious whether this is something we can improve, either in the renderer itself or in the slicer.

almarklein: As I understand it, this is an issue that always occurs, but is especially visible/annoying in the slicer, because it relies on interaction quite a bit. I found out when I saw that debug mode had a big impact on performance, and narrowed it down to this specific feature. The only "discussion" I know of is [this dash-slicer pr](https://github.com/plotly/dash-slicer/pull/26).

At the time I looked whether we can do something in the slicer to avoid the performance penalty, but I don't think we can. There seems to be a penalty for every callback. E.g. changing `data -> figure` to `data -> traces -> figure` caused a huge penalty. So my guess is that this needs to be solved in dash itself.

alexcjohnson (approved):
Love it! This is super fun to play with 🍰 💃

alexcjohnson:
The review app failed, though it was working prior to my suggestions. Investigating...

alexcjohnson:
After digging into this a bit, the issue was startup time but it doesn't seem as though any of the new or even recent code increased that time very much... may just be a slow day at Heroku for some reason. I will continue looking for ways to improve the startup, but for now I think we're OK to merge and deploy, just being more careful to ensure the production page restarts successfully with the new code.

PR #1047, opened 2020-12-10
Update generate_sitemap() to maintain original newlines

almarklein:
I'm on Windows, and I have git configured to *not* convert newlines on push/pull. If I run `generate_sitemap.py`, it will write with `\r\n` because of Pythons automatic newline handling, causing a huge diff.

This PR changes the logic in `generate_sitemap.py` to simply write the sitemap with the same newlines as the current file.

### DIFF
--- a/generate_sitemap.py
+++ b/generate_sitemap.py
@@ -19,8 +19,13 @@ def create_sitemap():
 
 def generate_sitemap():
     sitemap = create_sitemap()
-    with open('dash_docs/assets/sitemap.xml', 'w') as f:
-        f.write(sitemap)
+    # Use the same newlines as is currently used
+    with open('dash_docs/assets/sitemap.xml', 'rb') as f:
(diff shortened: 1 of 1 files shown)


almarklein:
Somewhat related question: if I generate a new sitemap, these lines get added, but CI does not seem to like that, any idea?
```
<url>
    <loc>https://dash.plotly.com/dash-bio/nglmoleculeviewer</loc>
</url>
```

alexcjohnson (approved):
Thanks, this is great! You can probably infer that we don't have a lot of Windows users in house ;)

> `<loc>https://dash.plotly.com/dash-bio/nglmoleculeviewer</loc>`

Ah right, that's a new component we haven't documented yet https://github.com/plotly/dash-bio/pull/496

I'll make an issue for it but for now yes, we'll just manually ignore that change.

💃

PR #1058, opened 2020-12-17
Pandas to_dict('rows') -> to_dict('records')

wbrgss:
`to_dict('rows')` causes warning on pandas 1.1.5:

`Using short name for 'orient' is deprecated. Only the options: ('dict', list, 'series', 'split', 'records', 'index') will be used in a future version. Use one of the above to silence this warning.`

See https://github.com/plotly/dash-docs/commit/07ac6fcd8a711a1abfeef6d7bdb1f883f17684cc for a description of how this got mixed up:
> rows “happens” to work because pandas only looks at the first letter (“r”)

Post-merge checklist:

- [x] I understand


alexcjohnson (approved):
heh yeah this one has been following us around for years... thanks for knocking these down. 💃

PR #1073, opened 2021-01-18
Adding snippet and a GIF demo about dash-embedded in /integrating-dash

xhlulu:
gif: 
embd_state_share

It seems that there's some some issues with JS highlighting; does anyone have an idea?

@harryturr thoughts on the framing and adding snippets?

---

Post-merge checklist:

- [ ] I understand


dash_docs/chapters/integrating_dash/index.py
```diff
@@ -48,6 +48,83 @@
...
+
+    function App() {
+        return (
+            <div style={...}>
+                ...
+                <h1>Embedded Dash Application</h1>
+                <DashApp
+                    config={{
+                        url_base_pathname: "http://dash.tests:8050",
+                    }}
```
harryturr, line 121:
```suggestion
                    }}
                    value={this.state.sharedData}
```

dash_docs/chapters/integrating_dash/index.py
```diff
@@ -48,6 +48,83 @@
     `https://<your-dash-enterprise-hostname>/Docs/embedded-middleware`**,
     replacing `<your-dash-enterprise-hostname>` with the hostname of your
     licensed Dash Enterprise in your VPC. [Look up the hostname for your company's license](https://go.plotly.com/company-lookup)
+    
+    #### Sharing State between a Javascript Host app and an embedded Dash App
+
+    You can find here a simple example illustrating multidirectional shared state, enabling the communication between your JavaScript host app and your Dash app.
+
+    ![GIF showing how to use dash embedded](https://raw.githubusercontent.com/plotly/dash-docs/master/images/dash-embedded-js-host.gif)
```
harryturr, line 56: image
I don't see the gif in the review app

dash_docs/chapters/integrating_dash/index.py
xhlulu: Ok actually it seems it's using the class API, so I'll just change function to class and move it to a `render` function

xhlulu:
I just changed the function into the correct class; let me know if the new version is OK

dash_docs/chapters/integrating_dash/index.py
```diff
@@ -48,6 +48,81 @@
...
+    If you are using a React app, you can import the component and use it inside JSX:
+    ```js
+    import { DashApp } from "dash-embedded-component";
+
+    window.React = React;
+    window.ReactDOM = ReactDOM;
+    window.PropTypes = PropTypes;
+
+    function App() {
+        return (
```
harryturr:
```suggestion
    function App() {
      const [exampleState, setExampleState] = useState(
          {sharedData: {
              ....
          }
      )
        return (
```

dash_docs/chapters/integrating_dash/index.py
harryturr: Sorry, the `this.state.shareData` would be a local state defined in a class object. 

So here maybe you can just add a dummy state variable at the top of our function and pass that as the `value` prop.

harryturr (approved):
@xhlulu that looks good to me, that way we can see the relationship between Consumer components and host app. But if you prefer functional components feel free

xhlulu:
@harryturr Thanks! I think `class` is fine.

@chriddyp anything missing?

dash_docs/chapters/integrating_dash/index.py
```diff
@@ -48,6 +48,100 @@
...
+                        id="host-app-multiply", path=["myObject", "multiplyFunc"]
+                    ),
+                    ConsumerFunction(id="host-app-sum", path=["myObject", "sumFunc"]),
+                    ddk.ControlItem(...),
+                    ...
+                ],
+                width=50,
+            ),
+            ...
+        ]
```
chriddyp, line 140: I'd recommend putting in the callback examples here too

xhlulu: Awesome! I just pushed some truncated callbacks. Let me know if that's what you had in mind

xhlulu:
@chriddyp is this ready to be merged?

chriddyp (approved):
💃🏼

PR #1078, opened 2021-01-20
Remove unused import

KiLJ4EdeN:
nothing fancy just an unused import i came across while implementing examples


alexcjohnson (approved):
Nice catch, thanks! 💃

PR #1079, opened 2021-01-21
inline flask-caching examples with markdown

jdamiba:
addresses https://github.com/plotly/dash-enterprise-docs/issues/716 

The purpose of this PR is to make sure that running `gunicorn --preload index:server` does not result in a `TypeError: sequence item 0: expected str instance, NoneType found` being thrown, which may lead to customer confusion as per the above issue. 

This is achieved by inlining examples that import `flask-caching` with Markdown instead of using the `load_examples()` function, which means that the code is not executed. The `example` folders are kept for future maintainability. 

Test Plan: Run `gunicorn --preload index:server` and note the absence of a `TypeError`.

### DIFF
--- a/dash_docs/chapter_index.py
+++ b/dash_docs/chapter_index.py
@@ -1228,7 +1228,7 @@ URLS = [
                 'url': '/app-lifecycle',
                 'content': chapters.app_lifecycle.index.layout,
                 'name': 'Dash App Lifecycle',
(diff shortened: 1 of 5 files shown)


alexcjohnson:
That'll work, but it was actually really easy to add a no-run option #1080

jdamiba:
superseded by https://github.com/plotly/dash-docs/pull/1080

PR #1080, opened 2021-01-21
load_examples option without running them

alexcjohnson:
🌴 Alternative to #1079 - it was pretty easy to add a `run=False` option to `load_examples`.

@jdamiba I replicated your typo fix, nice catch!

### DIFF
--- a/dash_docs/chapters/sharing_data/index.py
+++ b/dash_docs/chapters/sharing_data/index.py
@@ -5,7 +5,7 @@ import dash_html_components as html
 from dash_docs import tools
 from dash_docs import reusable_components as rc
 
-examples = tools.load_examples(__file__)
+examples = tools.load_examples(__file__, run=False)
 
 
 layout = html.Div([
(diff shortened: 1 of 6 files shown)


dash_docs/chapters/sharing_data/index.py
```diff
@@ -529,8 +529,7 @@ def update_graph_4(value):
         '''),
 
     rc.Syntax(
-        # with Syntax + load_example we are wrapping twice, hence replace()
-        examples['sharing_state_filesystem_sessions.py'][0].replace('```python ', ''),
```
alexcjohnson, line 533: This isn't needed anymore (but it was inconsistent, it should have removed the backticks at the end of the string too)

https://github.com/plotly/dash-docs/blob/353bf3fab78f3139150c934b44eb0e9b908ca90c/dash_docs/reusable_components/Syntax.py#L11-L12

jdamiba:
This looks great! Much more DRY than simply inlining every example we don't want to execute at build time into the chapter Markdown. 

I can confirm that running `gunicorn --preload index:server` in this branch boots up the app as expected, without the `TypeError` mentioned in https://github.com/plotly/dash-enterprise-docs/issues/716.

Looks good to me! 💃🏿

PR #1081, opened 2021-01-22
clarify behavior of prevent_initial_call

jdamiba:
The purpose of this PR is to clarify the behavior of the `prevent_intial_call` attribute of Dash callbacks.


dash_docs/chapters/advanced_callbacks/index.py
```diff
@@ -149,12 +149,63 @@ def slow_function(input):
...
+      @app.callback(Output('layout-div', 'children'), Input('input', 'value'), prevent_initial_call=True)
+      def update_layout_div(value):
+          print('>>> update_layout_div')
+          return value
+
+    ```
+
+    In this case, `prevent_initial_call` will prevent the `update_output()` callback from firing when it's input is first inserted into the app layout as a result of the `display_page()` callback. This is because both the callback's input and output are already contained within the app layout when the callback is set to execute.
+
+    However, `prevent_initial_call` will not prevent the `update_layout_div()` callback from firing, because the callback's output is already present in the app layout but its input is not. Since `suppress_callback_exceptions=True` is specified in this example Dash app, Dash has to asssume that the input is present in the app layout when the app is initialized, which means that it cannot prevent the initial call of the callback that would occur under normal circumstances.
```
alexcjohnson, line 208: It's not so much that it _cannot_ prevent this call, but we _decided it shouldn't_

```suggestion
    However, `prevent_initial_call` will not prevent the `update_layout_div()` callback from firing, because the callback's output is already present in the app layout but its input is not. Since `suppress_callback_exceptions=True` is specified in this example Dash app, Dash has to asssume that the input is present in the app layout when the app is initialized. This means that as far as the output element is concerned, the new input component is a change just like if the existing input had been given a new value, rather than its initial render.
```

dash_docs/chapters/advanced_callbacks/index.py
```diff
@@ -149,12 +149,63 @@ def slow_function(input):
 
 
     rc.Markdown('''
-    ## Prevent Callbacks From Being Executed on Initial Load
+    ## Prevent Callbacks From Being Executed When Their Inputs First Appear The App Layout
```
alexcjohnson, line 152:
```suggestion
    ## Prevent Callbacks From Being Executed When Their Inputs First Appear In The App Layout
```

dash_docs/chapters/advanced_callbacks/index.py
```diff
@@ -149,12 +149,63 @@ def slow_function(input):
...
+      @app.callback(Output('layout-div', 'children'), Input('input', 'value'), prevent_initial_call=True)
+      def update_layout_div(value):
+          print('>>> update_layout_div')
+          return value
+
+    ```
+
+    In this case, `prevent_initial_call` will prevent the `update_output()` callback from firing when it's input is first inserted into the app layout as a result of the `display_page()` callback. This is because both the callback's input and output are already contained within the app layout when the callback is set to execute.
+
+    However, `prevent_initial_call` will not prevent the `update_layout_div()` callback from firing, because the callback's output is already present in the app layout but its input is not. Since `suppress_callback_exceptions=True` is specified in this example Dash app, Dash has to asssume that the input is present in the app layout when the app is initialized. This means that as far as the output element is concerned, the new input component is a change just like if the existing input had been given a new value, rather than its initial render.
```
rpkyle, line 208: The last sentence is a little convoluted here, not sure if what I've suggested makes sense, but would revise in whatever way you feel makes the most sense.

The edits to the initial sentence are proposed to make it a little easier to follow what's happening here, but it works either way.

```suggestion
    However, because the app layout contains only the output of the callback, and not its input, `prevent_initial_call` will not prevent the `update_layout_div()` callback from firing. Since `suppress_callback_exceptions=True` is specified here, Dash has to assume that the input is present in the app layout when the app is initialized. From the perspective of the output element in this example, the new input component is handled as if an existing input had been provided a new value, rather than treating it as initially rendered.
```

rpkyle (approved):
A very useful PR @jdamiba, this is a deceptively simple feature with some real nuances in its application. Thanks for taking this on, just a few minor suggestions in case you want to consider them beyond what @alexcjohnson has already written. 

Mainly recommendations to clarify a tiny bit more. :upside_down_face:

jdamiba:
@ryan Thanks for the review, I've incorporated your edits and I think it is now a much clearer read.

PR #1085, opened 2021-01-26
Emphasize changing server host url when deploying

joelostblom:
This is branched from #1083, so please merge that first (I didn't want to add to it since it was already approved).

@rpkyle Only the two last commits are new here emphasizing the need to specify `host = '0.0.0.0'` since this is different from running locally. I also removed the port argument [since it is just set to the default for app$run_server](https://github.com/plotly/dashR/blob/add8dd3d93b23c0feb966e39f711ed58be256373/R/dash.R#L1183)

Post-merge checklist:

- [x] I understand

### DIFF
--- a/dash_docs/chapters/deployment/index.R
+++ b/dash_docs/chapters/deployment/index.R
@@ -93,7 +93,7 @@ app$callback(output=list(id='display-value', property='children'),
   }
 )
 
-app$run_server(host = '0.0.0.0', port = Sys.getenv('PORT', 8050))
+app$run_server(host = '0.0.0.0')
 ```
 
 ---
--- a/dash_docs/chapters/deployment/index.R
+++ b/dash_docs/chapters/deployment/index.R
@@ -63,7 +63,7 @@ $ git init        # initializes an empty git repo
 
 ---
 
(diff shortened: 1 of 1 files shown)


joelostblom:
Rebased on master and tweaked wording slightly for clarity.

dash_docs/chapters/deployment/index.R
```diff
@@ -93,7 +93,7 @@ app$callback(output=list(id='display-value', property='children'),
   }
 )
 
-app$run_server(host = '0.0.0.0', port = Sys.getenv('PORT', 8050))
+app$run_server(host = '0.0.0.0')
```
rpkyle, line 96: Nice catch! We already populate `port` if unspecified within `run_server`, so this can be removed.

rpkyle (approved):
@joelostblom Thanks again for this one, just one minor comment to address but will approve since it's an easy fix and shouldn't take you more than 15 seconds. Cheers! :woman_dancing:

joelostblom:
Thanks for catching the typo @rpkyle, updated!

joelostblom:
Rebasing on master, one sec

rpkyle:
@joelostblom There's a spurious visual diff that can't possibly be caused by your minor edits, so I'm going to use my magic wand to merge anyhow. I suspect this is caused by a nearly rendered index page, but will sort that out later. Thanks! :woman_dancing:

PR #1089, opened 2021-02-03
Replace this with self in React for Python guide classes example

### LINKED ISSUE #1088: Python class example in React for Python guide should pass self instead of this
The Python class example in the [React for Python Developers](https://dash.plotly.com/react-for-python-developers) guide should pass `self` to the `render` method instead of `this`.

https://github.com/plotly/dash-docs/blob/097ffa75f4ee662de1c6f58ee8229696673b2607/dash_docs/chapters/react_for_python_developers/index.py#L356

Also, this may be my ignorance with respect to JavaScript but shouldn't the `init` method in the JS class example take `a` as an argument?

https://github.com/plotly/dash-docs/blob/097ffa75f4ee662de1c6f58ee8229696673b2607/dash_docs/chapters/react_for_python_developers/index.py#L363-L366

Finally, there are two apparent differences between the two class examples:
1. The Python class passes `a` to the parent class' `__init__` method whereas in the JS class example, `super()` is called without arguments.
2. The Python class example assigns `self.a = a` before calling the parent class' `__init__` method whereas in the JS class example, the order is the opposite (probably not significant).

I can submit a PR following feedback on the different items above.

🚀 Great guide BTW!

DrGFreeman:
Make Python and JS class examples in [React for Python Developers](https://dash.plotly.com/react-for-python-developers) guide similar.

Fixes #1088

<hr>

Post-merge checklist:

- [x] I understand

### DIFF
--- a/dash_docs/chapters/react_for_python_developers/index.py
+++ b/dash_docs/chapters/react_for_python_developers/index.py
@@ -353,7 +353,7 @@ class MyComponent(Component):
         self.a = a;
         super().__init__(a)
 
-    def render(this):
+    def render(self):
         return self.a;
 ```
 
--- a/dash_docs/chapters/react_for_python_developers/index.py
+++ b/dash_docs/chapters/react_for_python_developers/index.py
@@ -347,7 +347,7 @@ console.log(add(4, 6)); // 10
 > Heads up! Classes, among other features, are new language features in JavaScript. Technically, they're part of a new version of JavaScript called ES6. When we build our JavaScript code, a tool called Babel will convert these new language features into simpler JavaScript that older browsers like IE11 can understand.
 
 JavaScript classes are very similar to Python classes. For example, this Python class:
-```js
+```py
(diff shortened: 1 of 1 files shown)


dash_docs/chapters/react_for_python_developers/index.py
```diff
@@ -350,17 +350,17 @@
 ```js
```
alexcjohnson, line 350:
```suggestion
```py
```
While we're at it might as well use the right language here!

alexcjohnson (approved):
Very nice - thanks again @DrGFreeman ! 💃

dash_docs/chapters/react_for_python_developers/index.py
DrGFreeman: I noticed that but wasn't sure if it was intended for consistency with ` ```js` which does not render a background color nor syntax highlighting.

PR #1090, opened 2021-02-03
Added IGV and NGL MoleculeViewer chapters

HammadTheOne:
This PR is a work in progress to add documentation for the `dash-bio` components IGV and NglMoleculeViewer. The documentation adds examples for each of the params, and a default component example in line with existing ones. 

TODO:

- [x] Add IGV chapter
- [x] Add NGL chapter
- [x] Re-generate sitemap
- [x] Update requirements.txt 
- [x] Add parser - Followed up [**Merged and Released to PyPi**](https://github.com/plotly/dash-bio-utils/tree/0.0.6-release)
- [x] Make patch release of 0.6.1 of `dash-bio` with ref fix for NGL [**Merged and Released to PyPi**](https://github.com/plotly/dash-bio/pull/544) 

URLs to visit:

/dash-bio/igv
/dash-bio/nglmoleculeviewer

Post-merge checklist:

- [ ] I understand


alexcjohnson:
@HammadTheOne great example apps! Three things I notice:

The dropdown built into IGV doesn't like the padding we apply in our CSS. Seems like it needs an override. Here's what I see:
Screen Shot 2021-02-16 at 6 45 17 PM
Then if I disable this padding it looks good:
Screen Shot 2021-02-16 at 6 46 18 PM

In the NGL viewer, this one may be tougher to figure out how to resolve, but when I first load the example page a lot of the molecules are not drawn. Each one will show up after I start interacting with it (changing the related controls, for example). Or if I click off to a different page and come back they usually all show up. I wonder if there are just too many of them? Or something about async loading?

Last, the "see it in action" links for both IGV and NGL point to restricted apps. I'm not sure we need those at all since we have live examples right here, but if we keep them they should be public.

HammadTheOne:
@alexcjohnson Turns out the `resetView` prop was the reason for some of the examples not showing, and I've gone ahead and fixed that in e13b9ea. To fix the awkward CSS, I've added an override for that particular selector to `dashbio.css` in the assets directory with 5408d1a . 

For the `See it in action` links, I have yet to deploy those demo apps to the gallery, but once that's done, they'll be accessible, but the links point to where they will eventually belong, but I think those can go up even after merging this PR.

alexcjohnson (approved):
💃 Fantastic! The `resetView` prop may need a little more thorough documentation in order to help users with this kind of issue, but that should be handled in the dash-bio repo (also when you have nested props like that, it's better if the docstrings are also nested, rather than one big docstring at the top level. Not urgent, but perhaps you can make an issue to follow up on this.)

PR #1098, opened 2021-02-10
Remove duplicate markdown in Editable DataTable page

cpascale43:
Currently, the Editable DashTable page displays the title section and Predefined Columns sections twice. This PR removes the duplicate sections.

Screen Shot 2021-02-10 at 5 58 44 PM

### DIFF
--- a/dash_docs/chapters/dash_datatable/editing/index.py
+++ b/dash_docs/chapters/dash_datatable/editing/index.py
@@ -6,61 +6,53 @@ from dash_docs import tools
 from dash_docs import styles
 from dash_docs import reusable_components as rc
 
-
 examples = tools.load_examples(__file__)
 
 Display = rc.CreateDisplay({
     'dash_table': dash_table,
 })
 
-
 layout = html.Div([
 
-    rc.Markdown(
-        '''
-        # Editable DataTable
-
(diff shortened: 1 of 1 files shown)


alexcjohnson:
Nice catch! But I probably would have removed the other one, there's no reason this needs to be three separate `dcc.Markdown` components. Or just take the three components that are left and combine them into one like the deleted section.

dash_docs/chapters/dash_datatable/editing/index.py
```diff
@@ -16,13 +15,13 @@
 
     rc.Markdown('''
     # Editable DataTable
-
```
alexcjohnson: My commit message says trailing newlines but I meant trailing whitespace. Unfortunately if you're editing in the github.com UI it's hard to avoid generating this kind of whitespace, but if you edited on your local machine you should be able to have your editor automatically strip trailing whitespace when saving a file.

alexcjohnson (approved):
💃 Looks good!

dash_docs/chapters/dash_datatable/editing/index.py
cpascale43: Good tip, thanks!

PR #1101, opened 2021-02-23
revamp installation page: more up-to-date description, and recommend pandas

alexcjohnson:
Most of our examples use pandas or require it indirectly via px - so until we bring it in implicitly (planned for plotly.py 6.0, dash 2.0) we should mention it in the installation section.

Also cleans up the prose - the displayed command was correct, but the description still said "In your terminal, install several dash libraries" left over from when you had to install dcc, html, and table separately.


chriddyp (approved):
Much better! 💃🏼

PR #926, opened 2020-07-10
Updated the example for active and selected cells to include border s…

ann-marie-ward:
…tyle

Post-merge checklist:

The master branch is auto-deployed to `dash.plotly.com`.
Once you have merged your PR, wait 5-10 minutes and check dash.plotly.com
to verify that your changes have been made.

- [ ] I understand

If this PR documents a new feature of Dash:

- [ ] Comment on the original Dash issue with a link to the new docs.
- [ ] Reply to any community thread(s) asking for this feature.

If this PR includes a new dataset available at a remote URL:
- [ ] I have added this dataset to the `datasets/` folder
- [ ] I have added a mapping between the remote URL and the filename in the
`datasets/` folder into the `find_and_replace` dict in `dash_docs/tools.py`

If this PR adds an image or animated GIF:
- [ ] This image was saved and referenced locally rather than via an external link

If I introduced a new relative link inside `dcc.Markdown`:
- [ ] I considered whether I could replace the `dcc.Markdown` call with `rc.Markdown`, which will replace relative links with `tools.relpath` internally. Otherwise, I used e.g. `<dccLink href=tools.relpath('/layout') children="the first chapter"/>` instead of `[the first chapter](/layout)` (importing `tools` from `dash_docs`), and set `dangerously_allow_html=true` in the `dcc.Markdown` call.

If I changed the `chapter_index` by removing or relocating a page:
- [ ] I added a redirect in `dash_docs/server.py` from the old URL to the new URL

PR #1125, opened 2021-04-15
Update dcc.Download

tobinngo:
Documents #926

Post-merge checklist:

- [x] I understand

- [x] Comment on the original Dash issue with a link to the new docs.

- [x] This image was saved and referenced locally rather than via an external link

- [x] I considered whether I could replace the `dcc.Markdown` call with `rc.Markdown`, which will replace relative links with `tools.relpath` internally. Otherwise, I used e.g. `<dccLink href=tools.relpath('/layout') children="the first chapter"/>` instead of `[the first chapter](/layout)` (importing `tools` from `dash_docs`), and set `dangerously_allow_html=true` in the `dcc.Markdown` call.


dash_docs/chapters/dash_core_components/Download/index.py
```diff
@@ -0,0 +1,50 @@
...
+    html.H3('Downloading Images'),
+    rc.Syntax(examples['download-image.py'][0], summary=rc.Markdown('''
+        To download a file from disk use `dcc.send_file`, taking care to
+        specify the file path.
+    ''')),
+    rc.Example(examples['download-image.py'][1]),
+
+    html.Hr(),
+
+    html.H3('Downloading Dataframes'),
```
chriddyp, line 40: let's put this one above the images one

dash_docs/chapters/dash_core_components/Download/index.py
```diff
@@ -0,0 +1,50 @@
...
+    rc.Syntax(examples['download-dataframe.py'][0], summary=rc.Markdown('''
+        For downloading dataframes the many pandas export methods are supported:
+    ''')),
+    rc.Example(examples['download-dataframe.py'][1]),
+
```
chriddyp, line 45: could we add a download excel file example too, where the excel file is auto-generated from a data frame?

dash_docs/chapters/dash_core_components/Download/index.py
chriddyp, line 40: let's also call this download data frame as csv, and then have a separate example for excel file

chriddyp (approved):
looks good! just one more example for an excel file and we should be good

dash_docs/chapters/dash_core_components/Download/index.py
tobinngo: Should I also add a link to the relevant panda docs?

PR #1127, opened 2021-04-17
add dcc.Clipboard

AnnMarieW:
This is a first draft of the documentation for the new dcc.Clipboard component added in [dcc 932](https://github.com/plotly/dash-core-components/pull/932

### DIFF
--- a/dash_docs/chapters/dash_core_components/Clipboard/examples/clipboard_table.py
+++ b/dash_docs/chapters/dash_core_components/Clipboard/examples/clipboard_table.py
@@ -1,0 +1,62 @@
+
+import dash
+from dash.dependencies import Input, Output, State
+import dash_core_components as dcc
+import dash_html_components as html
+import dash_table
+import pandas as pd
+
+external_stylesheets = ["https://codepen.io/chriddyp/pen/bWLwgP.css"]
+
+app = dash.Dash(__name__, external_stylesheets=external_stylesheets)
+
+df = pd.read_csv("https://raw.githubusercontent.com/plotly/datasets/master/solar.csv")
+
+app.layout = html.Div(
+    [
+        dcc.Clipboard(
+            id="DataTable_copy",
+            style={"fontSize": 30, "color": "white", "backgroundColor": "grey", "height":38},
+            className="button",
+        ),
+        dcc.RadioItems(
+            id="copy_selected",
+            options=[
+                {"label": "Copy All", "value": "all"},
+                {"label": "Copy Selected", "value": "some"},
+            ],
+            value="all",
+            style={"display": "inline-block"},
+        ),
+        dash_table.DataTable(
+            id="DataTable",
+            columns=[{"name": i, "id": i} for i in df.columns],
+            data=df.to_dict("records"),
+            sort_action="native",
+            page_size=10,
+        ),
+    ]
+)
+
+
+@app.callback(
+    Output("DataTable_copy", "text"),
+    Input("DataTable_copy", "n_clicks"),
+    State("DataTable", "start_cell"),
+    State("DataTable", "end_cell"),
+    State("DataTable", "derived_virtual_data"),
+    State("copy_selected", "value")
+)
+def custom_copy(_, start, end, data, copy_selected):
+    dff = pd.DataFrame(data)
+    if start and (copy_selected == 'some'):
+        copy_cells = dff.loc[
+            start["row"]: end["row"], start["column_id"]: end["column_id"]
+        ]
+        return copy_cells.to_string(header=False)
+    else:
+        return dff.to_string()  # includes headers
+
+
+if __name__ == "__main__":
+    app.run_server(debug=True)
--- a/dash_docs/chapters/dash_core_components/Clipboard/index.py
+++ b/dash_docs/chapters/dash_core_components/Clipboard/index.py
@@ -8,59 +8,68 @@ from dash_docs import reusable_components as rc
 
 examples = tools.load_examples(__file__)
 
-layout = html.Div(children=[
-    html.H1('Clipboard Examples and Reference'),
-
-    rc.Markdown("""
+layout = html.Div(
+    children=[
+        html.H1("Clipboard Examples and Reference"),
+        rc.Markdown(
+            """
         The Clipboard component copies text to the 
         browser's clipboard by clicking on a copy icon.    
-    """),
-    html.H3('Simple Clipboard Example'),
-    rc.Markdown("""
+    """
+        ),
+        html.H3("Simple Clipboard Example"),
+        rc.Markdown(
+            """
         The easiest way to trigger the copy is by using the the `target_id` 
         property. No callback is required!
 
         Place `dcc.Clipboard()` in the layout where you would like the copy
          icon located. Specify the `target_id` of the component with text to copy.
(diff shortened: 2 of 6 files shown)


dash_docs/chapters/dash_core_components/Clipboard/examples/clipboard_table.py
```diff
@@ -0,0 +1,62 @@
...
+            data=df.to_dict("records"),
+            sort_action="native",
+            page_size=10,
+        ),
+    ]
+)
+
+
+@app.callback(
+    Output("DataTable_copy", "text"),
```
chriddyp, line 43: This is so cool 😸

dash_docs/chapters/dash_core_components/Clipboard/index.py
```diff
@@ -0,0 +1,66 @@
...
+    html.H1('Clipboard Examples and Reference'),
+
+    rc.Markdown("""
+        The Clipboard component copies text to the 
+        browser's clipboard by clicking on a copy icon.    
+    """),
+    html.H3('Simple Clipboard Example'),
+    rc.Markdown("""
+        The easiest way to trigger the copy is by using the the `target_id` 
+        prop. No callback is required!
```
chriddyp, line 21:
```suggestion
        property. No callback is required!
```

dash_docs/chapters/dash_core_components/Clipboard/index.py
```diff
@@ -0,0 +1,66 @@
...
+
+        Place dcc.Clipboard() in the layout where you would like the copy
```
chriddyp, line 23:
```suggestion
        Place `dcc.Clipboard()` in the layout where you would like the copy
```

dash_docs/chapters/dash_core_components/Clipboard/index.py
```diff
@@ -0,0 +1,66 @@
...
+     The next example shows the icon styled like a button.  
+   """),
+
+    html.Div(
+        examples['clipboard_markdown.py'][1],
+        className='example-container',
+        style={'overflow': 'hidden', 'padding': '20px'}
+    ),
+
+    html.H3('Updating Text in a callback'),
```
chriddyp, line 46:
```suggestion
    html.H3('Updating the Clipboard Text in a callback'),
```

chriddyp: just to make it a bit clearer on what "text" is referring to.

chriddyp: Although I wonder if we should use a more descriptive property name. Instead of `text`, like `content`? To me, `text` feels like something that is visually rendered.

dash_docs/chapters/dash_core_components/Clipboard/index.py
```diff
@@ -0,0 +1,66 @@
...
+    rc.Markdown("""When `target_id` is not specified, the content of the `text` prop
+    is copied to the clipboard.  This works well with components like the DataTable where
+    you may want to customized the text in a callback.  In this example, 
+    the dataframe is converted to text with pandas `to_string()`.  See the pandas documentation
```
chriddyp, line 50: Looks like pandas even has a `to_clipboard`! Maybe we should use that? https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_clipboard.html

dash_docs/chapters/dash_core_components/Clipboard/index.py
```diff
@@ -0,0 +1,66 @@
...
+    rc.Markdown(
+        examples['clipboard_table.py'][0],
+        style=styles.code_container,
+    ),
+    html.Div(
+        examples['clipboard_table.py'][1],
+        className='example-container',
+        style={'overflow': 'hidden', 'padding': '20px'}
+    ),
+
```
chriddyp, line 63: One other consideration - Are there any known limitations? If so, let's add a `html.H3('Limitations')` section to describe any gotchas in advance. This might include things like:
- Browser (in)compatability?
- Is there a limit to the amount of data that can be put into a clipboard?
- Other things mentioned in https://developer.mozilla.org/en-US/docs/Web/API/Clipboard?

dash_docs/chapters/dash_core_components/Clipboard/index.py
AnnMarieW: I started with the to_clipboard() but @alexcjohnson pointed out that it would fail in production because it would copy to the servers clipboard.

dash_docs/chapters/dash_core_components/Clipboard/index.py
```diff
@@ -0,0 +1,66 @@
...
+    ),
+    rc.Markdown(
+        examples['clipboard_table.py'][0],
+        style=styles.code_container,
+    ),
+    html.Div(
+        examples['clipboard_table.py'][1],
+        className='example-container',
+        style={'overflow': 'hidden', 'padding': '20px'}
+    ),
```
chriddyp, line 62: It might be nice to have a section that describes how other programs _interpret_ the clipboard content. We could specifically call out Excel.

Basically try to explain to the community how the clipboard "text" is like "interpreted" / "converted" by other programs to turn that text into first class objects in that program (like cells of a spreadsheet). 

We could pull some of the links or copy from MDNs docs on the subject:

image

AnnMarieW: good idea - the text does copy very nicely to Excel

dash_docs/chapters/dash_core_components/Clipboard/index.py
chriddyp: Ah I see. I wonder if `to_clipboard`'s data structure accounts for things in a special way that is more easily interpreted by excel than that of `to_string`.

chriddyp: Here's the source for `to_clipboard` - https://github.com/pandas-dev/pandas/blob/0bd8138316e797712cd1de8e36baaa0730941ac5/pandas/io/clipboards.py#L104-L139. Looks like it uses `to_csv`, `utf-8` encoding, and a couple of other things to make it work well. So perhaps we incorporate some of those pieces into the example

dash_docs/chapters/dash_core_components/Clipboard/index.py
AnnMarieW: I added some content to the Limitations section.   I'm couldn't find a definitive answer on the clipboard size limits

dash_docs/chapters/dash_core_components/Clipboard/index.py
AnnMarieW: I'm not sure how to create an example based on the pandas source code.  Would this be a good use-case for a dash-labs component plug-in?

dash_docs/chapters/dash_core_components/Clipboard/index.py
AnnMarieW: I missed your second comment here.  You have a good point about changing the prop name from `text` to `content`.  Currently dcc.Clipboard only supports copying text, but this could be expanded to include things like images in the future.  Then `content` would be a better prop name.

alexcjohnson:
Brought this into the new docs some time ago, based on the work here - thanks @AnnMarieW !

Issue #1133, opened 2021-04-27
Add dash-vtk API reference to dash-docs


xhlulu:
During my previous PR I only added the user guide - I should have also added the API reference. It should be fairly straightforward since I've done it in the past.

PR #1141, opened 2021-04-30
Dash VTK: Add API reference and new interaction section

xhlulu:
Closes https://github.com/plotly/dash-docs/issues/1133
closes https://github.com/plotly/dash-docs/issues/1142
closes https://github.com/plotly/dash-docs/issues/1143

This will also include information about using picking information (e.g. hover and click events)

FYI @jackparmer 

Post-merge checklist:

- [ ] I understand


jackparmer:
Is it possible to also clean up this sidebar? It's not useful atm - it's just a big clutter of stuff:

image

dash_docs/chapters/dash_vtk/representations/index.py
```diff
@@ -0,0 +1,335 @@
...
+                dash_vtk.PointData([
+                    dash_vtk.DataArray(
+                    registration='setScalars',
+                    values={kwargs.get('scalars')}
+                    )
+                ])
+                ],
+            )
+            ],
+        )
```
alexcjohnson, line 271:
```suggestion
            return dash_vtk.GeometryRepresentation(
                id=kwargs.get('id'),
                colorMapPreset=kwargs.get('colorMapPreset'),
                colorDataRange=kwargs.get('colorDataRange'),
                property=kwargs.get('property'),
                children=[
                dash_vtk.PolyData(
                    points=kwargs.get('xyz'),
                    connectivity='points',
                    children=[
                    dash_vtk.PointData([
                        dash_vtk.DataArray(
                        registration='setScalars',
                        values={kwargs.get('scalars')}
                        )
                    ])
                    ],
                )
                ],
            )
```

dash_docs/chapters/dash_vtk/representations/index.py
```diff
@@ -0,0 +1,335 @@
...
+                                registration='setScalars',
+                                values=kwargs.get('scalars'),
+                            )
+                        ])
+                    ],
+                ),
+                ],
+            )
+            ],
+        )
```
alexcjohnson, line 315:
```suggestion
            return dash_vtk.VolumeRepresentation(
                id=kwargs.get('id'),
                colorMapPreset=kwargs.get('colorMapPreset'),
                colorDataRange=kwargs.get('colorDataRange'),
                property=kwargs.get('property'),
                mapper=kwargs.get('mapper'),
                volume=kwargs.get('volume'),
                children=[
                    dash_vtk.VolumeController(
                        rescaleColorMap=kwargs.get('rescaleColorMap'),
                        size=kwargs.get('size'),
                    ),
                    dash_vtk.ImageData(
                        dimensions=kwargs.get('dimensions'),
                        origin=kwargs.get('origin'),
                        spacing=kwargs.get('spacing'),
                        children=[
                            dash_vtk.PointData([
                                dash_vtk.DataArray(
                                    registration='setScalars',
                                    values=kwargs.get('scalars'),
                                )
                            ])
                        ],
                    ),
                ],
            )
```

dash_docs/chapters/dash_vtk/representations/index.py
```diff
@@ -0,0 +1,335 @@
...
+        ```
+
+        The set of convenient properties are as follows:
+        - __xyz__ = list of xyz of each point inside a flat array
+        - __colorMapPreset__ = color preset name to use
+        - __colorDataRange__ = rescale color map to provided that range
+        - __property__ = {} # Same as GeometryRepresentation/property
+        - __rgb__ / __rgba__ / __scalars__ = [...] let you define the field you want to color your point cloud with. The rgb(a) expects numbers up to 255 for each component: Red Green Blue (Alpha).
```
alexcjohnson, line 279:
```suggestion
        - __rgb__ / __rgba__ / __scalars__ = `[...]` let you define the field you want to color your point cloud with. The rgb(a) expects numbers up to 255 for each component: Red Green Blue (Alpha).
```
I don't think this was intended to be a link 😉

alexcjohnson:
Running this locally in debug mode, (because for some reason the review app didn't deploy on the latest commit?) if I try to click from a different page onto `vtk/other` I get an error every time, but each error is different, and then it fails to go to the page. Here's clicking to that page 4 times. None of the other pages show any errors, and I don't get an error if I just type the `vtk/other` url directly (but I still do get an error if I click away and try to go back to it)
Screen Shot 2021-05-11 at 10 20 22 PM

alexcjohnson:
Hmm, it's erratic: sometimes I can get to `vtk/other` once from another page, but then as soon as I've tried and gotten an error, every time after that I get another error.

dash_docs/chapters/dash_vtk/other/index.py
```diff
@@ -0,0 +1,176 @@
...
+    ## Mesh
+
+    This element is a helper on top of __PolyData__ which has a Python helper function that goes with it which will help you map a __vtkDataSet__ into a single property of the __Mesh__ element.
+
+    ```py
+    def Mesh(**kwargs):
+        return dash_vtk.PolyData(
+            **kwargs.get('state').get('mesh'),
+            children=[
+                dash_vtk.[kwargs.get('state').get('field').get('location')]([
```
jourdain, line 23: This should only happen if `field` exist. In some cases a mesh don't have any field.

jourdain: But I guess here it is just to illustrate the idea behind the structure and helper

xhlulu: Not sure I follow. Do you think you can make a suggestion (with `ctrl+g`)?

jourdain: I don't know about `ctrl+g`.

I was just saying that that piece of code is not fully true. It mostly illustrate how the helper works but we need to have a disclaimer that mention the fact that `state.field` is optional and therefore when not available, the `children` is not created. But for clarity and understanding that part was left off in that simple code snippet.

dash_docs/chapters/dash_vtk/other/index.py
```diff
@@ -0,0 +1,176 @@
...
+                    dash_vtk.DataArray(
+                      **kwargs.get('state').get('field'),
+                    )
+                ])
+            ]
+        )
+    ```
+
```
xhlulu, line 31:
```suggestion

Note that we assume that `state.field` exists when we use `kwargs.get('state').get('field').get('location')` in the snippet above. However, `state.field` is optional, so when it's not available the `children` is not created, and `kwargs.get('state').get('field').get('location')` would not be needed.

```

@jourdain is that what you meant?

jourdain: perfect!

alexcjohnson (approved):
OK, as discussed on Slack, it seems like the problem with `/vtk/other` was likely about running out of GL contexts, between the ones on the previous page not yet released by the browser when the ones on `/vtk/other` were instantiated. I guess the solution is to never have more than two VTK demos on one page.

I'll also copy the other issue I mentioned on slack - I think we should merge this PR (so: 💃 ), but separately investigate this:

sometimes some of the demos will get confused about mouse interactions, deciding to spin wildly with a small drag. Not shown in that gif, but when this happens the scene also gets confused about its own depth: if I zoom in or out just a little bit the rendered object gets clipped and soon disappears.
This gif is the bottom example on `/vtk/intro` but I’ve seen this a few other places too.

Looks like I can trigger this reliably by loading `/vtk/intro` initially then clicking to `/vtk/structure` and back to `/vtk/intro`. If I initially load structure and then click to intro, the bottom demo is already messed up and I only see glimmers of it showing anything.
vtk-spin

PR #1150, opened 2021-05-24
Fix datatable example so pagination buttons are visible

AnnMarieW:
This fixes the example in the [DataTable Interactivity chapter](https://dash.plotly.com/datatable/interactivity) so  the pagination buttons are visible. 

Thanks to @sujeet.cusat  for reporting this on the [forum](
https://community.plotly.com/t/pagination-is-not-visible-when-using-dash-table-with-graph/53216)

The only real change is to add some margin, the other changes are from running `black`

image

### DIFF
--- a/dash_docs/chapters/dash_datatable/interactivity/examples/interactivity_connected_to_graph.py
+++ b/dash_docs/chapters/dash_datatable/interactivity/examples/interactivity_connected_to_graph.py
@@ -5,47 +5,55 @@ import dash_core_components as dcc
 import dash_html_components as html
 import pandas as pd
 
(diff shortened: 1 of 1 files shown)


alexcjohnson:
This is a good workaround in the short term; however I think the more useful solution would be to ensure the pagination buttons make space for themselves within the table itself. I suspect this is the same issue as https://github.com/plotly/dash-table/issues/834

AnnMarieW:
> This is a good workaround in the short term; however I think the more useful solution would be to ensure the pagination buttons make space for themselves within the table itself. I suspect this is the same issue as [plotly/dash-table#834](https://github.com/plotly/dash-table/issues/834)

RIght - looks like the same problem.  I'll see if I can make the fix in dash-table.

PR #1203, opened 2022-03-17
minor typo in app_json.md

rgbailey:
NOTICE: This repo is no longer used to host dash.plotly.com
It is still (as of Sept 1 2021) used for dashr.plotly.com, but this will change soon, so even for R please do not contribute here.

If you are a Plotly org member, please visit https://github.com/plotly/ddk-dash-docs and make your PR there.

Unfortunately for our community contributors, this repo is private. That allows us to provide a better docs experience by using some of our closed-source packages such as dash-design-kit, but it means we can no longer accept community PRs. We still very much welcome your feedback on the docs! Please continue to make issues in this repo describing the changes you would like to see, and a Plotly member will take it from there.


alexcjohnson:
Thanks @rgbailey - @LiamConnors looks like this typo DOES exist (3 copies of it!) in the current structure?

red-patience:
Nice catch @rgbailey, thanks! This is fixed now.

PR #1208, opened 2022-07-27
Fix typo in `dash_docs/chapters/testing/index.py`

janosh:


red-patience:
Thanks @janosh! We're only using this repo to track issues these days, but I've recreated this fix in the new repo (which unfortunately we've had to make closed-source so it can use some of our enterprise component packages). It should be live soon. :)

PR #1218, opened 2024-06-12
Update README.md

LiamConnors:


red-patience (approved):
Makes sense to me to direct folks toward a different repo so that we can consolidate!

